PR #1, opened 2020-11-03
minor fixes

atomotic:


ikreymer:
Thanks, good catch! Still testing things, thanks for being an early adopter @atomotic :)

PR #5, opened 2021-01-14
Added limit info to statsFilename

rgaudin:
- added new `limit` dict to statsFilename
- `limit` dict composed of:
  - `max`: the limit requested (or `0`)
  - `hit`: boolean whether limit was reached or not
- limit now limiting to `n` pages for `--limit n` instead of `n - 1`

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -32,6 +32,9 @@ class Crawler {
     // links crawled counter
     this.numLinks = 0;
 
+    // was the limit hit?
+    this.limitHit = false;
+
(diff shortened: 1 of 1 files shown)


ikreymer:
The stats change looks good, but I think `--limit n` should be n total pages, not n links.
With this change, `--limit 1` would crawl the starting page + 1 link, so actually 2 pages.

rgaudin:
OK, restored the check to what it was.

Issue #2, opened 2020-12-16
Add support for WACZ creation.


ikreymer:
This can be a command-line flag, say `--generateWACZ` similar to the `--generateCDX` option, which will generate a WACZ file after the crawl is done. This will also require keeping track of the pages crawled in a list that can be passed into py-wacz

This would involve:
- [ ] Adding a `--generateWACZ` command line option.
- [ ] Generate a `pages/pages.jsonl` file in the collection directory. Will need to make the `pages` dir also.
- [ ] Run py-wacz to create the WACZ at the end of the crawl. For now, can just regenerate the CDX during wacz creation. In the future, can use the existing index in redis to speed up the process.

PR #6, opened 2021-01-20
Add wacz support to browsertrix

emmadickson:
Supports https://github.com/webrecorder/browsertrix-crawler/issues/2
`

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -468,7 +468,7 @@ class Crawler {
       }
       fs.writeFileSync('pages/pages.jsonl', jsonl)
     } catch (err) {
-      console.warn("Stats output failed", err);
+      console.warn("Pages output failed", err);
     }
   }
 
@@ -500,7 +500,7 @@ class Crawler {
   createPage(url){
     var id = uuidv4();
     var today = new Date();
-    var row = {"id": id, "url": url.href, "ts": today}
+    var row = {"id": id, "url": url.href, "title": page.title()}
     this.pages.push(row)
   }
   
--- a/crawler.js
+++ b/crawler.js
@@ -33,6 +33,9 @@ class Crawler {
     // links crawled counter
     this.numLinks = 0;
 
+    // was the limit hit?
+    this.limitHit = false;
+
     this.monitor = true;
 
     this.userAgent = "";
@@ -66,7 +69,7 @@ class Crawler {
     if (this.emulateDevice) {
       this.userAgent = this.emulateDevice.userAgent;
     } else {
-      let version = "84";
+      let version = "88";
 
       try {
         version = child_process.execFileSync("google-chrome", ["--product-version"], {encoding: "utf8"}).trim();
@@ -333,6 +336,7 @@ class Crawler {
       "--no-sandbox",
       "--disable-background-media-suspend",
       "--autoplay-policy=no-user-gesture-required",
+      "--disable-features=IsolateOrigins,site-per-process",
     ];
   }
 
@@ -435,8 +439,8 @@ class Crawler {
       const total = this.cluster.allTargetCount;
       const workersRunning = this.cluster.workersBusy.length;
       const numCrawled = total - this.cluster.jobQueue.size() - workersRunning;
-
-      const stats = {numCrawled, workersRunning, total};
+      const limit = {max: this.params.limit || 0, hit: this.limitHit};
+      const stats = {numCrawled, workersRunning, total, limit};
 
       try {
         fs.writeFileSync(this.params.statsFilename, JSON.stringify(stats, null, 2))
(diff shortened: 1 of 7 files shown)


crawler.js
```diff
@@ -463,9 +497,17 @@ class Crawler {
     return true;
   }
 
+  createPage(url){
+    var id = uuidv4();
+    var today = new Date();
+    var row = {"id": id, "url": url.href, "ts": today}
```
ikreymer, line 503: Currently, py-wacz would reject the page, if the `ts` doesn't match the timestamp in the WARC/CDX.
Depending on how slow the page loading is, this may or may not always work.

I was thinking we can just omit the `ts` here and let the auto-detection find the best match.
Or, we could alter py-wacz, and use this as a hint, to find the closest match to this ts?

Also, could add the title here as that should be available from `page.title()`

crawler.js
```diff
@@ -433,15 +459,23 @@ class Crawler {
...
     this.queueUrls(results);
+
+    try {
+      var jsonl = ''
+      for (const page of this.pages){
+        jsonl = jsonl + JSON.stringify(page) + "\n"
+      }
+      fs.writeFileSync('pages/pages.jsonl', jsonl)
+    } catch (err) {
+      console.warn("Stats output failed", err);
```
ikreymer, line 471:
```suggestion
      console.warn("Pages output failed", err);
```

crawler.js
```diff
@@ -433,15 +459,23 @@ class Crawler {
...
     }
-
     this.queueUrls(results);
+
+    try {
+      var jsonl = ''
+      for (const page of this.pages){
+        jsonl = jsonl + JSON.stringify(page) + "\n"
+      }
+      fs.writeFileSync('pages/pages.jsonl', jsonl)
```
ikreymer, line 469: The pages directory should be in the same directory as the collection, ideally, eg. next to `archive/`, and should also be created if it doesn't exist..

crawler.js
```diff
@@ -90,6 +91,7 @@ class Crawler {
     this.configureUA();
 
     this.headers = {"User-Agent": this.userAgent};
+    this.pages = [{"format": "json-pages-1.0", "id": "pages", "title": "All Pages", "hasText": false}]
```
ikreymer, line 95: Since we currently allow adding to an existing collection (and probably good to allow, though maybe shouldn't be the default?), this should probably check if `pages/pages.jsonl` already exists, and if it does, we don't output this, and instead append to the pages/pages.jsonl

crawler.js
```diff
@@ -402,7 +409,26 @@ class Crawler {
...
+    
+      // Access the collections the user has specified
+      var dir = "collections/" + this.params.collection + "/archive"
+      
+      // Get a list of the warcs inside
+      var file_list = fs.readdirSync(dir)
+      
+      // Build the argument list to pass to the wacz create command
+      var argument_list = ["create", "-f"]
+      file_list.forEach((val, index) => argument_list.push("collections/" + this.params.collection + "/archive/" + val));
```
ikreymer, line 424: small nitpick: maybe safer to use path.join() for these?

crawler.js
ikreymer: As per above comment, should probably append to an existing pages.jsonl if it already exists, otherwise create new and write the header

crawler.js
```diff
@@ -463,9 +497,17 @@ class Crawler {
...
+    var id = uuidv4();
+    var today = new Date();
+    var row = {"id": id, "url": url.href, "title": page.title()}
+    this.pages.push(row)
+  }
+  
   shouldCrawl(url) {
     try {
       url = new URL(url);
+      this.createPage(url)
```
ikreymer, line 510: Probably shouldn't call createPage() here as this function just checks is a URL should be crawled..
The best place to add it might be in the task function or in the driver, but then more work in the driver code..

crawler.js
```diff
@@ -381,7 +389,6 @@ class Crawler {
         console.warn(e);
       }
     });
-
```
ikreymer, line 384: One place to add the page would be in the task() function above after the driver call:
```
this.createPage(opts.data.url)
```

crawler.js
```diff
@@ -437,15 +463,28 @@ class Crawler {
...
       return;
     }
-
     this.queueUrls(results);
+
+    try {
+      var filePath = path.join("collections/", this.params.collection, 'archive/pages/pages.jsonl')
+      var dirPath = path.join("collections/", this.params.collection, 'archive/pages')
+      if (fs.existsSync(dirPath) == false) {
+        fs.mkdirSync(path.join('collections/', this.params.collection, 'archive/pages'))
```
ikreymer, line 472:
```suggestion
        fs.mkdirSync(dirPath)
```

crawler.js
```diff
@@ -437,15 +463,28 @@ class Crawler {
       console.warn("Link Extraction failed", e);
       return;
     }
-
     this.queueUrls(results);
+
+    try {
+      var filePath = path.join("collections/", this.params.collection, 'archive/pages/pages.jsonl')
```
ikreymer, line 469: small nitpick: just reuse dirPath
```suggestion
      var filePath = path.join(dirPath, 'pages.jsonl')
```

Issue #19, opened 2021-02-04
Add eslint


ikreymer:
Add eslint to cleanup all the code and keep it consistent!

PR #21, opened 2021-02-04
add ci and linting

emmadickson:
Supports https://github.com/webrecorder/browsertrix-crawler/issues/19


crawler.js
```diff
@@ -549,11 +551,9 @@ class Crawler {
     }
   }
 
-  async directFetchCapture(url) {
+  async directFetchCapture() {
     //console.log(`Direct capture: ${this.capturePrefix}${url}`);
     const abort = new AbortController();
-    const signal = abort.signal;
```
ikreymer, line 555: Why did this get removed? This is the functionality of direct fetch...

PR #24, opened 2021-02-09
background behaviors refactor: (fixes #23)

### LINKED ISSUE #23: Make background behaviors more modular
Currenly, the defaultDriver has several hard-coded functions for autoplay, autofetch and autoscroll behaviors.
Instead, move the to separate files, and add a background behavior maanager.
Allow toggling which background behaviors are run via a separate command-line option.

This will make the background behaviors more modular and further simplify the defaultDriver

ikreymer:
- move auto-play, auto-fetch and auto-scroll behaviors to behaviors/global/*
- bgbehaviors manages these background behaviors
- command line --bgbehaviors option specifies which background behaviors to run (defaults to auto-fetch and auto-play)


emmadickson (commented):
I like this refactoring I think it will make things more modular

PR #30, opened 2021-02-24
tests text extraction

emmadickson:


.github/workflows/ci.yaml
```diff
@@ -36,11 +36,20 @@ jobs:
...
+      run: npm install
     - name: build docker
       run: docker-compose build
     - name: run crawl
-      run: docker-compose run crawler crawl --url http://www.example.com/ --generateWACZ  --collection wr-net --workers 2              
+      run: docker-compose run crawler crawl --url http://www.example.com/ --generateWACZ  --text --collection wr-net --workers 2              
     - name: validate existing wacz
       run: docker-compose run crawler wacz validate --file collections/wr-net/wr-net.wacz
+    - name: unzip wacz
+      run: docker-compose run crawler unzip collections/wr-net/wr-net.wacz -d collections/wr-net/wacz
```
ikreymer, line 48: I guess could also unzip the `unzip collections/wr-net/wr-net.wacz pages/pages.jsonl` here, which would then fail if it didn't exist, saving the extra check later. But, this makes it easier to add additional tests if we want to later on other files.

package.json
```diff
@@ -7,6 +7,7 @@
   "license": "MIT",
   "dependencies": {
     "abort-controller": "^3.0.0",
+    "md5": "^2.3.0",
```
ikreymer, line 10: very minor, but probably should move this to devDependencies, right?

tests/text.test.js
```diff
@@ -0,0 +1,22 @@
...
+test('check that the pages.jsonl file exists in the collection under the pages folder', () => {
+  expect(fs.existsSync('crawls/collections/wr-net/pages/pages.jsonl')).toBe(true);
+});
+
+test('check that the pages.jsonl file exists in the wacz under the pages folder', () => {
+  expect(fs.existsSync('crawls/collections/wr-net/wacz/pages/pages.jsonl')).toBe(true);
+});
+
+test('check that the hash in the pages folder and in the unzipped wacz folders match', () => {
+  const browsertrix_hash = md5(JSON.parse(fs.readFileSync('crawls/collections/wr-net/wacz/pages/pages.jsonl', 'utf8').split('\n')[1])['text']);
```
ikreymer, line 14: also very minor, could rename this to `crawl_hash` or `collection_hash`, at first i was confused what it meant and why it was different from wacz_hash in the first place : )

.github/workflows/ci.yaml
emmadickson: yeah i figured we would be adding more tests as we add to development

PR #32, opened 2021-03-08
Browsertrix Behaviors + Logging Improvements + Docker Image improvementts

ikreymer:
This PR switches form the internal behaviors to instead use the standalone https://github.com/webrecorder/browsertrix-behaviors repo, which contains all the behaviors as a single file to be injected into the page.

Also cleans up the logging options into a single `--logging` option, which can be set to include the stats (default), pywb logging and behavior logging.

The docker image is also updated with latest gpg necessary for installation, and python requirements moved to requirements.txt


emmadickson (approved):
LGTM. I was able to confirm it worked by running ``` docker-compose run crawler crawl --url https://twitter.com/CaptureSample --limit 1 --logging behaviors --generateWACZ```
```

PR #34, opened 2021-03-22
Profiles: Support for running with existing profiles + saving profile after a login

ikreymer:


create-login-profile.js
```diff
@@ -0,0 +1,178 @@
...
+      "-screen",
+      "0",
+      process.env.GEOMETRY,
+      "-ac",
+      "+extension",
+      "RANDR"
+    ]);
+  }
+
+  //await new Promise(resolve => setTimeout(resolve, 2000));
```
emmadickson, line 65: Why is this commented out?

crawler.js
```diff
@@ -43,6 +44,7 @@ class Crawler {
 
     this.userAgent = "";
     this.behaviorsLogDebug = false;
+    this.profileDir = fs.mkdtempSync(path.join(os.tmpdir(), "profile-"));
```
emmadickson, line 47: so will the profile information be stored in this temp folder in the docker image?

ikreymer: yeah, the profile is created in a temp directory, but then exported as a tarball and copied to the location specified by `--filename`. Need to add docs to explain this, and ideally, add a test as well..

create-login-profile.js
ikreymer: Was just testing something during development and thought a delay was needed there, but that turned out to not be needed, so should probably just remove this.

emmadickson (approved):
LGTM

Issue #35, opened 2021-03-24
Collection name validation


rebeccacremona:
If the desired collection name passed to browsertrix via `--collection` contains `.`, `/`, `:`, or potentially other special characters, `pywb` silently fails to create the necessary directory structure for the crawl. Despite that failure, the crawl then proceeds, resulting in a series of `ENOENT` errors as the process attempts to write to non-existent paths as it goes.

Browsertrix should make sure the collection name is valid before launching the crawl. `wb manager init` should also fail noisily if it is passed an invalid collection name, ideally cleanly by throwing a validation error, but also messily if directory creation fails for whatever reason.

Sample output:
```
HLSC02VK05ZHTDG:browsertrix-crawler rcremona$ time docker-compose run crawler crawl --url https://example.com/ --limit 1 --collection job-1-example.com
Creating browsertrix-crawler_crawler_run ... done
Exclusions Regexes:  []
Scope Regexes:  [ /^https:\/\/example\.com\// ]
pages/pages.jsonl creation failed Error: ENOENT: no such file or directory, mkdir '/crawls/collections/job-1-example.com/pages'
    at Object.mkdirSync (fs.js:987:3)
    at Crawler.initPages (/app/crawler.js:596:12)r 503.0 ms)
    at Crawler.crawl (/app/crawler.js:492:10)0%)
    at async Crawler.run (/app/crawler.js:411:7) {
  errno: -2,: 0.0% CPU / 0.0% memory
  syscall: 'mkdir',
  code: 'ENOENT',
  path: '/crawls/collections/job-1-example.com/pages'
}
pages/pages.jsonl append failed Error: ENOENT: no such file or directory, open '/crawls/collections/job-1-example.com/pages/pages.jsonl'tart:     2021-03-24 20:08:28.092
    at Object.openSync (fs.js:476:3)6 (running for 3.5 seconds)
    at Object.writeFileSync (fs.js:1467:35)0%)
    at Object.appendFileSync (fs.js:1506:6)
    at Crawler.writePage (/app/crawler.js:624:10)
    at Crawler.crawlPage (/app/crawler.js:458:12)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/crawls/collections/job-1-example.com/pages/pages.jsonl'
}

== Start:     2021-03-24 20:08:28.092
[...]

PR #37, opened 2021-03-31
add collection name validation

emmadickson:
Supports https://github.com/webrecorder/browsertrix-crawler/issues/35


crawler.js
```diff
@@ -114,7 +114,13 @@ class Crawler {
     }
 
     this.configureUA();
-
+    
+    // Check that the collection name is valid.
+    if (this.params.collection.search(/^[a-zA-Z0-9_-]*$/) === -1){
```
ikreymer, line 119: Let's move to this to `validateArgs()` function and also throw an error, similar to other param validation

Issue #12, opened 2021-02-03
Support for submitting an initial seed list instead of single URL


ikreymer:
Support for crawling multiple URLs in a fixed list.
These may be 'seeds' to a larger crawl or just the fixed list (need a depth option)

Probably support via multiple `--url` args as well as via a pre-made YAML config file.

PR #38, opened 2021-04-09
Create urlFileList flag for users to pass additional urls

emmadickson:
Supports #12

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ ARG REBUILD
 
 RUN yarn install
 
-ADD tests/fixtures app/fixtures
+ADD tests/fixtures/urlSeedFile.txt /app/fixtures/urlSeedFile.txt
 ADD config.yaml /app/
 ADD uwsgi.ini /app/
 ADD *.js /app/
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,6 @@ ARG REBUILD
 
 RUN yarn install
 
-ADD tests/fixtures/urlSeedFile.txt /app/fixtures/urlSeedFile.txt
 ADD config.yaml /app/
 ADD uwsgi.ini /app/
 ADD *.js /app/
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,9 +38,11 @@ ARG REBUILD
 
 RUN yarn install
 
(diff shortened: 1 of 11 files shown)


crawler.js
```diff
@@ -310,6 +317,16 @@ class Crawler {
...
       //argv.scope = url.href.slice(0, url.href.lastIndexOf("/") + 1);
       argv.scope = [new RegExp("^" + this.rxEscape(argv.url.slice(0, argv.url.lastIndexOf("/") + 1)))];
+      if (argv.urlFileList != "") {
+        let urlFileList = fs.readFileSync( path.join(__dirname, argv.urlFileList), "utf-8").split("\n");
+        for (let url of urlFileList) {
+          if (url.substr(-1) != "/"){
+            url = url + "/";
+          }
+          let urlScope = new RegExp("^" + this.rxEscape(url.slice(0, url.lastIndexOf("/") + 1)));
+          argv.scope.push(urlScope);
```
ikreymer, line 327: I wonder if we should just add a catch all regex, `.*` instead of adding one for each URL?
Since we're only going through a fixed list, so number of regexes to check doesn't grow with the list.

crawler.js
```diff
@@ -310,6 +317,16 @@ class Crawler {
     if (!argv.scope) {
       //argv.scope = url.href.slice(0, url.href.lastIndexOf("/") + 1);
       argv.scope = [new RegExp("^" + this.rxEscape(argv.url.slice(0, argv.url.lastIndexOf("/") + 1)))];
+      if (argv.urlFileList != "") {
+        let urlFileList = fs.readFileSync( path.join(__dirname, argv.urlFileList), "utf-8").split("\n");
```
ikreymer, line 321: Lets use the async api, `await fsp.readFile`

Dockerfile
```diff
@@ -38,9 +38,11 @@ ARG REBUILD
 
 RUN yarn install
 
+ADD tests/fixtures/urlSeedFile.txt /app/fixtures/urlSeedFile.txt
```
ikreymer, line 41: It would be better not to add this to the Dockerfile, since this is just for tests. Can instead mount it as a volume when running the test.

PR #51, opened 2021-05-19
Add scopeType options + option to crawl hashtags + simplify defaultDriver.js

ikreymer:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "browsertrix-crawler",
-  "version": "0.3.2",
+  "version": "0.4.0-beta.0",
```
emmadickson, line 3: Good version bump

docker-compose.yml
```diff
@@ -2,13 +2,12 @@ version: '3.5'
...
 services:
     crawler:
-        image: webrecorder/browsertrix-crawler:0.3.2
+        image: webrecorder/browsertrix-crawler:0.4.0-beta.0
         build:
           context: ./
 
         volumes:
           - ./crawls:/crawls
-          - ./tests/fixtures:/fixtures
```
emmadickson, line 11: why get rid of adding these to the docker container? Shouldn't we also change the way we call the tests in various test files or the way we call the .ci so tests continue to work as expected?

ikreymer: The volumes are just mounted manually as needed during the tests, this removes there from always being used during normal execution to avoid confusion.

PR #52, opened 2021-05-27
Screencast Support  for Debugging (fixes #43)

### LINKED ISSUE #43: Experiment with supporting screencasting
Experiment with support for optional viewing of a screencast of all crawlers in a separate HTML page, streaming frames via a websocket.

ikreymer:


screencaster.js
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    this.allWS.add(ws);
+
+    if (this.allWS.size === 1) {
+      this.startCastAll();
+    }
+
+    ws.on("close", () => {
+      console.log("ws closed");
```
emmadickson, line 67: I would expand the 'ws' to be a big clearer. 'web server' ect. I think that would be helpful

emmadickson (commented):
I would add a section to the readme explaining how you should pass the command and maybe include an example screenshot of what a user should expect to see.

PR #56, opened 2021-06-18
update collection creation

emmadickson:
Add a line to explicitly create the collection directory


ikreymer:
Closing, as issue was caused by bug in pywb, the dir should always exist at this point, created via `wb-manager init`, unless pywb crashed (and the crawl will fail)

PR #54, opened 2021-06-09
Create an argument parser class + YAML config support

emmadickson:
parse yaml
reconfigure some code

PR #59, opened 2021-06-21
Arg Parsing Refactor + Support for YAML Config Support (take 2!)

ikreymer:
Includes refactor work from #54
Simplified yaml config loading via yargs
Support for `seeds` properties, both a list via YAML config and via cli
Tests for YAML config with seeds


emmadickson (commented):
Looks good. I had a few comments.

1. I think we should get rid of the sample.yaml file if were not using it in tests. 
2. I think we should add a test that confirms the user command-line arguments will override the yaml arguments. I have one I could push to this branch. 
3. we should update the readme in the same pr. I have an updated readme i can push to this branch.

ikreymer:
Great! Yes, please feel free to do 1-3 and add to this PR!

Issue #53, opened 2021-06-02
Add optional fields to warcinfo that can identify who created the warc


karenhanson:
The warcinfo metadata does not currently include properties that could identify which institution generated the WARC file. Fields such as `operator`, `ip`, and `hostname` would be helpful additions. These could be enabled/defined via the new YAML configuration file (https://github.com/webrecorder/browsertrix-crawler/issues/13).

ikreymer:
Supported via #60, will be in 0.4.0-beta.3 release

PR #60, opened 2021-06-21
Add fields to warcinfo in combinedwarc

emmadickson:
Supports https://github.com/webrecorder/browsertrix-crawler/issues/53


crawler.js
```diff
@@ -258,11 +258,12 @@ class Crawler {
     const warcioPackageJson = JSON.parse(await fsp.readFile("/app/node_modules/warcio/package.json"));
     const pywbVersion = child_process.execSync("pywb -V", {encoding: "utf8"}).trim().split(" ")[1];
 
-    const info = {
+    var info = {
       "software": `Browsertrix-Crawler ${packageFileJSON["version"]} (with warcio.js ${warcioPackageJson} pywb ${pywbVersion})`,
       "format": "WARC File Format 1.1"
     };
 
+    info = Object.assign(info, this.params.warcInfo);
```
ikreymer, line 266: This can just be:

```suggestion
    info = {...this.params.warcInfo, ...info};
```

so we don't allow overriding `software` or `format`

util/argParser.js
```diff
@@ -233,6 +239,10 @@ class ArgParser {
       }
     }
 
+    if (argv.warcInfo){
```
ikreymer, line 242: This is not needed as argv.warcInfo is already parsed automatically!

karenhanson:
Looking at this, I noticed the WARC version in the warcinfo is WARC/1.1 while the other entries are all WARC/1.0. Not sure if there are other reasons to use 1.1. If not, could warcinfo be modified to match? WARC/1.0 is supported by more tools. Specifically, JWAT (which is also in JHOVE) will only validate to WARC 1.0 at the moment.

ikreymer:
> Looking at this, I noticed the WARC version in the warcinfo is WARC/1.1 while the other entries are all WARC/1.0. Not sure if there are other reasons to use 1.1. If not, could warcinfo be modified to match? WARC/1.0 is supported by more tools. Specifically, JWAT (which is also in JHOVE) will only validate to WARC 1.0 at the moment.

Good point, yeah, pywb is generating WARC/1.0 records by default. The main reason for 1.1 would be to support higher precision timestamps, beyond second. I'd like to switch to that eventually, but maybe not needed for now, and can probably be a flag to enable that.

ikreymer:
@karenhanson let us know if you have any other feedback on this, getting ready to merge this in :)

PR #66, opened 2021-07-07
Scope Handling Improvements + Tests

ikreymer:
Ensure custom scope includes take precedence over non-custom scope
Set scope to 'custom' if include/scopeIncludeRx is provided.
Use `scopeType` instead of `type` for scope at both top and seed level.
Add --scopeIncludeRx and --scopeExcludeRx for more specific cmdline options, still allow --include/--exclude
Add tests for various top-level and seed-level scope configurations
Update includes to work with tests
Tests: switch to yarn instead of npm for consistency, fix test issues
Update README
Bump to 0.4.0-beta.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,13 +73,15 @@ Options:
                                                            [number] [default: 0]
       --timeout                             Timeout for each page to load (in
                                             seconds)      [number] [default: 90]
-      --scope                               Regex of page URLs that should be
+      --scopeType                           Predefined for which URLs to crawl,
+                                            can be: prefix, page, host, any, or
+                                            custom, to use the
+                                            scopeIncludeRx/scopeExcludeRx
+                                                                        [string]
+      --scopeIncludeRx, --include           Regex of page URLs that should be
                                             included in the crawl (defaults to
                                             the immediate directory of URL)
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -122,7 +124,7 @@ Options:
       --userAgentSuffix                     Append suffix to existing browser
                                             user-agent (ex: +MyCrawler,
                                             info@example.com)           [string]
-      --useSitemap                          If enabled, check for sitemaps at
+      --useSitemap, --sitemap               If enabled, check for sitemaps at
```
emmadickson, line 127: do we want to aslo allow usesitemap since we do that for some other flags

util/argParser.js
```diff
@@ -169,6 +170,7 @@ class ArgParser {
       },
 
       "useSitemap": {
+        alias: "sitemap",
```
emmadickson, line 173: might want to allow usesitemap

emmadickson (approved):
LGTM we'll need to change the tests for the pr I just pushed to

README.md
ikreymer: Eh, maybe better to avoid so many variations?

PR #69, opened 2021-07-20
Interactive Profiles + bug fixes

### LINKED ISSUE #67: Invalid URLs in seeds file
Hi,
crawl doesn't start if there is even one 'Invalid URL' in seeds file, for example 

yle.fi 
instead of 
https://yle.fi

Crawler should jump over invalid URLs, log them in errors and continue with valid URLs?

ikreymer:
- Initial support for creating profiles interactively via devtools UI
- 
* Misc improvements:
- Bug fixes: ensure screencasting errors are caught and do not interrupt crawl
- Seed validation: print invalid seed urls and exit (fixes #67)
- Optimization: don't extract links if seed depth is already at maximum
- Simplify link extraction logic

* Update to browsertrix-behaviors 0.2.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,13 +319,13 @@ will start a crawl with 3 workers, and show the screen of each of the workers fr
 Browsertrix Crawler also includes a way to use existing browser profiles when running a crawl. This allows pre-configuring the browser, such as by logging in
 to certain sites or setting other settings, and running a crawl exactly with those settings. By creating a logged in profile, the actual login credentials are not included in the crawl, only (temporary) session cookies.
 
-Browsertrix Crawler currently includes a script to login to a single website with supplied credentials and then save the profile.
-It can also take a screenshot so you can check if the login succeeded. The `--url` parameter should specify the URL of a login page.
+Browsertrix Crawler includes a script to login to a single website with supplied credentials and then save the profile, as well as a new 'interactive' profile creation mode.
+The script profile creation system also take a screenshot so you can check if the login succeeded. The `--url` parameter should specify the URL of a login page.
 
 For example, to create a profile logged in to Twitter, you can run:
 
 ```bash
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -319,13 +319,13 @@ will start a crawl with 3 workers, and show the screen of each of the workers fr
 Browsertrix Crawler also includes a way to use existing browser profiles when running a crawl. This allows pre-configuring the browser, such as by logging in
 to certain sites or setting other settings, and running a crawl exactly with those settings. By creating a logged in profile, the actual login credentials are not included in the crawl, only (temporary) session cookies.
 
-Browsertrix Crawler currently includes a script to login to a single website with supplied credentials and then save the profile.
-It can also take a screenshot so you can check if the login succeeded. The `--url` parameter should specify the URL of a login page.
+Browsertrix Crawler includes a script to login to a single website with supplied credentials and then save the profile, as well as a new 'interactive' profile creation mode.
```
emmadickson, line 322: much better wording

crawler.js
```diff
@@ -405,14 +409,30 @@ class Crawler {
   }
 
   async extractLinks(page, seedId, depth, selector = "a[href]") {
-    let results = [];
+    const results = [];
```
emmadickson, line 412: should this be a const if we're pushing results to it? would let me more correct?

create-login-profile.js
```diff
@@ -36,6 +44,10 @@ function cliOpts() {
       type: "boolean",
       default: false,
     },
+
+    "interactive": {
+      describe: "Start in interactive mode!",
```
emmadickson, line 49: we could have the default be false for consistency

crawler.js
ikreymer: the way JS works, adding to the array is still ok for const, it just won't allow re-assigning to the results objects.

create-login-profile.js
ikreymer: ok, will add

emmadickson (approved):
LGTM

PR #77, opened 2021-08-05
Revise docker syntax for screencast examples

rlskoeser:
Specify port binding option as a parameter of `docker run` instead of within the `crawl` command

--- 

I couldn't get the screencast syntax to work as written — I had to move the port binding to earlier in the command; once I did, it worked great. I'm running this with `Docker version 20.10.7, build f0df350` in case that matters.


ikreymer:
Thanks! That was indeed a typo on our end, thanks for catching that!

PR #78, opened 2021-08-06
State Save + Restore State from Config

ikreymer:
Support ability to save crawl state (queued, pending, done urls) to the crawl config `state` variable.
When starting a crawl, the `state` dict is read and state is populated with this data.

Override default state handling with custom implementation, supporting in-memory or redis-based state (requires using a fork of puppeteer-cluster which supports async crawl queue)

Improved interrupt handling:
- on first interrupt, gracefully wait for pending pages to finish, then save state before exit (pending array will be empty).
- on second interrupt, immediately close browsers and save state as is, including currently  in-progress pages as part of 'pending' state

- crawl state saved to unique filename `crawl-<timestamp-with-millis>-<hostname>.yaml` in `./crawls` dir (and can be added as part of WACZ later). New crawl can be restarted with this yaml as the crawl config.


Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
     && curl -sL https://deb.nodesource.com/setup_16.x -o /tmp/nodesource_setup.sh && bash /tmp/nodesource_setup.sh \
 	&& apt-get update -y && apt-get install -qqy nodejs yarn \
     && curl https://bootstrap.pypa.io/get-pip.py | python3.8 \
-    && pip install -U setuptools
+    && pip install 'setuptools<58.0'
```
emmadickson, line 24: 👍

Dockerfile
```diff
@@ -36,9 +36,8 @@ ENV PROXY_HOST=localhost \
     BROWSER_VERSION=${BROWSER_VERSION} \
     BROWSER_BIN=${BROWSER_BIN}
 
-COPY --from=browser /tmp/*.deb /deb/
-COPY --from=browser /app/libpepflashplayer.so /app/libpepflashplayer.so
```
emmadickson, line 40: Why do we no longer need the flashplayer?

crawler.js
```diff
@@ -13,14 +13,16 @@ const HTTP_AGENT = require("http").Agent();
...
 const AbortController = require("abort-controller");
 const Sitemapper = require("sitemapper");
 const { v4: uuidv4 } = require("uuid");
 const Redis = require("ioredis");
+const yaml = require("js-yaml");
 
 const warcio = require("warcio");
 
-const behaviors = fs.readFileSync("/app/node_modules/browsertrix-behaviors/dist/behaviors.js", "utf-8");
+const behaviors = fs.readFileSync(path.join(__dirname, "node_modules", "browsertrix-behaviors", "dist", "behaviors.js"), {encoding: "utf8"});
```
emmadickson, line 25: 👍

package.json
```diff
@@ -14,7 +14,7 @@
     "ioredis": "^4.27.1",
     "js-yaml": "^4.1.0",
     "node-fetch": "^2.6.1",
-    "puppeteer-cluster": "^0.22.0",
+    "puppeteer-cluster": "github:ikreymer/puppeteer-cluster#async-job-queue",
```
emmadickson, line 17: I haven't seen this way of declaring a version before what does this do?

emmadickson (commented):
Could you add an example of using the flags to the readme? otherwise lgtm

package.json
ikreymer: Installs from a particular branch on github rather than from npm!

Issue #80, opened 2021-08-13
More blocking rules examples in README.md


billsargent:
Could you provide a couple more examples of blockingRules on the README? 

I'm moving away from httrack and I'm having trouble figuring out how to block urls like https://www.dummyurl.com/gallery/* 

I've tried regex and just plain simple urls in a yaml...

```
blockRules:
- url: dummyurl.com/gallery/
  type: block
- url: /^https://www\.dummyurl\.com/gallery/[a-zA-Z]+$/i
  type: block
```
 
 But all I get after a few minutes of running the container is 
 
```
/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/assert.js:26
        throw new Error(message);.361
              ^

Error: Request is already handled! 
    at Object.assert (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/assert.js:26:15)
    at HTTPRequest.abort (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/HTTPRequest.js:308:21)ihavenreborn.com/auth/login
    at BlockRules.handleRequest (/app/util/blockrules.js:97:17)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /app/util/blockrules.js:70:9
```
 
I'm positive this is something very basic that I'm not understanding. Any help would be appreciated.

ikreymer:
Thanks for reporting, that is an odd error message, will look into that.

Since standard YAML doesn't support regexes, the rules should just be strings. Try:

```
blockRules:
- url: dummyurl.com/gallery/
  type: block
- url: "https://www\.dummyurl\.com/gallery/[a-zA-Z]+$"
  type: block
```

If you run with `--logging stats,debug`, you should also get a printout of how the blockRules have been parsed in JS.

billsargent:
Is this a problem with the host server? 

```
oot@bill-HP-Pavilion-15-Notebook-PC:/home/bill# cat crawls/block.yaml | docker run -i -p 9037:9037 -v $PWD/crawls:/crawls/ webrecorder/browsertrix-crawler crawl --profile /crawls/profiles/profile.tar.gz --url https://www.dummyurl.com/ --screencastPort 9037 --behaviors autoscroll,autoplay,autofetch,siteSpecific --scopeType host --workers 2 --newContext window --text  --combineWarc true --collection testcollection --waitUntil load,networkidle0 --timeout 300 --config stdin --logging stats,debug
[...]

PR #81, opened 2021-08-14
Page Resource Block Rules Avoid Duplicate Handlers + Ignore top-level pages + README update (0.4.4)

ikreymer:
Avoid adding duplicate interception handlers on same page object, which caused errors indicated in #80

Update README to clarify difference between page resource block rules (blockRules) and seed/crawl scope rules.

Update docs to indicate param renamed to `scopeType` instead of `type`.

Block Rules do not apply to top-level pages (causes page load errors), instead warning printed to use scope rules if intent is to exclude full pages.

Also: update `page` scope to really be a single page (was `none`) before while page + hashtags scope renamed to `page-spa`

Ensure current working directory exists by creating it in case of a custom `--cwd` points to a custom path.

bump version to 0.4.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,38 +218,84 @@ Certain settings such scope type, scope includes and excludes, and depth can als
 seeds:
   - url: https://webrecorder.net/
     depth: 1
-    type: "prefix"
+    scopeType: "prefix"
 ```
 
-### Scope Types
+### Crawl Scope -- Configuring Pages Included or Excluded from a Crawl
 
 The crawl scope can be configured globally for all seeds, or customized per seed, by specifying the `--scopeType` command-line option or setting the `type` property for each seed.
 
-The scope controls which linked pages are also included in the crawl.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -218,38 +218,84 @@ Certain settings such scope type, scope includes and excludes, and depth can als
 seeds:
   - url: https://webrecorder.net/
     depth: 1
-    type: "prefix"
+    scopeType: "prefix"
```
emmadickson, line 221: I like this rename

util/seeds.js
```diff
@@ -78,10 +82,6 @@ class ScopedSeed
       include = [/.*/];
       break;
```
emmadickson, line 84: should we be removing the none case?

emmadickson (approved):
Looks good to me. The readme update is really clear

util/seeds.js
ikreymer: yeah, wasn't very clear before, now updated in README and the error message to print the valid values for scopeType.

PR #96, opened 2021-12-04
feat/ Permit custom chrome flags

CreativeCactus:


ikreymer:
Thanks! Perhaps should also support via cmdline args, but maybe this is fine for advanced use cases for now.

PR #155, opened 2022-08-16
Exclude example needs protocol

edsu:
I think that the exclude will not match unless you have the protocol portion of the URL?

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,8 +324,8 @@ seeds:
   - url: https://example.com/startpage.html
     scopeType: "host"
     exclude:
-      - example.com/skip.*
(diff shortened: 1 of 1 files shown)


ikreymer:
Hm, I think it should still work, I believe its gets evaluated on a URL as:
`new RegExp("example.com").exec("https://example.com/")` which does match?

Can probably add another test to verify, and updating the example doesn't hurt, but it shouldn't be needed, i don't think..

edsu:
Thanks, you're right! I was trying to track down why a particular exclusion wasn't working and I thought I had found the reason.

PR #171, opened 2022-09-25
Run in Docker as User

### LINKED ISSUE #170: Run crawl as current user instead of as root
Currently, the crawler runs as root when running in Docker (though other container runtimes, like podman, use the current user).

Perhaps should look into instead creating a user based on the uid/gid of the `crawls` directory, and creating a user automatically,
very similar to what we have in:
https://github.com/webrecorder/pywb/blob/main/docker-entrypoint.sh

edsu:
This follows a [similar pattern]( https://github.com/webrecorder/pywb/blob/main/docker-entrypoint.sh) to pywb to ensure that the crawler is run as the user that owns the crawls directory. This will make sure that the files that are created will be owned by that user.

I've updated `create-login-profile.js` to use `/crawls/profiles/` instead of `/output/` as the default location to write the profile to. This simplifies the permissions issue introduced by this change, since `docker-entrypoint.sh` doesn't need to look for it. If the directory doesn't exist it will create it.

I've also updated the docs to reflect this change to profile creation, and noticed a small bug in the command line argument ordering for the profile extension example (I think that `--profile` needs to appear after `create-login-profile`).

It's possible that this could impact code that is using browsertrix-crawler, like browsertrix-cloud perhaps?

Closes #170

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,5 +40,8 @@ RUN ln -s /app/main.js /usr/bin/crawl; ln -s /app/create-login-profile.js /usr/b
 
 WORKDIR /crawls
 
+ADD docker-entrypoint.sh /docker-entrypoint.sh
+ENTRYPOINT ["/docker-entrypoint.sh"]
+
 CMD ["crawl"]
 
--- a/README.md
+++ b/README.md
@@ -550,7 +550,7 @@ The script profile creation system also take a screenshot so you can check if th
 For example, to create a profile logged in to Twitter, you can run:
 
 ```bash
-docker run -v $PWD/crawls/profiles:/output/ -it webrecorder/browsertrix-crawler create-login-profile --url "https://twitter.com/login"
+docker run -v $PWD/crawls/profiles:/crawls/profiles -it webrecorder/browsertrix-crawler create-login-profile --url "https://twitter.com/login"
 ```
 
 The script will then prompt you for login credentials, attempt to login and create a tar.gz file in `./crawls/profiles/profile.tar.gz`.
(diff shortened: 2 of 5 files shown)


edsu:
I was able to create a profile on AppleOS. I haven't had a chance to test yet on Linux since I only have it remotely.

ikreymer:
Thanks! More testing is needed for the profile creation script, which writes to the `/output` directory instead of `/crawls`.
Perhaps it make sense to check `/output` first, then `/crawls`?
Or default to `/crawls/profiles` as the default dir for profiles?

edsu:
@ikreymer thanks yes -- I went with changing the default directory that the profile is being written to.

Dockerfile
```diff
@@ -19,6 +19,7 @@ ENV PROXY_HOST=localhost \
 WORKDIR /app
 
 ADD requirements.txt /app/
+ADD docker-entrypoint.sh /docker-entrypoint.sh
```
ikreymer, line 22: Should move this towards the end, after all the installs, to avoid rebuilding if this needs to change.

edsu: Ok, reordered!

royteeuwen:
When running the current version 0.7.0 with Rancher Desktop, I get the following exception. Any way to fix this?

groupadd: GID '20' already exists
groupmod: group 'archivist' does not exist

We are also looking into using this image on openshift, where you can't run as root. Will this feature enable that?

ikreymer:
@royteeuwen could you open another issue for this, to ensure this doesn't get lost? Do you know what container runtime you're using. openshift i believe uses podman, and we've run the crawler under podman in the past. We may need to adjust the user script, in case the user/group already exist..

ikreymer:
@royteeuwen we now have this tracked via #174

royteeuwen:
Super :)! Marten is a colleague of mine so

edsu:
If you get a chance to try the test docker image please leave a comment on that issue!

PR #173, opened 2022-10-20
Dev/issue 154 adblock

tw4l:


util/blockrules.js
```diff
@@ -225,5 +228,42 @@ class BlockRules
   }
 }
 
-module.exports.BlockRules = BlockRules;
+class AdBlockRules extends BlockRules
```
ikreymer, line 231: I wonder if it might make sense to just merge this to have a single BlockRules?
On one hand, this is cleaner, but on the other, having a single handleRequest() may remove ambiguity about what handles blocking first? I guess could go either way for now. Can also check how Playwright would handle multiple intercepted requests.

Dockerfile
```diff
@@ -27,6 +30,12 @@ ADD package.json /app/
 # to allow forcing rebuilds from this stage
 ARG REBUILD
 
+# Download and format ad host blocklist as JSON
+RUN mkdir -p /tmp/ads && cd /tmp/ads && \
```
ikreymer, line 34: Great! For now, lets keep this here rather than base image, as that way can update the block list more quickly.

Issue #11, opened 2021-02-03
Support Screenshot Creation


ikreymer:
Add option to generate screenshots, as per:
https://github.com/webrecorder/browsertrix/blob/pywb-instance/simple-driver/index.js#L177

Store in WARC records as `urn:screenshot:<timestamp>/<url>` based on previous convention.

To decide: need options for different dimensions (eg. thumbnails?), default dimensions for screenshot?  (1024x768?)

tw4l:
@ikreymer to review PR

PR #188, opened 2022-11-17
Add screenshot functionality

tw4l:
Connected to #11 

Introduces a `--screenshot` CLI option, which takes a comma-separated list of screenshot types: `view,fullPage,thumbnail`.

Thumbnails are the same size in terms of height and width as the other screenshots (1920x1080) to capture standard page layout but are taken as JPEGs at 75% quality, which are significantly smaller in terms of bytes than the pngs.

This PR also simplifies `newContext` by deprecating the command line argument and simply using page context for 1 worker and our custom window context for >1 workers. A deprecation notice has been added making it clear that the `--newContext` option has been retained but any values passed to it will be ignored.

### DIFF
--- a/util/screenshots.js
+++ b/util/screenshots.js
@@ -5,7 +5,27 @@ import * as warcio from "warcio";
 
 // ============================================================================
 
-class Screenshots {
+export const screenshotTypes = {
+  "view": {
+    type: "png",
+    omitBackground: true,
+    fullPage: false
+  },
+  "thumbnail": {
+    type: "jpeg",
+    omitBackground: true,
+    fullPage: false,
+    quality: 75
+  },
+  "fullPage": {
+    type: "png",
+    omitBackground: true,
+    fullPage: false
+  }
+};
+
+
+export class Screenshots {
 
   constructor({page, url, date, directory}) {
     this.page = page;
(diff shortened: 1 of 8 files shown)


util/screenshots.js
```diff
@@ -0,0 +1,69 @@
...
+    this.url = url;
+    this.directory = directory;
+    this.warcName = path.join(this.directory, "screenshots-" + this.id + ".warc.gz");
+    this.date = date ? date : new Date();
+  }
+
+  async take({fullPage = false, imageType = "png"} = {}) {
+    let options = {omitBackground: true, fullPage: fullPage, type: imageType};
+    let screenshotType = fullPage ? "screenshot-fullpage" : "screenshot";
+    if (imageType === "jpeg") {
```
ikreymer, line 26: Rather than the logic here, perhaps we can have just an array of settings that encapsulates all the settings:
```
{"screenshot-fullpage": {"viewport", ..., "quality": ..., }
 "thumbnail": ...
}
```
That way, will also be easy to add custom types in the future if needed?
Could just be in this file for now

util/screenshots.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+class Screenshots {
+
+  constructor({page, id, url, date, directory}) {
+    this.page = page;
+    this.id = id;
+    this.url = url;
+    this.directory = directory;
+    this.warcName = path.join(this.directory, "screenshots-" + this.id + ".warc.gz");
```
ikreymer, line 19: - We should use one WARC file per browsertrix-crawler instance, so that crawls with 1000s of pages don't result in thousands of WARC files.
- Probably should just place the WARC files in the archives/ dir, but with the `screenshots-` prefix, so that they are included in the WACZ

tw4l: I have the screenshots now being written to `archives/screenshots.warc.gz` and removed the extra logic that was previously adding warcs in the `snapshots` dir to the WACZ

util/screenshots.js
tw4l: I added discrete options objects for each screenshot type in the constructor for the class, which I think should be nice and extendable. Makes it a bit easier to grab the screenshot type rather than having to access the keys but accomplishes the same result I think. Let me know if you'd rather I change it to use an array.

util/screenshots.js
```diff
@@ -0,0 +1,84 @@
...
+class Screenshots {
+
+  constructor({page, url, date, directory}) {
+    this.page = page;
+    this.url = url;
+    this.directory = directory;
+    this.warcName = path.join(this.directory, "screenshots.warc.gz");
+    this.date = date ? date : new Date();
+
+    this.screenshotOptions = {
```
ikreymer, line 17: Can we make this an exported array / json blob and just check for the config that way?

eg.

```
export const screenshotTypes = {
"view": ...,
"thumbnail": ...
}
```
and then just check if the passed in values are one of these in params validation

Issue #74, opened 2021-07-27
Improved Crawl Log Data


ikreymer:
Determine what crawl logs should be generated to help debug crawl.
The page list (stored in `pages.jsonl`) already includes info on what pages were visited and when.
Possible additions:
- Page crawl graph data? (seed and crawl depth of each page)
- Behavior state log?
- Page resources? (which pages were each resources loaded from?)

Thinking the crawl graph data per page (seed and depth) and additional logging of behaviors would be most useful. The page resources will of course be available in the cdx.

In future, a pageId may also be added to WARC headers to be able to better map resources->page

iolantheith:
The Information you shared above is great. I have been reading all you shared here. In this you explained everything very well. If i want any further guideline we will contact you here https://newjerseyloteryresults.com/ Thanks

tw4l:
@despens it seems like the main outstanding issue from your comment is that getting TLDs from `pages.jsonl` can be difficult because of the presence of extracted full text, which seems best handled in a separate issue. I'm going to close this issue and we can track that in https://github.com/webrecorder/browsertrix-crawler/issues/203. Thanks!

PR #195, opened 2022-12-19
Improve crawler logging

tw4l:
Connected to https://github.com/webrecorder/browsertrix-crawler/issues/74

Related to https://github.com/webrecorder/browsertrix-cloud/issues/330

This PR improves browsertrix-crawler logging. Changes:

- Add Logger class with methods for info, error, warn, and debug
- Replace `console.log`, `console.warn` and `console.error` with appropriate logger methods
- Log messages as [JSON Lines](https://jsonlines.org/) with the following structure:
    - info - general:
    ```
    {"logLevel":"info","timestamp":"2022-12-21T17:58:36.114Z","context":"general","message":"Seeds","details":[{"url":"https://webrecorder.net/","include":[{}],"exclude":[],"scopeType":"prefix","sitemap":false,"allowHash":false,"maxExtraHops":1,"maxDepth":99999}]}
    ```
    - info - pageGraph:
    ```
    {"logLevel":"info","timestamp":"2022-12-21T17:58:45.601Z","context":"pageGraph","message":"Page graph data for successfully crawled page","details":{"url":"https://webrecorder.net/","seedId":0,"depth":0,"started":"2022-12-21T17:58:36.586Z"}}
    ```
    - info - behavior:
    ```
    {"logLevel":"info","timestamp":"2022-12-21T17:59:42.914Z","context":"behavior","message":"Behavior line","details":{"state":{"tweets":1,"images":3,"videos":0},"msg":"Capturing Tweet: https://twitter.com/DaleLore/status/1600896285124399104?cxt=HHwWgMCo0Y-sw7csAAAA"}}
    ```
    - error:
    ```
    {"logLevel":"error","timestamp":"2022-12-21T18:00:01.096Z","context":"general","message":"Failed to load resource: the server responded with a status of 404 (Not Found)","details":{"location":{"url":"https://webrecorder.net/2021/01/04/assets/favicon.ico"}}}
    ```
    - warn:
    ```
    {"logLevel":"warn","timestamp":"2022-12-21T18:07:22.868Z","context":"general","message":"Page Load Failed: https://docnow.io/","details":{"message":"Timeout hit: 180000"}}
    ```
- Remove puppeteer-cluster stats
- Log behaviors by default
- Amend argParser to reflect logging changes

Sample log for a page crawl of https://webrecorder.net with `--extraHops 1`: [webrecorder-sample-log-simple.txt](https://github.com/webrecorder/browsertrix-crawler/files/10280291/webrecorder-sample-log-simple.txt)

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -23,6 +23,7 @@ import { ScreenCaster, WSTransport, RedisPubSubTransport } from "./util/screenca
 import { Screenshots } from "./util/screenshots.js";
 import { parseArgs } from "./util/argParser.js";
 import { initRedis } from "./util/redis.js";
+import { Logger } from "./util/logger.js";
 
 import { getBrowserExe, loadProfile, chromeArgs, getDefaultUA, evaluateWithCLI } from "./util/browser.js";
 
@@ -46,6 +47,13 @@ const behaviors = fs.readFileSync(new URL("./node_modules/browsertrix-behaviors/
 // ============================================================================
 export class Crawler {
   constructor() {
+    const res = parseArgs();
+    this.params = res.parsed;
+    this.origConfig = res.origConfig;
+
+    const debugLogging = this.params.logging.includes("debug");
+    this.logger = new Logger(debugLogging);
+
     this.headers = {};
     this.crawlState = null;
 
@@ -63,23 +71,17 @@ export class Crawler {
 
     this.userAgent = "";
 
-    const res = parseArgs();
-    this.params = res.parsed;
-    this.origConfig = res.origConfig;
-
     this.saveStateFiles = [];
     this.lastSaveTime = 0;
-    this.saveStateInterval = this.params.saveStateInterval * 1000;
-
-    this.debugLogging = this.params.logging.includes("debug");
+    this.saveStateInterval = this.params.saveStateInterval * 1000;    
 
     if (this.params.profile) {
-      this.statusLog("With Browser Profile: " + this.params.profile);
+      this.logger.info(`With Browser Profile: ${this.params.profile}"`);
     }
 
     this.emulateDevice = this.params.emulateDevice;
 
-    this.debugLog("Seeds", this.params.scopedSeeds);
+    this.logger.info("Seeds", this.params.scopedSeeds);
 
     this.captureBasePrefix = `http://${process.env.PROXY_HOST}:${process.env.PROXY_PORT}/${this.params.collection}/record`;
     this.capturePrefix = process.env.NO_PROXY ? "" : this.captureBasePrefix + "/id_/";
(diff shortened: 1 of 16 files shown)


ikreymer:
Great! We probably want the actual logs to be on a single line, right?
And then converting all existing logging to use the format (no more random lines of text in the output!)

tw4l:
> Great! We probably want the actual logs to be on a single line, right?

Yup! Printing each json-l object on its own line is just a matter of passing an option to `util.inspect`, which the latest commit shows :)

Turning my attention to those random lines now - for the child_processes we can probably just add some handlers to capture the output and then write it to the logs via the logger object.

ikreymer:
Wonder if we should readd the puppeteer cluster stats, or at least something similar, with maybe a context of "crawlState" that would include some of the stats that it printed, mainly: crawled pages, total pages, list of current pages?
This is obtainable from the crawl state object, similar to what puppeteer cluster does. The other data there (cpu, time  estimates) is not that accurate so probably don't need to include that.
Do we want to be able to filter by the different types of info here? Or make another level, such as 'verbose' that would include this? Hm

ikreymer:
Some other things to decide also, what should be the default for command-line usage, vs the default for running in browsertrix cloud. For example, the crawlState maybe we want to have printed periodically (like puppeteer-cluster) when running via command-line, but may be redundant when running in the cloud since that data will be available via the UI..
Also wondering about JS errors, and we should treat them separately (eg. under verbose level), as user may not necessarily want to see them..

tw4l:
Hm, good points! It seems like maybe the simplest thing to do would be to add `stats` (new implementation per @ikreymer 's note above without CPU and time estimate and with `crawlState` context) and `jserrors` back to the `--logging` CLI open, and include `stats` by default.

That keeps it more similar to the previous API and seems more configurable than a single `verbose` option - e.g. Browsertrix Cloud can simply omit the `stats` option.

tw4l:
Updated logging examples.

Domain crawl of https://bitarchivist.net, default `--logging` options, with new crawl statistics implementation:
[logging-stats-default.txt](https://github.com/webrecorder/browsertrix-crawler/files/10386337/logging-stats-default.txt)

Domain crawl of https://bitarchivist.net, with `--logging jserrors`:
[logging-jserrors-no-stats.txt](https://github.com/webrecorder/browsertrix-crawler/files/10386343/logging-jserrors-no-stats.txt)

tw4l:
One last detail: the outputs of thrown `Error`s won't be formatted as JSON as they break execution.

@ikreymer what would you think about replacing the few thrown `Error`s with a method to log the error message and then quit execution - perhaps `Logger.fatalError()`?

tw4l:
I implemented `Logger.fatal()` in https://github.com/webrecorder/browsertrix-crawler/pull/195/commits/025625b11ece8ec1a3a5b69dcaad14c649b0c673. There might be room for improvement eventually with trying to quit more gracefully, but this at least ensures that all stdout should be formatted in JSON and gives the user a clear idea why a crawl failed.

ikreymer (approved):
Looks great, the new fatal logging before exit is a great to have as well1

PR #197, opened 2022-12-31
Use VNC for headful profile creation

### LINKED ISSUE #194: Use VNC for interactive profile creation
Instead of the current debug tools based screencast, actually use VNC over websocket (via noVNC) to support interactive profile creation. (The devtools screencast is a bit hacky and only works in Chromium-based browsers).
Probably simplest is just to add x11vnc + websockify to the image, and use vnc_lite.html to provide VNC for profile browsing.

ikreymer:
- Adds @novnc/novnc package
- For profile creation, serves novnc_lite.html (modified to support reconnects) with core vnc lib when creating headful profile
- Add x11vnc for vnc
- closes #194

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG BROWSER_VERSION=105
 FROM ${BROWSER_IMAGE_BASE}:${BROWSER_VERSION}
 
 # TODO: Move this into base image
-RUN apt-get update && apt-get install -y jq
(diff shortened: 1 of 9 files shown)


tw4l (approved):
Looks great! Tested and working with both Chromium and Brave, and working much better in interactive mode with Brave settings than the old implementation.

PR #198, opened 2023-01-03
Remove dead pywb configuration

edsu:
This PR originally addressed the need to turn on autofetch in pywb because I noticed it was commented out, and srcset images were not being archived as part of a crawl I was doing.

But after further conversation I see now that failure was the result of my browsertrix-crawler configuration which specified behaviors, but not `autofetch`--effectively turning it off.  browsertrix-crawler uses [browsertrix-behaviors](https://github.com/webrecorder/browsertrix-behaviors)' autofetch, which is different from pywb's, which will perhaps one day replace.

I removed the dead/commented out portions of the default pywb configuration to hopefully prevent future confusion.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,5 +1,4 @@
 debug: true
-
  
 proxy:
   coll: ${COLL}
@@ -14,10 +13,6 @@ recorder:
   cache: always
   rollover_size: ${ROLLOVER_SIZE}
   enable_put_custom_record: true
-
-#autoindex: 10
-
-#enable_auto_fetch: true
 
 collections:
   live: $live


edsu:
via @ikreymer in Discord:

browsertrix-crawler's use of [behaviors](https://github.com/webrecorder/browsertrix-behaviors) should use autofetch by default instead of relying on webrecorder's separate autofetch implementation.

edsu:
I see that `autoplay,autofetch,siteSpecific` are the default behaviors, and that I had inadvertently turned off `autofetch` in my use of browsertrix-crawler. I will update this PR just to remove the commented out code, to hopefully prevent future confusion.

PR #201, opened 2023-01-16
Add requests[socks] python dependency

kuechensofa:
Add requests[socks] python dependency to enable SOCKS proxy support for pywb inside the docker container. This fixes a bug with a missing dependency when trying to use a socks proxy with the browsertrix-crawler.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 pywb>=2.7.2
 uwsgi
 wacz>=0.4.6
+requests[socks]


ikreymer:
Sure! Do you mind sharing more about your use case, curious how you're using SOCKS with Browsertrix Crawler?

kuechensofa:
> Sure! Do you mind sharing more about your use case, curious how you're using SOCKS with Browsertrix Crawler?

Thanks for merging @ikreymer! I'm using SOCKS with Browsertrix Crawler to route the Browsertrix Crawler traffic through Tor by setting up another Docker container which acts as a Tor Proxy.

PR #205, opened 2023-01-22
crawl state: add getPendingList() to return pending state from either…

ikreymer:
… memory or redis crawl state, fix stats logging with redis state

logging: check if data object is an error, log fields from error. Convert missing console.* to new logger


tw4l (approved):
LGTM! Thanks for catching those last remaining console.logs too

PR #211, opened 2023-01-23
behaviors: don't run behaviors in iframes that are about:blank or are…

### LINKED ISSUE #210: Smarter behavior injection
Currently, behaviors are injected into all frames, mostly for autoplay and autofetch background behaviors.
However, this can result in errors and unnecessary waiting for behaviors in frames that don't need them.
Instead, only inject behavior into main frame and frames that are not ads, using the ad block checking mechanism.

Potentially, also look at CDP [Page.adFrameType](https://chromedevtools.github.io/devtools-protocol/tot/Page/#type-AdFrameType](Page.adFrameType), though currently this may not be working yet.

ikreymer:
… from an ad-host (even if ad-blocking is not disabled)

fixes #210

Possibly could add more logging of whether behaviors are injected or skipped

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -448,7 +448,7 @@ export class Crawler {
       return false;
     }
 
-    return await this.adBlockRules.shouldBlock(null, url);
+    return !!(await this.adBlockRules.shouldBlock(null, url));
   }
 
   async createWARCInfo(filename) {
--- a/crawler.js
+++ b/crawler.js
@@ -610,9 +610,7 @@ export class Crawler {
 
     await this.initPages();
 
-    //if (this.params.blockAds) {
     this.adBlockRules = new AdBlockRules(this.captureBasePrefix, this.params.adBlockMessage, this.logger);
-    //}
 
     if (this.params.blockRules && this.params.blockRules.length) {
       this.blockRules = new BlockRules(this.params.blockRules, this.captureBasePrefix, this.params.blockMessage, this.logger);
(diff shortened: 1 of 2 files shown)


crawler.js
```diff
@@ -592,9 +610,9 @@ export class Crawler {
 
     await this.initPages();
 
-    if (this.params.blockAds) {
-      this.adBlockRules = new AdBlockRules(this.captureBasePrefix, this.params.adBlockMessage, this.logger);
-    }
+    //if (this.params.blockAds) {
```
tw4l, line 613: Should we just remove this conditional instead of commenting it out?

crawler.js
```diff
@@ -433,6 +437,20 @@ export class Crawler {
...
+      return true;
+    }
+
+    const url = frame.url();
+
+    if (url === "about:blank") {
+      return false;
+    }
+
+    return await this.adBlockRules.shouldBlock(null, url);
```
tw4l, line 451: I think this will return `null` or `"advertisement"` rather than a boolean - does that work as expected in the filter?

tw4l (commented):
A few small questions here but overall looks good!

crawler.js
ikreymer: Yes

crawler.js
ikreymer, line 451:
```suggestion
    return !!(await this.adBlockRules.shouldBlock(null, url));
```

ikreymer: Should just make it a bool to be safe

Issue #214, opened 2023-01-25
Remove puppeteer-cluster


tw4l:
Related to #8 

The first step of switching to Playwright will be to remove [puppeteer-cluster](https://github.com/thomasdondorf/puppeteer-cluster) (keeping Puppeteer at this stage).

ikreymer:
Fixed by #219

PR #219, opened 2023-02-02
Remove puppeteer-cluster

tw4l:
Connected to #214


util/worker.js
```diff
@@ -0,0 +1,291 @@
+import { EventEmitter } from "node:events";
+import PQueue from "p-queue";
+import puppeteer from "puppeteer-core";
+
+import { Logger } from "./logger.js";
+
+const logger = new Logger();
+
+export const CONCURRENCY_PAGE = "page";
```
ikreymer, line 9: We can probably just get ride of concurrency options, and just always use the window concurrency, eg. open a new window

tw4l: Done

tw4l:
Looks like latest changes are leading to a failure to write cdx index: https://github.com/webrecorder/browsertrix-crawler/actions/runs/4319906519/jobs/7539596680#step:6:179

PR #229, opened 2023-02-23
Logging and Behavior Tweaks

ikreymer:
- Ensure `page` is included in all logging details
- Update logging messages to be a single string, with variables added in the details
- Always wait for all pending wait requests to finish (unless counter <0)
- Don't set puppeteer-cluster timeout (prep for removing puppeeteer-cluster)
- Add behaviorTimeout to running behaviors in crawler, in addition to in behaviors themselves.
- Add logging for behavior start, finish and timeout
- Move writeStats() logging to beginning of each page as well as at the end, to avoid confusion about pending pages.
- deps: bump browsertrix-behaviors to 0.4.2
- version: bump to 0.8.1

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -432,7 +432,7 @@ export class Crawler {
           this.logger.info("Skipping behaviors for non-HTML page", logDetails, "behavior");
         } else {
           const behaviorTimeout = this.params.behaviorTimeout / 1000;
-          this.logger.info("Behaviors started", {...logDetails, behaviorTimeout}, "behavior");
+          this.logger.info("Behaviors started", {behaviorTimeout, ...logDetails}, "behavior");
           const res = await Promise.race([
             this.sleep(behaviorTimeout),
             Promise.allSettled(
@@ -442,7 +442,7 @@ export class Crawler {
             )
           ]);
           if (res && res.length) {
-            this.logger.info("Behaviors finished", {...logDetails, finished: res.length}, "behavior");
+            this.logger.info("Behaviors finished", {finished: res.length, ...logDetails}, "behavior");
           } else {
             this.logger.error("Behaviors timed out", logDetails, "behavior");
           }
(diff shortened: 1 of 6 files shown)


util/browser.js
```diff
@@ -143,11 +143,13 @@ export function chromeArgs (proxy, userAgent=null, extraArgs=[]) {
 }
 
 
-export async function evaluateWithCLI(frame, funcString, name = "behaviors") {
+export async function evaluateWithCLI(frame, funcString, logData, contextName) {
   const context = await frame.executionContext();
   const url = frame.url();
 
-  logger.info(`Running ${name}...`, {url});
+  let details = {url, ...logData};
```
tw4l, line 150: This results in lines like:

```
{"logLevel":"info","timestamp":"2023-02-23T22:14:17.355Z","context":"behavior","message":"Run Script Finished","details":{"url":"https://www.bitarchivist.net/","page":"https://bitarchivist.net/"}}
```

We can probably remove the `url` field, since it duplicates `page`?

crawler.js
```diff
@@ -410,45 +429,58 @@ export class Crawler {
...
       return true;
     }
 
     const url = frame.url();
 
+    let res;
+
     if (url === "about:blank") {
-      return false;
+      res = true;
```
tw4l, line 474: I'm wondering about the change in logic here - `about:blank` would result in an empty page in the iframe, no?

tw4l (approved):
Two small comments but overall very nice!

crawler.js
tw4l: Is the idea that the iframe may be loaded via JS so we should run the behavior just in case?

ikreymer: yes, exactly - was trying this out in case it resulted in additional content, but was inconclusive. maybe only ads do this?

util/browser.js
ikreymer: Yeah, for top level frame, it will be the same, but for iframes, the url would be of the iframe.
Perhaps only add if it is different?

ikreymer: And to be clearer, can call it `frameUrl`.

tw4l: with the name change I think it's fine to have both

crawler.js
ikreymer: Reverted back to false for now, let's see if its needed.

PR #231, opened 2023-02-23
Crawl log

ikreymer:
Write crawl log to `{coll}/logs/crawl-{iso-timestamp}.log` from when crawl is started to just before WACZ creation starts (to allow storing log in WACZ).

Still requires py-wacz support to add logs.


tw4l (approved):
Nice! Code looks good and worked well when I tested 🎉

tw4l (commented):
One thought: the timestamp included in the filename has some colons, which are reserved characters in a number of file systems. Might be better to remove the colons or replace them with dashes?

PR #238, opened 2023-03-06
Healthcheck + Logging Fixes + Iframe Filtering Fixes

ikreymer:
- Refactor healthchecker into own class, increment healthcheck for new page load failures
- Reuse pages (upto 5 page loads per page) before creating a new page (similar to old window-concurrency system)
- iframes: ensure blank iframes are filtered out correctly for both behaviors and link extraction
- adblock: add sync isAdUrl() check
- link extraction: set timeout to avoid hanging in link extraction
- logging: set debug logging globally
- deps: bump puppeteer-core to latest

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -3,8 +3,6 @@ import path from "path";
 import fs from "fs";
 import os from "os";
 import fsp from "fs/promises";
-import http from "http";
-import url from "url";
 
 import fetch from "node-fetch";
 import { RedisCrawlState, MemoryCrawlState } from "./util/state.js";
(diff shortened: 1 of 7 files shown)


crawler.js
```diff
@@ -679,6 +668,7 @@ export class Crawler {
       puppeteerOptions: this.puppeteerArgs,
       crawlState: this.crawlState,
       screencaster: this.screencaster,
+      healthChecker: this.healthChecker,
```
tw4l, line 671: Should remove `timeout` here as well, yeah?

tw4l (approved):
One small comment, otherwise looks great. I like the dedicated HealthChecker class, and thanks for fixing the ad block frame issues!

crawler.js
tw4l: (it wouldn't let me select the actual `timeout` line for a comment as it's not in the diff for this PR)

PR #243, opened 2023-03-09
Add timedRun to prevent async operations from hanging

### LINKED ISSUE #239: Add timeouts to all async operations that could potentially hang
To make the crawler more resilient, we should have timeouts on all operations that could potentially hang, this includes any network request or browser communication (even though that we don't expect to hang).

Perhaps most async calls should be wrapped in something like `await timedRun(promise, timeout, "Timeout Message")` which would log the timeout and return null..

tw4l:
Fixes #239 

`timedRun` is not applied to Puppeteer methods likely to time out such as `page.goto()`, as Puppeteer has its own timeouts for these methods, as described here: https://stackoverflow.com/a/58001115.


ikreymer:
Great! Also should wrap the couple other places currently using `Promise.race` directly with timedRun, I think

PR #248, opened 2023-03-13
Add total timeout + limit redis queue retries

ikreymer:


tw4l (approved):
Looks like we may need additional fixes in PRs to come but this makes a big improvement, thanks!

PR #253, opened 2023-03-17
Dev 0.9.0 Beta 1 Work - Playwright Removal + Worker Refactor + Redis State

ikreymer:
Combines previous PR:
- Switch to Playwright
- Refactor Crawl State to always use redis-based crawl state
- Use sorted queue
- PageWorker runs loop per worker, keeps track of cdp per-worker state


tw4l:
Latest changes look great! I've tested with Chrome/Chromium and Brave, and seeing faster crawls with no errors. Was also able to successfully create, save, and load profiles with both browser bases.

tw4l (approved):

PR #254, opened 2023-03-17
Logger cleanup

ikreymer:
Switch logger to a singleton that is imported, instead of creating a new Logger in each file.


tw4l (approved):
LGTM and test crawl went well

PR #257, opened 2023-03-21
Add option for sleep interval after behaviors run

### LINKED ISSUE #131: [Request] Add option for sleep interval between page crawls to avoid captchas/rate limits
Hello!
I'm trying to crawl a huge website which starts asking for captchas after crawling a few hundreds of pages in a short amount of time.
Since setting workers=1 is not enough to avoid hitting the captcha "rate limit", I'm here to ask for the addition of an option to specify a custom sleep interval (e.g. 5 seconds) which makes the crawler do nothing for the specified amount of time before crawling the next page.
[Youtube-dl has a similar option too](https://github.com/ytdl-org/youtube-dl/blob/master/README.md#workarounds), and in my experience it has been useful in other similar circumstances.
Thanks!

tw4l:
Fixes #131 

Very open to ideas for a better name for this arg! Could be `waitAfterPage`? or even just `waitAfter`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,13 +68,12 @@ Browsertrix Crawler includes a number of additional command-line options, explai
       --crawlId, --id                       A user provided ID for this crawl or
                                              crawl configuration (can also be se
                                             t via CRAWL_ID env var)
-                                              [string] [default: "06bf9a4df9f7"]
+                                              [string] [default: "ce75810e6874"]
       --newContext                          Deprecated as of 0.8.0, any values p
                                             assed will be ignored
                                                         [string] [default: null]
-      --waitUntil                           Puppeteer page.goto() condition to w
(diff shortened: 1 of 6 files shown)


ikreymer:
> Fixes #131
> 
> Very open to ideas for a better name for this arg! Could be `waitAfterPage`? or even just `waitAfter`?

Hm, `pageDelay` or `pageExtraDelay`? This is a hard one :)

Separately, wonder if we should rename `--timeout` to `pageLoadTimeout` or something more descriptive and deprecate generic timeout

tw4l:
@ikreymer Updated names and README

ikreymer:
Added some more time-related refactoring, mostly just consolidating settings to seconds, and only converting to ms when needed for API calls

ikreymer (approved):

PR #264, opened 2023-03-24
Ensure crawler can't run out of space with --diskUtilization param

### LINKED ISSUE #242: Ensure Crawler Can Not Run out of Disk Space / Stops at Disk Utilization
In addition to, `--sizeLimit` option, which checks for fixed size, also add a `--diskUtilization=<percent>` threshold that the crawl can not exceed. Part of webrecorder/browsertrix-cloud#579 but specifically focused on the crawler itself.

- If the diskUtilization threshold is reached, the crawl should run the post-crawl steps and exit.
- If the threshold is reached on startup, the crawler should exit (after running the post-crawl steps).
- By default, the threshold should be 90%. if creating WACZ or WARC, should be otherwise 45% (to allow for double write, until we have WACZ streaming write). If both WACZ and WARC, should be 30%.

tw4l:
Fixes #242 

The first commit implements `--diskUtilization` as described in the issue. I found that this stopped a lot of crawls prematurely in local testing even though the total size, even after combining WARCs and generating WACZ, still wouldn't have crossed the disk threshold.

The second commit modifies the routine to keep the disk utilization percentage high (90% by default) but instead projects whether that threshold is likely to be reached based on the size of the current archive directory (x4 if combineWARC and generateWACZ are passed, x2 if only one of them).

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -553,11 +553,14 @@ export class Crawler {
   async checkLimits() {
     let interrupt = false;
 
+    let dir;
+    let size;
+    if (this.params.sizeLimit || this.params.diskUtilization) {
+      dir = path.join(this.collDir, "archive");
+      size = await getDirSize(dir);
+    }
+
     if (this.params.sizeLimit) {
-      const dir = path.join(this.collDir, "archive");
-
-      const size = await getDirSize(dir);
-
       if (size >= this.params.sizeLimit) {
         logger.info(`Size threshold reached ${size} >= ${this.params.sizeLimit}, stopping`);
         interrupt = true;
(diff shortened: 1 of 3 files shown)


util/storage.js
```diff
@@ -148,6 +150,21 @@ export async function getDirSize(dir) {
   return size;
 }
 
+export async function getDiskUsage(path="/") {
+  const exec = util.promisify(child_process.exec);
+  const result = await exec(`df ${path}`);
```
ikreymer, line 155: I wonder if this is the best way to check disk usage... i guess maybe it is, instead of using some node library for it?
I suppose if it works well, can go with this for now..

crawler.js
```diff
@@ -573,6 +576,33 @@ export class Crawler {
...
+      if (usedPercentage >= this.params.diskUtilization) {
+        logger.info(`Disk utilization threshold reached ${usedPercentage}% > ${this.params.diskUtilization}%, stopping`);
+        interrupt = true;
+      }
+
+      // Check that disk usage isn't likely to cross threshold
+      const kbUsed = parseInt(diskUsage["Used"]);
+      const kbTotal = parseInt(diskUsage["1K-blocks"]);
+      let kbArchiveDirSize = Math.floor(size/1024);
+      if (this.params.combineWARC && this.params.generateWACZ) {
```
ikreymer, line 592: Nice, yes, this is a better way to do this!

util/storage.js
tw4l: I looked for node libraries first (like https://www.npmjs.com/package/check-disk-space and https://www.npmjs.com/package/diskusage) but they all seemed to be wrappers for OS tools like `df` or its Windows equivalent so I figured I'd spare us another dependency and just implement it myself.

PR #266, opened 2023-03-28
worker index: set worker index automatically to work with k8s naming

ikreymer:
- if CRAWL_ID env var set to 'crawl-id-name' while hostname is 'crawl-id-name-N' (automatically set via k8s statefulsets), then set starting worker index to N * numWorkers


tw4l (approved):
Looks good and tested and working well with 3 crawler instances, nice work!

PR #267, opened 2023-03-29
Reset locked pending URLs when crawler restarts.

ikreymer:
Currently, for every URL added to the pending set `<crawl-id>:p`, a 'lock' key is created `<crawl-id>:p:<url>` to indicate that this pending URL is still locked, eg. being processed by a crawler.

If a crawler crashes/is restarted however, the lock keys remain in Redis, even though those URLs should no longer be locked.

This PR sets the lock key to the unique id (usually hostname) of the crawler.
Then, on restart, the crawler will clear its own lock keys before resuming the crawl, as they are no longer being processed.
This allows pending URLs to be retried again (or marked as failed) more quickly.

This clearOwnPending() checks for every `<crawl-id>:p:<url>` if the lock is held by the current crawler, and if so, removes it.

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -161,7 +161,7 @@ export class Crawler {
     this.crawlState = new RedisCrawlState(redis, this.params.crawlId, this.maxPageTime, os.hostname());
 
     // clear any pending URLs from this instance
-    await this.crawlState.clearOwnPending();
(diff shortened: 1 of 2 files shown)


tw4l (approved):
Looks good!

PR #273, opened 2023-03-31
Add unit test for sizeLimit

### LINKED ISSUE #258: Test Improvements: Add tests to ensure sizeLimit (and possibly timeLimit) are applied correctly!
We may not currently have a test specifically for sizeLimit (related to #256)
Should add test for this, and possibly timeLimit as well

stavares843:
How to test:

- run the following command

` yarn test check_limits.test.js`

Captura de ecrã 2023-03-31, às 23 00 53

Resolves #258


stavares843:
works locally but fails in CI 😅

PR #275, opened 2023-04-01
Add --maxPageLimit override

ikreymer:
Adding a new `--maxPageLimit` flag as the best solution for webrecorder/browsertrix-cloud#716

The new flag overrides the `--pageLimit/--limit` flag to ensure the limit is no greater than maxPageLimit if both are set.
This allows for the regular `--pageLimit` to be configured by user in user crawl config settings, while the `--maxPageLimit` can be configured by an admin.

Also rename `--limit` -> `--pageLimit` for clarity, keeping `--limit` as alias.

### DIFF
--- a/util/argParser.js
+++ b/util/argParser.js
@@ -68,8 +68,15 @@ class ArgParser {
         type: "number"
       },
 
-      "limit": {
+      "pageLimit": {
+        alias: "limit",
         describe: "Limit crawl to this number of pages",
+        default: 0,
+        type: "number",
+      },
+
+      "maxPageLimit": {
+        describe: "maximum pages to crawl, overriding pageLimit if both are set",
         default: 0,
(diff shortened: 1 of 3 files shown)


util/argParser.js
```diff
@@ -68,12 +68,19 @@ class ArgParser {
...
-      "limit": {
+      "pageLimit": {
+        alias: "limit",
         describe: "Limit crawl to this number of pages",
         default: 0,
         type: "number",
       },
 
+      "maxPageLimit": {
+        describe: "maximum pages to crawl, overriding pageLimit if both are set",
```
tw4l, line 79:
```suggestion
        describe: "Maximum pages to crawl, overriding pageLimit if both are set",
```

Bit of a nit, but the other arg descriptions start with a capital letter

Issue #268, opened 2023-03-29
Support Contextual Information in datapackage.json for WACZ


markpbaggett:
With browsertrix-crawler, a user can use `combineWARC` to write contextual information defined in the `warcinfo` property into the destination warc. When the warc is read, the fields defined in the property can be read.

```
WARC/1.0
WARC-Filename: exhibitstwo_0.warc.gz
WARC-Date: 2023-03-29T13:58:06Z
WARC-Type: warcinfo
WARC-Record-ID: <urn:uuid:0b216a80-253c-4324-9dc7-a0e18e535b66>
Content-Type: application/warc-fields
Content-Length: 712

software: Browsertrix-Crawler 0.8.0-beta.2 (with warcio.js 1.6.2 pywb 2.7.3)
format: WARC File Format 1.0
title: Database of the Smokies
type: collection
operator: Mark Baggett askDI@utk.edu
hostname: www.lib.utk.edu
creator: University of Tennessee, Libraries
description: This warc contains web assets associated with the Database of the Smokies (DOTS). DOTS provided citations to written works about the Great Smoky Mountains National Park and bordering  communities from 1935 to the present day. In 2023, a decision was made to sunset DOTS but provide ongoing access to its contents. While DOTS will be spun down, its web archive will be made available for perpetuity.
```

There does not appear to be a way currently with Browsertrix-Crawler (0.8.0-beta.2) to do a similar thing where one could add contextual information about the contents of a WACZ to the datapackage.json file. The WACZ spec states that a "WACZ file includes all the data that is needed for the rendering archived content as well as [contextual information](https://specs.webrecorder.net/wacz/1.1.1/#dfn-context) required for users to interpret it." It goes further to state that "the datapackage.json SHOULD include properties that allow rendering applications to present the user with [contextual information](https://specs.webrecorder.net/wacz/1.1.1/#dfn-context) about the web archive:" including `title`, `description`, "other properties from the [[FRICTIONLESS-DATA-PACKAGE](https://specs.webrecorder.net/wacz/1.1.1/#bib-frictionless-data-package)] specification such as licenses, version, organization, contributors, email," and "custom properties that do not interfere with pre-existing properties."

Is there a way currently or are there plans to add a method to add contextual information from the config to `datapackage.json`? If not, I think this would be a great addition.

ikreymer:
[...]

PR #276, opened 2023-04-03
Add --title and --description CLI args to write metadata into datapackage.json

tw4l:
Initial implementation of #268 

Multi-word values with spaces must be passed to `--title` and `--description`/`--desc` (alias added for consistency with py-wacz) using quoted strings, e.g.:

```bash
docker run crawler crawl --url https://example.com --collection test-collection --generateWACZ --title "Test title" --description "Lorem ipsum"
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,6 +23,6 @@ if [ "$MY_GID" != "$VOLUME_GID" ] || [ "$MY_UID" != "$VOLUME_UID" ]; then
     cmd="cd $PWD; $@"
     su btrix -c "$cmd"
 else
-    exec $@
+    exec "$@"
 fi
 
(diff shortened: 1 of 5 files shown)


docker-entrypoint.sh
```diff
@@ -23,6 +23,6 @@ if [ "$MY_GID" != "$VOLUME_GID" ] || [ "$MY_UID" != "$VOLUME_UID" ]; then
     cmd="cd $PWD; $@"
     su btrix -c "$cmd"
```
ikreymer, line 24: A quick change to also 
```suggestion
su btrix -c 'cd $PWD; "$@"' -- argv0-ignore "$@"
```

Based on https://unix.stackexchange.com/questions/479587/first-argument-passed-to-wrapper-script-is-ignored/479645#479645

docker-entrypoint.sh
```diff
@@ -23,6 +23,6 @@ if [ "$MY_GID" != "$VOLUME_GID" ] || [ "$MY_UID" != "$VOLUME_UID" ]; then
     cmd="cd $PWD; $@"
```
ikreymer, line 23: No longer needer, per next line change.
```suggestion
```

PR #281, opened 2023-04-05
origin override: add --originOverride source=dest to allow routing wh…

ikreymer:
…ere https://src-host:src-port/path/page.html -> http://dest-host:dest-port/path/page.html where source=https://src-host:src-port and dest=http://dest-host:dest-port


tw4l (approved):

ikreymer:
This is used to support ReproZip Web (https://github.com/reprozip-news-apps/reproserver) / combination with a 'server-side emulator' where certain domains should be routed to an existing archive/emulator, and only the content 'around' it (from other domains) should be archived.

PR #284, opened 2023-04-12
Store done in redis as integer and only save full json in redis for failed pages

### LINKED ISSUE #280: Investigate removing done from Redis
Storing finished pages in Redis uses a lot of memory and may be unnecessary, especially for successfully finished pages. It is in use several places, however, including the crawler statistics, so may require some discussion/investigation.

tw4l:
Fixes #280 

This PR removes the done list from Redis and replaces it with an integer that is incremented each time a page is successfully finished.

It also adds a failed list to Redis and modifies serialization and crawler stats accordingly.

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -912,12 +912,14 @@ export class Crawler {
     const realSize = await this.crawlState.queueSize();
     const pendingList = await this.crawlState.getPendingList();
     const done = await this.crawlState.numDone();
-    const total = realSize + pendingList.length + done;
+    const failed = await this.crawlState.numFailed();
+    const total = realSize + pendingList.length + done + failed;
     const limit = {max: this.pageLimit || 0, hit: this.limitHit};
     const stats = {
       "crawled": done,
       "total": total,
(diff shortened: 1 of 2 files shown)


crawler.js
```diff
@@ -912,12 +912,14 @@ export class Crawler {
     const realSize = await this.crawlState.queueSize();
     const pendingList = await this.crawlState.getPendingList();
     const done = await this.crawlState.numDone();
-    const total = realSize + pendingList.length + done;
+    const failed = await this.crawlState.numFailed();
+    const total = realSize + pendingList.length + done + failed;
```
ikreymer, line 916: Only thought is: do we want `done` to store only succeeded, and not failed, or include the failed ones?
Could go either way, i guess for backwards compatibility maybe slight preference for done = succeeded + failed.
In that case, the `movefailed` should also include the incr command?

tw4l: Change made, good point!

tw4l: yup, sorry!

PR #285, opened 2023-04-12
feat: Add custom behavior injection

lambdahands:


ikreymer:
@lambdahands is it possible to turn off the spacing changes, or, we should try to do them in a separate pass.. (assume this was from ide reformatting)

util/file_reader.js
```diff
@@ -0,0 +1,33 @@
+import fs from "fs";
+import path from "path";
+
+const MAX_DEPTH = 2;
+
+export function collectAllFileSources(fileOrDir, ext = null, depth = 0) {
```
tw4l, line 6: Is the `ext` argument necessary here? AFAICT, it's only being called with ".js" from the crawler. I could see it being useful to be able to specify e.g. `.ts` and `.js` extensions in the future, but possibly that could just be done within the function without needing an argument?

crawler.js
```diff
@@ -394,10 +399,26 @@ export class Crawler {
...
   }
 
+  loadCustomBehaviors(filename) {
+    let funcs = [];
+
+    for (const source of collectAllFileSources(filename, ".js")) {
+      funcs.push(source);
+    }
+
+    return "[" + funcs.join(",") + "]";
```
tw4l, line 419: We may want to test with multiple custom behaviors to make sure this works as expected

tw4l (approved):
Overall looks great, and this will be such a good feature! Left a few comments/suggestions

crawler.js
edsu: I think it only works with the first script found in the directory.

ikreymer: Updated to browsertrix-behaviors 0.5.1 and also added test for loading multiple behaviors at once, should now be working.

ikreymer: (Testing two behaviors that activate on two different sites)

PR #289, opened 2023-04-17
worker: lower wait time, in case where no additional pages remain and…

ikreymer:
… other workers will finish quickly. otherwise, results in a min 10 seconds wait for >1 workers if only one page is encountered.

Seems like simplest solution just to lower this wait time.


tw4l (approved):
Makes sense!

PR #303, opened 2023-05-01
crawl stopping / additional states:

ikreymer:
- adds check for 'isCrawlStopped()' which checks redis key to see if crawl has been stopped externally, and interrupts work loop and prevents crawl from starting on load
- additional crawl states: 'generate-wacz', 'generate-cdx', 'generate-warc', 'uploading-wacz', and 'pending-wait' to indicate when crawl is no longer running but crawler performing work
- addresses part of webrecorder/browsertrix-cloud#263, webrecorder/browsertrix-cloud#637

### DIFF
--- a/util/state.js
+++ b/util/state.js
@@ -212,6 +212,8 @@ return 0;
     return await this.redis.get("crawl-stop") === "1";
   }
 
+  // note: not currently called in crawler, but could be
+  // crawl may be stopped by setting this elsewhere in shared redis
   async stopCrawl() {
     await this.redis.set("crawl-stop", "1");
   }
--- a/util/state.js
+++ b/util/state.js
@@ -208,6 +208,14 @@ return 0;
     return await this.redis.hset(this.crawlSizeKey, this.uid, size);
   }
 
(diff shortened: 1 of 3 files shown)


util/state.js
```diff
@@ -208,6 +208,14 @@ return 0;
     return await this.redis.hset(this.crawlSizeKey, this.uid, size);
   }
 
+  async isCrawlStopped() {
+    return await this.redis.get("crawl-stop") === "1";
+  }
+
+  async stopCrawl() {
+    await this.redis.set("crawl-stop", "1");
+  }
```
tw4l, line 217: I don't see a place where this method is getting called - is the idea that we'll be setting this in redis from Browsertrix Cloud but we just want a method in the crawler to be able to do the same if needed?

tw4l (approved):
LGTM! One comment

util/state.js
ikreymer: Yep, that's gonna be next PR is to add to the operator to set this. maybe should even remove it from here, but left it in for complteness, can add a comment though.

PR #306, opened 2023-05-03
Consolidate wacz error loglines

tw4l:
This PR modifies `awaitProcess` to log stdout and stderr as single lines to consolidate e.g. stacktraces from the subprocesses being called. These messages are only logged if debug is enabled, otherwise just a generic error message is printed to the logs if called for.

In addition, for CDX generation we now check the status code and log a message appropriately.

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -927,7 +927,7 @@ export class Crawler {
 
     return new Promise((resolve) => {
       proc.on("close", (code) => {
-        if (stdout) {
+        if (stdout.length) {
           logger.debug(stdout.join("\n"));
         }
         if (stderr && this.params.logging.includes("debug")) {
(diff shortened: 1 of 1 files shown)


crawler.js
```diff
@@ -909,15 +914,27 @@ export class Crawler {
...
 
     proc.stderr.on("data", (data) => {
-      logger.error(data.toString());
+      stderr.push(data.toString());
     });
+
     return new Promise((resolve) => {
-      proc.on("close", (code) => resolve(code));
+      proc.on("close", (code) => {
+        if (stdout) {
```
ikreymer, line 930:
```suggestion
        if (stdout.length) {
```

crawler.js
```diff
@@ -909,15 +914,27 @@ export class Crawler {
...
+          logger.debug(stdout.join("\n"));
+        }
+        if (stderr && this.params.logging.includes("debug")) {
```
ikreymer, line 933:
```suggestion
        if (stderr.length && this.params.logging.includes("debug")) {
```

PR #312, opened 2023-05-10
Disable Chrome optimization logic

### LINKED ISSUE #311: Crawler often downloads 40-50MB worth of unnecessary Chrome model files
These are not needed for the crawl and are not part of the websites being crawled, but still end up in the WARC files.

Here's a discussion of the problem: https://bugs.chromium.org/p/chromium/issues/detail?id=1311753

The solution is to disable additional features in Chrome:

`--disable-features=OptimizationGuideModelDownloading,OptimizationHintsFetching,OptimizationTargetPrediction,OptimizationHints`

These should probably be added to https://github.com/webrecorder/browsertrix-crawler/blob/main/util/browser.js#L326 (the default Chrome args).

malemburg:
These optimizations can often lead to Chrome downloading large ML models in the background, which then end up in the web crawling archives, even though they don't have anything to do with the crawl.

Fixes #311.

### DIFF
--- a/util/browser.js
+++ b/util/browser.js
@@ -323,7 +323,8 @@ export const defaultArgs = [
   "--disable-extensions",
   // AvoidUnnecessaryBeforeUnloadCheckSync - https://github.com/microsoft/playwright/issues/14047
   // Translate - https://github.com/microsoft/playwright/issues/16126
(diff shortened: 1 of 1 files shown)


malemburg:
Please let me know, if there are any news or change patches needed as well.

stavares843 (commented):
The change seems to do what it says in the commit message and disables the Chrome optimization logic, which is known to download large ML models in the background.
The code changes look simple and safe and only modify the defaultArgs array in the util/browser.js file.
There are no bugs or issues with the code changes that I can see.

PR #316, opened 2023-05-20
Chrome 112 + new headless mode + consistent viewport tweaks

ikreymer:
- update base browser to chrome 112
- headless: switch to using new headless mode available in 112 which is more in sync with headful mode
- viewport: use fixed viewport matching screen dimensions for headless and headful mode
- profiles: fix catching new window message, reopening page in current window in headless mode, don't use in headful
- versions: bump to pywb 2.7.4, update puppeteer-core to (20.2.1)
- bump to 0.10.0-beta.4


util/browser.js
```diff
@@ -34,17 +34,21 @@ export class BaseBrowser
 
     const args = this.chromeArgs(chromeOptions);
 
+    const geom = process.env.GEOMETRY.split("x");
+
+    const defaultViewport = {width: Number(geom[0]), height: Number(geom[1])};
```
tw4l, line 39: Perhaps naive question (I'm not super familiar with the `GEOMETRY` env var in this context, but will this always be set? Should we have a fallback?

tw4l (approved):
LGTM - one comment

util/browser.js
ikreymer: We always set it in the Dockerfile, but I suppose could also add a default here 'just in case', or not set it, if its missing for some reason..

PR #319, opened 2023-05-23
allow adding --include with pre-existing --scopeType values (besides …

### LINKED ISSUE #318: Allow --includes to be added to --scopeType values
Currently, specifying `--includes` overrides the default include set by `--scopeType` if both are provided.
However, it'll simpler to just add the includes on top of the existing scopeType, rather than dropping it.

Eg. currently to set a seed of `https://example.com/prefix/` and set scope to URLs starting with `https://example.com/prefix/` and also add `https://example.org/some/other/path` to the scope, the config would have to be:
`--url https://example.com/prefix/ --include https://example.com/prefix/ --include https://example.org/some/other/path` as setting `--scopeType prefix` would not be supported.
Proposing to instead allow the more straightforward `--url https://example.com/prefix/ --scopeType prefix --include https://example.org/some/other/path`.

The former behavior could still be available by omitting scopeType or setting it to custom `--url https://example.com/prefix/ --include https://example.com/prefix/ --include https://example.org/some/other/path --scopeType custom`

This will simplify solution to webrecorder/browsertrix-cloud#873

ikreymer:
…custom) (fixes #318)

remove warning when --scopeType and --include used together tests: update tests to reflect new semantics of --include + --scopeType

Simplifies 'extra scope URL' in webrecorder/browsertrix-cloud#378

### DIFF
--- a/tests/scopes.test.js
+++ b/tests/scopes.test.js
@@ -186,10 +186,12 @@ seeds:
    - url: https://example.com/subpath/file.html
      scopeType: prefix
 
+   - url: https://example.com/subpath/file.html
+
 include: https://example.com/onlythispath
 `);
 
-  expect(seeds.length).toEqual(3);
+  expect(seeds.length).toEqual(4);
 
   expect(seeds[0].scopeType).toEqual("custom");
   expect(seeds[0].url).toEqual("https://example.com/1");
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Nice!

Issue #307, opened 2023-05-04
Handling of CRAWL_ARGS cannot cope with quoted strings


anjackson:
https://github.com/webrecorder/browsertrix-crawler/blob/f4c4203381522db64c46d540df27651a4bd77c67/util/argParser.js#L388-L390

As discussed in webrecorder/browsertrix-cloud#787 I cannot pass a user agent suffix (or any other argument) that contains a space character via the `CRAWL_ARGS` environment variable.  The code (above) just splits on spaces, so setting `--userAgentSuffix "bl.uk_ldfc_bot/btrix (+ http://www.bl.uk/aboutus/legaldeposit/websites/websites/faqswebmaster/index.html)"` does not work as expected.

Easiest workaround is to drop spaces from the UA suffix, but this means we are not strictly compliant with the policy framework we are currently working in.

PR #322, opened 2023-05-25
Update argParser.js for #307

anjackson:
This change uses the yargs argument parser rather than just splitting the CRAWL_ARGS environment variable.

**WARNING:** This has not been tested yet!


anjackson:
I don't think this was necessary, so I'll close it.

PR #330, opened 2023-06-14
browsertrix as a service

adityaraj-28:


main.js
```diff
@@ -62,6 +63,20 @@ process.on("SIGUSR2", () => {
...
 crawler = new Crawler();
-crawler.run();
 
+const app = express();
+const port = 3000;
+console.log("crawler created");
+app.get("/", (req, res) => {
+  if (crawler) {
+    res.send("Crawl initiated!");
+    crawler.run();
```
edsu, line 73: Every time a client does a GET to `/` it will start a crawl? I believe that the crawler will call `process.exit()` once the crawl completes, which will stop your service?

main.js
```diff
@@ -62,6 +63,20 @@ process.on("SIGUSR2", () => {
 });
 
 crawler = new Crawler();
-crawler.run();
 
+const app = express();
+const port = 3000;
+console.log("crawler created");
+app.get("/", (req, res) => {
```
edsu, line 70: I wonder if this should be a POST instead? Do you really want to trigger a new crawl any time someone does a GET to this URL?

edsu (commented):
You might be interested in looking at https://github.com/webrecorder/browsertrix-crawler which provides a full featured API for running browsertrix-crawler. I don't think this PR will work very well since the service will stop once the crawl is completed, and it will potentially start multiple identical crawls.

edsu:
Hopefully my comments weren't too negative, my apologies if they were interpreted that way @adityaraj-28. I fixed the link in my previous comment which I wanted to point to [browsertrix-cloud](https://github.com/webrecorder/browsertrix-cloud) which has a nice API for controlling crawls. But it does require more resources than the minimal service API you proposed. Unfortunately, managing potentially multiple crawls is kind of resource intensive.

adityaraj-28:
Hi @edsu No it was actually my bad, i realised there were issues in my code and hence i closed the PR. Sorry for communicating first.

Issue #331, opened 2023-06-15
Page not archived after verification check and reload


wvengen:
Certain websites / CDNs have a kind of verification dance before the actual content is shown:
- The URL is loaded, no real page content is shown, except some Javascript.
- This Javascript does an AJAX request with some way to mark a session as ok for viewing.
- After the AJAX request, the page is reloaded using `window.location = (same url as current page)`.
- The new page is loaded, and the server returns the true page content.

It appears that only the first request/response for the page is included in the WACZ (both index and actual WARCs). When viewing screenshots, it can be seen that the web browser does go through the full process. Also, `pages/` contains the full-text of the true page contents, so that confirms that the true content is being loaded.

It looks like browsertrix-crawler only records the first request/response for the URL.

To make this reproducible, I've come up with a minimal verifiable example. Here is a little Flask app that does something like described above, a [docker-compose setup with app to generate the archive](https://github.com/webrecorder/browsertrix-crawler/files/11754745/sample.zip) and [the resulting WACZ](https://github.com/webrecorder/browsertrix-crawler/files/11754757/sample.wacz.zip).

```python
#!/usr/bin/env python3
from flask import Flask, request

app = Flask(__name__)

SCRIPT = '''
var xhr = new XMLHttpRequest();
xhr.addEventListener("loadend", function() {
    document.cookie = 'verified=1; SameSite=strict';
    window.location = "/";
});
xhr.open("POST", "/v", false);
xhr.send();
'''

@app.get("/")
def home():
    verified = request.cookies.get('verified')
    if verified == '1':
        return '<html><head><title>Success</title><body><h1>Success</h1></body></html>'
    else:
        return '<html><head><title>Check</title></head><body><h1>Check</h1><script>' + SCRIPT + '</script></body></html>'

@app.post("/v")
def v():
    return ''

if __name__ == '__main__':
    app.run(host='0.0.0.0', port=5000)
```

Ideally, when generating a WACZ from this page and viewing it with ReplayWeb.page, it would show the _Success_ page. It does show _Success_ as the page title in the page overview, but when viewing the page in ReplayWeb.page, it keeps showing the _Check_ content and redirect all the time.

![image](https://github.com/webrecorder/browsertrix-crawler/assets/503804/3fb6b947-a9d0-4a53-bc9e-3fe2c02eec32)
![image](https://github.com/webrecorder/browsertrix-crawler/assets/503804/8bb96d84-e788-4520-b252-9377f05e9142)

[...]

PR #332, opened 2023-06-15
Allow configuration of deduplication policy

wvengen:
The simplest solution for #331 (whole or partial solution, depending how you look at it), allowing configuration of [pywb's `dedup_policy`](https://pywb.readthedocs.io/en/latest/manual/configuring.html#dedup-options-for-recording).

- [x] Test this actually works.

### DIFF
--- a/util/argParser.js
+++ b/util/argParser.js
@@ -250,6 +250,13 @@ class ArgParser {
         type: "number",
       },
 
+      "dedupPolicy": {
+        describe: "Deduplication policy",
+        default: "skip",
+        type: "string",
+        choices: ["skip", "revisit", "keep"],
+      },
+
       "profile": {
         describe: "Path to tar.gz file which will be extracted and used as the browser profile",
         type: "string",
--- a/README.md
+++ b/README.md
@@ -176,6 +176,8 @@ Options:
                                             econds) after behaviors before movin
                                             g on to next page
                                                            [number] [default: 0]
+      --dedupPolicy                         Deduplication policy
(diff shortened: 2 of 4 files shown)


ikreymer:
One issue however, is that we are planning to move away form pywb for recording (see: https://github.com/webrecorder/browsertrix-crawler/tree/recorder-work) so it may be more difficult to support this exact functionality, though probably should have some sort of dedup options..

util/argParser.js
```diff
@@ -250,6 +250,13 @@ class ArgParser {
         type: "number",
       },
 
+      "dedupPolicy": {
+        describe: "Deduplication policy",
+        default: "skip",
+        type: "string",
+        choices: ["skip", "revisit", "keep"],
```
tw4l, line 257: Might not be a bad idea to add some input validation here so that if a user adds an incorrect value it'll default back to `skip` rather than throwing an error. Otherwise looks great, thank you!

tw4l (approved):

util/argParser.js
tw4l: Going to add in a follow-up commit after merging :)###

ikreymer: yargs will validate the `choices` list so should be good here!

wvengen:
Thank you for including this change! :tada:

For reference: maybe it's good to point to the issue from here (https://github.com/webrecorder/browsertrix-crawler/issues/331#issuecomment-1592933687) and note that I took a safe set of options, which work with redis duplicate detection.

Good to know that a new recording path is being taken, thanks for mentioning. Curious to see how that develops!

PR #342, opened 2023-07-06
Add input validation for dedupPolicy

tw4l:
Connected to #331 

Follow-up to #332

Also updates crawler options in README to include latest options.


tw4l:
Closing because the `choices` parameter already does validation for us! Neat :)

PR #356, opened 2023-08-19
improve crawl stopped check with unified isCrawlRunning() check with …

ikreymer:
…checks both interrupted + redis-based state

handle browser crash -- if getting new page fails after 5 tries, assume browser crashed and exit

Fixes potential 'getting new page' infinite loop with a clear exit (addressing openzim/zimit#204).

Could later add an option to try to restart the browser, although no guarantees there either..


tw4l (approved):
Nice!

Issue #361, opened 2023-08-30
Support saving faviconUrl


ikreymer:
ArchiveWeb.page stores the favicons as part of the pages.jsonl in the `favIconUrl` property.
For consistency, we should also store this if we can determine what the favicon used by the browser is.
(Seems to be no puppeteer API for this, so may have to determine based on parsing html).

ikreymer:
While the favicon isn't directly available over CDP (surprisingly), it is possible to get the actual favicon (without having to look at the HTML) from the `/json` endpoint.

eg testing with:
```
console.log("target", page.target());
const resp = await fetch("http://localhost:9221/json");
console.log("json", await resp.json());
```

prints out:
```
target PageTarget {
  _initializedDeferred: Deferred {},
  _isClosedDeferred: Deferred {},
  _targetId: 'F4CFEA713AEED16B6E0939789386DB89',
  pagePromise: Promise { CDPPage { emitter: [Object], eventsMap: [Map] } }
}
json [
  {
    description: '',
    devtoolsFrontendUrl: '/devtools/inspector.html?ws=localhost:9221/devtools/page/F4CFEA713AEED16B6E0939789386DB89',
    faviconUrl: 'https://www.iana.org/_img/bookmark_icon.ico',
    id: 'F4CFEA713AEED16B6E0939789386DB89',
    title: 'Internet Assigned Numbers Authority',
    type: 'page',
    url: 'https://www.iana.org/',
    webSocketDebuggerUrl: 'ws://localhost:9221/devtools/page/F4CFEA713AEED16B6E0939789386DB89'
  },
  {
    description: '',
    devtoolsFrontendUrl: '/devtools/inspector.html?ws=localhost:9221/devtools/page/64E785D571D351914BE0CE6CB0A345F3',
    id: '64E785D571D351914BE0CE6CB0A345F3',
    title: 'about:blank',
    type: 'page',
    url: 'about:blank',
    webSocketDebuggerUrl: 'ws://localhost:9221/devtools/page/64E785D571D351914BE0CE6CB0A345F3'
  }
]
```

This seems to be a promising way to quickly get the favicon URL.

Chickensoupwithrice:
Resolved by https://github.com/webrecorder/browsertrix-crawler/pull/362

PR #362, opened 2023-08-31
Capture Favicon

Chickensoupwithrice:
Small change to also capture favicon and serialize it as jsonl

Resolves https://github.com/webrecorder/browsertrix-crawler/issues/361

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -431,7 +431,7 @@ self.__bx_behaviors.selectMainBehavior();
     return str;
   }
 
-  async getFavicon(page) {
+  async getFavicon(page, logDetails) {
     const resp = await fetch("http://localhost:9221/json");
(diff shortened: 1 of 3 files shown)


ikreymer (approved):

ikreymer:
Looks good! One suggestion, maybe the test file should be renamed to something like `advanced_crawl_and_favicon` so we remember that the test for favicon is there, or split into two files.. Not sure if I like 'advanced' crawl per se, just as there's so many options that could be considered advanced ;), but can revisit that later.

Chickensoupwithrice:
yeah agreed, I'll rename it. I copied the basic crawl and added a second url so called it advanced lol. maybe it should be `multi_url_and_favicon_test`?

PR #363, opened 2023-08-31
feat: precommit

Chickensoupwithrice:
Add precommit through husky + pre-commit (similar to browsertrix-cloud)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+-#!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn lint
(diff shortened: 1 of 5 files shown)


ikreymer:
@Chickensoupwithrice is the pre-commit hook actually running? Wasn't able to get it to run after `yarn install` and then committed something that failed lint.

Chickensoupwithrice:
Ahh, sorry about that, I had the file name wrong. You also will need to run `yarn prepare` to for the `husky` install

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+-#!/usr/bin/env sh
```
ikreymer, line 1: typo here
```suggestion
#!/usr/bin/env sh
```

ikreymer:
yep, i think that fixed it!

ikreymer (approved):

PR #364, opened 2023-08-31
link extraction optimization: for scopeType page, set depth == extraH…

ikreymer:
For `scopeType: page`, we know there won't be any additional links, unless extraHops is set, so set the depth == extraHops, so the depth check can skip the link extraction.
Add debug logging and bump to 0.10.5


tw4l (approved):
Nice optimization!

PR #370, opened 2023-09-10
various fixes regarding state restart:

ikreymer:
This PR is designed to address a few edge-cases when using crawler in Browsertrix Cloud (>=1.7.0 beta 0)
- Add --restartOnError flag, which tells crawler that it will be restarted if it receives a signal and will not perform post-crawl processes on non-zero exit.
- If a crawl is being stopped gracefully, but is also interrupted via signal (due to size/time/disk limit or other interruption) but is able to complete, it *will* attempt post-crawl processes / exit with success
- If no WARCs are found, but number of pages written >0, assume crawler was restarted, and don't fail!
- use distinct exit code for subsequent interrupt (13) and fatal interrupt (17)

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -310,7 +310,10 @@ export class Crawler {
     try {
       await this.crawl();
       const finished = await this.crawlState.isFinished();
-      if (this.interrupted && !finished) {
+      const stopped = await this.crawlState.isCrawlStopped();
+      if (stopped) {
+        status = "crawl gracefully stopped";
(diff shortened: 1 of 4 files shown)


Chickensoupwithrice (approved):
Tested locally by `kubectl exec` into container and sending `kill -S SIGTERM 1` as well as using `kubectl destroy pod <pod-id>` and in both cases the pod restarts. 

Then tried cancelling the workflow and then killing the pod and it didn't come back as expected.

PR #373, opened 2023-09-12
Add ability to load behaviours from URL

### LINKED ISSUE #368: Load Custom Behaviour from URL
We'd like to be able to point to a URL that contains a javascript file that defines a custom behavior and have `browsertrix-crawler` be able to load and use that behavior. 

### Requirements
Be able to pass a URL to `--customBehaviors` flag which contains a single javascript file containing the custom behavior to be loaded and executed. 

### Future goals
Future goals could include passing an S3 directory, and have `crawler` be able to load all behaviors in that S3 bucket.

Chickensoupwithrice:
Fixes #368 

If you pass multiple `customBehaviors` parameters they're all assumed to be URLs, which is not great but I couldn't think of a use case for mixing local behaviours with URL loaded behaviours?

### DIFF
--- a/tests/url-behavior.test.js
+++ b/tests/url-behavior.test.js
@@ -1,9 +1,9 @@
 import child_process from "child_process";
 
 test("test custom behaviors", async () => {
-  const res = child_process.execSync("docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url --url https://webrecorder.net/ --customBehaviors 'https://raw.githubusercontent.com/webrecorder/browsertrix-behaviors/main/dist/behaviors.js' --scopeType page");
+  const res = child_process.execSync("docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url --url https://webrecorder.net/ --customBehaviors 'https://raw.githubusercontent.com/webrecorder/browsertrix-behaviors/main/dist/behaviors.js' --scopeType page --behaviors \"\"");
 
   const log = res.toString();
 
-  expect(log.indexOf("{\"state\":{\"segments\":1},\"msg\":\"Skipping autoscroll, page seems to not be responsive to scrolling events\",\"page\":\"https://webrecorder.net/\",\"workerid\":0}") > 0).toBe(true);
+  expect(log.indexOf("File downloaded to /app/behaviors/") > 0).toBe(true);
 });
--- a/tests/url-behavior.test.js
+++ b/tests/url-behavior.test.js
@@ -1,0 +1,9 @@
+import child_process from "child_process";
+
+test("test custom behaviors", async () => {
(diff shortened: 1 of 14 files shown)


util/file_reader.js
```diff
@@ -1,8 +1,44 @@
...
+      collectSingleOnlineFile(u);
+    }
+  }
+  return collectAllFileSources("/app/behaviors", ".js");
+}
+
+export function collectSingleOnlineFile(url) {
+  const split = url.split("/");
+  const filename = split[split.length - 1];
+  const file = fs.createWriteStream("/app/behaviors/" + filename);
```
ikreymer, line 32: maybe we just create a temp filename to avoid issues like duplicate filenames, query args in URLs, etc..?

util/file_reader.js
```diff
@@ -1,8 +1,44 @@
...
+  // TODO handle case where file is sent over HTTP?
+  https.get(url, function(response) {
```
ikreymer, line 34: We can use `await fetch()` now in node instead of the old http/https

util/file_reader.js
```diff
@@ -1,8 +1,44 @@
 import fs from "fs";
 import path from "path";
+import https from "https";
 
 const MAX_DEPTH = 2;
 
+export function determineFileSource(fileOrUrl, ext = null) {
+  // Currently assuming if we pass more than one param, they're *all* URLs
```
ikreymer, line 8: hm, i think it should be flexible - if its a URL, we download to temporary file first, then we just load a list of files from disk..

util/file_reader.js
Chickensoupwithrice: yeah i think that makes more sense, you're right

util/file_reader.js
Chickensoupwithrice: We would then need to combine this with loading files / folders that could be passed through the same param as well.

tests/url-behavior.test.js
```diff
@@ -0,0 +1,9 @@
+import child_process from "child_process";
+
+test("test custom behaviors", async () => {
+  const res = child_process.execSync("docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url --url https://webrecorder.net/ --customBehaviors 'https://raw.githubusercontent.com/webrecorder/browsertrix-behaviors/main/dist/behaviors.js' --scopeType page");
+
+  const log = res.toString();
+
+  expect(log.indexOf("{\"state\":{\"segments\":1},\"msg\":\"Skipping autoscroll, page seems to not be responsive to scrolling events\",\"page\":\"https://webrecorder.net/\",\"workerid\":0}") > 0).toBe(true);
```
Chickensoupwithrice, line 8: this test is wrong, since we already include behaviors in crawler by default, it'll show this line in the logs regardless of whether the behavior was downloaded or not...

tw4l:
Closing in favor of https://github.com/webrecorder/browsertrix-crawler/pull/707

PR #377, opened 2023-09-13
logging fixes: avoid duplicate logging for same error

ikreymer:
The goal is to reduce number of `error` logging entries, which show up in Browsertrix Cloud error logs, to only key errors / avoid redundant entries.

- avoid duplicate logging for same error, if logging more specific message and rethrowing exception, set e.detail to "logged" and worker exception handler will not log same error again
- add option to log timeouts as warnings instead of errors
- remove unneed async method in browser, get headers directly
- fix logging in screenshots to include page

### DIFF
--- a/util/worker.js
+++ b/util/worker.js
@@ -167,14 +167,16 @@ export class PageWorker
           this.crawler.crawlPage(opts),
           this.maxPageTime,
           "Page Worker Timeout",
-          {workerid},
+          this.logDetails,
           "worker"
         ),
         this.crashBreak
       ]);
 
     } catch (e) {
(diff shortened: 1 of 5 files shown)


util/worker.js
```diff
@@ -167,14 +167,16 @@ export class PageWorker
...
           "worker"
         ),
         this.crashBreak
       ]);
 
     } catch (e) {
-      logger.error("Worker Exception", {...errJSON(e), ...this.logDetails}, "worker");
+      if (e.message !== "logged") {
+        logger.error("Worker Exception", {...errJSON(e), ...this.logDetails}, "worker");
+      }
```
tw4l, line 179: Nice!

tw4l (approved):

PR #380, opened 2023-09-14
optimize link extraction: (fixes #376)

### LINKED ISSUE #376: More efficient link extraction / link extraction behaviors.
Currently, the link extraction returns the entire array of links, which can be quite large on such pages, as a single response via CDP.
Suspect this may be slowing down the crawling in certain cases.
Instead, we should use the behavior system callback to iterate over each link and add them that way. This can be done similar to how logging is handled, and can be further customized with custom link extraction behaviors.

ikreymer:
- dedup URLs in browser first
- don't return entire list of URLs, process one-at-a-time via registered callback function
- add exposeFunction per page in setupPage, then register 'addLink' callback for each pages' handler
- optimize addqueue: atomically check if already at max urls and if url already seen in one redis call, as links can be added async
- better logging: log rejected promises for link extraction
- browser: set protocolTimeout to 0 to avoid CDP timeout errors

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -4,7 +4,7 @@ import fs from "fs";
 import os from "os";
 import fsp from "fs/promises";
 
-import { RedisCrawlState, LoadState } from "./util/state.js";
+import { RedisCrawlState, LoadState, QueueState } from "./util/state.js";
 import Sitemapper from "sitemapper";
 import { v4 as uuidv4 } from "uuid";
 import yaml from "js-yaml";
@@ -1305,7 +1305,7 @@ self.__bx_behaviors.selectMainBehavior();
         const {url, isOOS} = res;
 
         if (url) {
-          await this.queueUrl(seedId, url, depth, isOOS ? newExtraHops : extraHops);
+          await this.queueUrl(seedId, url, depth, isOOS ? newExtraHops : extraHops, logDetails);
         }
       }
     } catch (e) {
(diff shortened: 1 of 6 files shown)


util/state.js
```diff
@@ -67,9 +67,17 @@ export class RedisCrawlState
...
       numberOfKeys: 3,
       lua: `
-redis.call('sadd', KEYS[3], ARGV[1]);
+local size = redis.call('scard', KEYS[3]);
+local limit = tonumber(ARGV[4]);
+if limit > 0 and size >= limit then
+  return 0;
+end
+if redis.call('sadd', KEYS[3], ARGV[1]) == 0 then
+  return 0;
```
tw4l, line 76: This is causing the limit to be set as hit prematurely if we encounter a duplicate URL

ikreymer: Fixed! Added different return values

crawler.js
```diff
@@ -1333,19 +1333,29 @@ self.__bx_behaviors.selectMainBehavior();
...
+    case QueueState.LIMIT_HIT:
+      logger.debug("Not queued page url, at page limit", {url, ...logDetails}, "links");
       this.limitHit = true;
       return false;
-    } else {
-      logger.debug(`Queued url ${url}`);
+
+    case QueueState.DUPE_URL:
+      logger.debug("Not queued page url, already seen", {url, ...logDetails}, "links");
+      return false;
```
tw4l, line 1355: Cases should be indented by 2

tw4l (approved):

crawler.js
ikreymer: Our lint settings actually format it this way currently, maybe something to take a look at in a separate pass.

tw4l: Ah okay, noted!

ikreymer:
Also added additional test which ensures exact page limit is reached, would've caught the early limit issue.

PR #381, opened 2023-09-14
behavior logging tweaks, add netIdle

ikreymer:
Slight improvement to behavior logging, indicate if one of the frames' promises has failed.

bump to latest browsertrix-behaviors


Chickensoupwithrice (approved):
fairly small change, detection of iframe seems reasonable

PR #383, opened 2023-09-15
Update tldextract cache for pywb during build

vnznznz:
Pywb is unable to function when it's unable to refresh its tld cache. Unfortunately that's not so straightforward in the environment where I have to deploy it. With this change I'm able to run crawls without any internet access except for a socks proxy.


tw4l (approved):
Nice change, thanks for the contribution!

PR #385, opened 2023-09-18
Error handling fixes to avoid crawler getting stuck.

ikreymer:
Logging and error handling improvements, in particular to avoid crawler getting stuck.
- Catch correct event when page crashes, `page.on('error'`) not `page.on('crash')`
- Also catch browser crash/disconnect with `browser.on('disconnected')` and immediately interrupt, but allow the post-crawl state to run to await for pending response to be written to WARC. (Or, could also change logger.fatal() to wait for that as well).
- Add timeout for page.teardown which calls screencaster and `await cdp.send("Page.stopScreencast");`, which can timeout as well (possibly if page crashed).
- Avoid duplicate logging of crash events where possible.
- Switch logging format to put timestamp first.
- Just to be safe: If new page is on a different origin than previous page, always open a new page, don't reuse previous one.

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -820,6 +820,10 @@ self.__bx_behaviors.selectMainBehavior();
         proxy: !process.env.NO_PROXY,
         userAgent: this.emulateDevice.userAgent,
         extraArgs: this.extraChromeArgs()
+      },
+      ondisconnect: (err) => {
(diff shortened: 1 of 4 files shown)


tw4l (approved):
Nice catches! Good to see these improvements

PR #397, opened 2023-09-26
Store crawler start and end times in Redis lists

### LINKED ISSUE #378: Track start and end time of each crawler session in Redis
Connected to [#1118](https://github.com/webrecorder/browsertrix-cloud/issues/1118)

Will be used in Browsertrix Cloud to calculate execution minutes for billing.

tw4l:
Fixes #378


ikreymer:
For consistency, the process.exit() in logger.fatal() should also set the end time.
Can probably do it w/o await to avoid turning that into an async function..

Issue #189, opened 2022-11-18
Consider switching to Brave for base browser.


ikreymer:
The actually change would be made via https://github.com/webrecorder/browsertrix-browser-base, but probably should discuss here.

Main benefit would be the ability to enable adblock + cookie popup blocking out-of-the-box via the features provided by Brave, instead of having to maintain them separately, as per #152 and #182. Some trade-offs:

Pros:
 - Less surface area to maintain, can use the state-of-the-art ad + cookie blocking provided by Brave.
 - Potentially simpler release cycle for base browser image, can just use the built binaries for amd and arm from GitHub, rather than Chrome for amd64, and Chromium for arm64
 - Potential enhancements to crawl data, using Brave-specific features, like https://github.com/brave/brave-browser/wiki/PageGraph
 
 Cons:
 - Mostly already implemented adblocking + cookie popup blocking
 - Brave doesn't provide a command-line option to enable ad + cookie blocking, would need to use a profile, instead of a command-line option.
 
Currently leaning towards making the switch.

ikreymer:
Thinking of switching to Brave for next version to make it easier to simplify keeping up-to-date with latest browsers.
Unfortunately, Canonical is no longer releasing Ubuntu .deb for Chromium, and there is no Linux Chrome for ARM64,
so this is our best option to release both ARM64 and AMD64 images.
Brave has ARM64 and AMD64 release published to GH.
Let us know if you this is will pose issues for your use cases (we may be able to support both Brave and Chrome for AMD64)

Shrinks99:
Considering closing this issue because we did this and it's complete!

PR #400, opened 2023-09-29
Switch to Brave Base Image

ikreymer:
To simplify keeping up with latest Chromium changes, and to benefit from improved ad-blocking in Brave, switching to using Brave for crawling. See #189 for more details.

Includes:
- switch base browser to brave base image 1.58.135
- bump to 0.11.3


tw4l (approved):
Will want to do more testing before release, but looks good!

Issue #403, opened 2023-10-04
Improved Text Extraction, stored to WARC


ikreymer:
Currently, we extract the initial text if the `--text` flag is set and add this to the pages.jsonl. However, this has proven not entirely scalable, especially if the text is very large. Instead, we should just extract the text and put it into WARCs as `resource` records, similar to screenshots, with type `text/plain`

Additionally, can extract text immediately after page load, as well as after behaviors run.

The resource records can be:
- `urn:text:<url>` for initial page load text
- `urn:textFinal:<url>` for text extracted after behaviors run.

Could further optimize and only store `textFinal` if it is different then initial text, allowing for better detection of when running behaviors causes the dynamic changes to the text of the page.

The idea would be that the CDXJ could then be searched for `urn:text:` or `urn:textFinal:` prefix to find all WARC records with extracted text, which could then be ingested directly into full-text search systems.

Some questions:
- Does `urn:text` and `urn:textFinal` make sense?
- Should there be a textFinal always, or only if different? Or should it be a `revisit` if unchanged from initial?
- Compared to screenshots, there is even more variability with how text is extracted. How should this metadata be reflected? There is the CDP function used (eg. DOMSnapshot.captureSnapshot) + extracting logic, which could change over time. For example, could put the extracting function as separate repo, and include its commit, something like:
`WARC-Json-Metadata: {"extractMethod": "DOMSnapshot.captureSnapshot", "extractParser": "https://github.com/webrecorder/extract-text-lib@abcdefg"}`
- Other ways to version text extraction for later comparison?

Feedback welcome!

rgaudin:
Not using these texts yet but planning to, I'd actually prefer to have them in the WARC. Makes sense and WARC is flexible enough to filter them out easily.

I like the processing metadata.

Questions around text/textFinal could be resolved by requesting user to choose what he wants. A `--textInitial` and a `--textFinal` would settle this. I don't see how the initial text can be more useful than the final one… besides knowing if behavior changed anything but in that case it's a metadata or a debug info.

edsu:
I like this proposal!

I wonder, would it be helpful to use the `conversion` record type instead of `resource` records, and then linking the record with `WARC-Refers-To`?

[...]

PR #404, opened 2023-10-04
improved text extraction: (addresses #403)

ikreymer:
- use DOMSnapshot.captureSnapshot instead of older DOM.getDocument to get the snapshot (consistent with ArchiveWeb.page) - should be slightly more performant
- keep option to use DOM.getDocument
- refactor warc resource writing to separate class, used by text extraction and screenshots
- write extracted text to WARC files as 'urn:text:<url>' after page loads, similar to screenshots
- also store final text to WARC as 'urn:textFinal:<url>' if it is different
- cli options: update `--text` to take one more more comma-separated string options `--text to-warc,to-pages,final-to-warc`. For backwards compatibility, support `--text` and `--text true` to be equivalent to `--text to-pages`.

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -12,7 +12,7 @@ import yaml from "js-yaml";
 import * as warcio from "warcio";
 
 import { HealthChecker } from "./util/healthcheck.js";
-import { TextExtract } from "./util/textextract.js";
+import { TextExtractViaSnapshot } from "./util/textextract.js";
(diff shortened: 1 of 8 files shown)


tw4l (approved):
Tested manually and working with a few changes, but would be a good candidate for some new tests

ikreymer:
Added test!

PR #405, opened 2023-10-05
Execution Time Follow-Up Work

ikreymer:
Update execution time system to new approach described in https://github.com/webrecorder/browsertrix-cloud/pull/1218#issuecomment-1747988456, follow-up to #397 

Using a single `:start:<id>` key per instance

Also move setStartTime() earlier in init process.

### DIFF
--- a/util/state.js
+++ b/util/state.js
@@ -202,9 +202,11 @@ return 0;
     if (!endTimeTs) {
       endTimeTs = Date.now();
     }
-    const prevStartTime = await this.redis.getset(this.startkey, "0");
+    const prevStartTime = await this.redis.getset(this.startkey, 0);
     if (prevStartTime) {
-      await this.redis.incrby(this.execTimeKey, (endTimeTs - Number(prevStartTime)) / 1000);
+      const newExecTime = Math.ceil((endTimeTs - Number(prevStartTime)) / 1000);
+      logger.debug("Adding exec time", {newExecTime}, "state");
+      await this.redis.incrby(this.execTimeKey, newExecTime);
       return true;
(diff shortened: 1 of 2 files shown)


util/state.js
```diff
@@ -187,23 +187,29 @@ return 0;
...
 
   async setStartTime() {
-    const startTime = this._timestamp();
-    return await this.redis.rpush(`${this.startkey}:${this.uid}`, startTime);
-  }
-
-  async getStartTimes() {
-    return await this.redis.lrange(`${this.startkey}:${this.uid}`, 0, -1);
+    const startTimeTs = Date.now();
+    await this.incExecTime(startTimeTs);
```
tw4l, line 191:
```suggestion
    // if previous start time exists, incr execution time by difference before setting new time
    await this.incExecTime(startTimeTs);
```
Something like that? Otherwise from the code itself without context it's not super obvious what's going on here

util/state.js
```diff
@@ -187,23 +187,29 @@ return 0;
...
+  }
 
-    const endTime = this._timestamp();
-
-    return await this.redis.rpush(`${this.endkey}:${this.uid}`, endTime);
+  async incExecTime(endTimeTs = null) {
+    if (!endTimeTs) {
+      endTimeTs = Date.now();
+    }
+    const prevStartTime = Number(await this.redis.getset(this.startkey, "0") || 0);
```
tw4l, line 205: `getset` seems to be deprecated, see: https://redis.io/commands/getset/

ikreymer:
Closing in favor for #405

PR #406, opened 2023-10-06
Fast cancelation + remove time counter

ikreymer:
Follow-up to #378, supersedes #405.

Removing start/end time count from the crawler itself.

Instead, adding 'isCrawlCanceled()` which, if true, will immediately exit the process.

Also checking isCrawlCanceled() on interrupts, to allow immediate shutdown when crawl is canceled.

When in 'restart on exit' mode, have fatal() return a 0 exit code. This then tells k8s to not restart the pod (the actual failure state is set in redis), while a non-zero exit code will result in a restart.

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -65,6 +65,13 @@ export class Crawler {
     logger.setLogLevel(this.params.logLevel);
     logger.setContext(this.params.context);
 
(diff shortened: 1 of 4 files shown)


tw4l (approved):
LGTM!

PR #407, opened 2023-10-09
tests: disable ad-block tests: seeing inconsistent ci behavior

ikreymer:
Disable these tests until we get consistent repro of the issues.
They seem to be failing sporadically, though the content does appear to be blocked in Brave on manual testing.


tw4l (approved):
Should get back to this soon to figure it out, but looks good for now

PR #408, opened 2023-10-10
Support adding/removing exclusions without restarting the crawler

ikreymer:
Part of work for webrecorder/browsertrix-cloud#1216:
- support adding/removing exclusions dynamically via a Redis message list
- add processMessage() which checks <uid>:msg list for any messages
- handle addExclusion / removeExclusion messages to add / remove exclusions for each seed
- also add filterQueue() which filters queue, one URL at a time, async when a new exclusion is added

### DIFF
--- a/util/seeds.js
+++ b/util/seeds.js
@@ -45,6 +45,25 @@ export class ScopedSeed
     }
   }
 
+  addExclusion(value) {
+    if (!value) {
+      return;
+    }
+    if (!(value instanceof RegExp)) {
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Looks good and seems to be working. I'm wondering if there's a way we can add test coverage for this? I don't think we have tests yet that respond to redis commands but seems possible?

ikreymer:
Will add tests in separate PR

PR #420, opened 2023-10-24
Return User-Agent on all code path to set headers appropriately

### LINKED ISSUE #419: When passed as CLI argument, User-Agent is not always set
When User-Agent is passed through the `--userAgent` CLI argument, this value is used for most calls but not all (e.g. it is not used for direct fetch of non-HTML documents).

Fix is obvious (once issue has been spotted), I will open a PR in few secs.

benoit74:
Fix #419 

Do you plan to release a v0.11.3 or do we have to wait for v0.12.0 out of beta? And do you have any estimate for v0.12.0 out of beta?

Thank you!

### DIFF
--- a/crawler.js
+++ b/crawler.js
@@ -138,7 +138,7 @@ export class Crawler {
     // override userAgent
     if (this.params.userAgent) {
       this.emulateDevice.userAgent = this.params.userAgent;
-      return;
+      return this.params.userAgent;
     }
 
     // if device set, it overrides the default Chrome UA


tw4l (approved):

tw4l:
Hi @benoit74 , thanks for that! This will be in 0.12.0 beta, we should have another beta shortly to test with this and some other features and then will get out the 0.12.0 release once the beta's been tested a bit.

PR #425, opened 2023-11-04
TypeScript Conversion

ikreymer:
Follows #424. Converts the upcoming 1.0.0 branch based on in-browser recording to TypeScript.
Fixes #426

PR #428, opened 2023-11-08
Add Prettier to the repo, and format all the files!

emma-sg:
This adds prettier to the repo, and sets up the pre-commit hook to auto-format as well as lint.

Should be merged after #425.


ikreymer:
@emma-sg might be easier to redo the prettier past off dev-1.0.0 to avoid the conflicts?

emma-sg:
Yeah probably!

emma-sg:
@ikreymer rebased! Waiting on builds but would appreciate a review

ikreymer:
Thanks for the rebase!

only thought what if we specified the directories explicitly for `prettier src/ tests/` and `eslint src/ tests/` instead of `prettier .` and `eslint .`?
Have a bunch of other files, and also crawls go into the same directory, so may be easier rather than adding ignores?

package.json
```diff
@@ -8,7 +8,10 @@
   "license": "AGPL-3.0-or-later",
   "scripts": {
     "tsc": "tsc",
-    "lint": "eslint *.js tests/*.test.js",
+    "format": "prettier . --check",
+    "format:fix": "prettier . --write",
+    "lint": "eslint .",
+    "lint:fix": "yarn format:fix && eslint . --fix",
```
ikreymer, line 14: How about this, focusing on just src and test
```suggestion
    "format": "prettier src/ tests/ --check",
    "format:fix": "prettier src/ tests/ --write",
    "lint": "eslint src/ tests/",
    "lint:fix": "yarn format:fix && eslint src/ tests/ --fix",
```

emma-sg: Is there a particular reason we'd want to avoid formatting other things? My thinking is that this is our source code, and non-formatted files should _generally_ be avoided, and so I'd rather make exclusions. With this approach we'd have to include every new file not in the `src` and `tests` directory too — `prettier src/ tests/ .github/ config/config.yaml html tsconfig.json test-setup.js README.md package.json docker-compose.yml CHANGES.md .pre-commit-config.yaml .eslintrc.cjs --check` seems unnecessary to me

ikreymer: I don't think we need to run prettier on any of those other files, they're rarely edited, and things like README and docker-compose don't need a specific format. Maybe we just start with this and reevaluate later?

emma-sg: I mean, I guess, but by that token why use an automated formatter at all? The whole point here is minimizing the amount of thinking we have to do in the future about formatting, so in my mind the more we can have auto-formatted the better, unless there's a specific reason not to format something — a file not being edited often doesn't mean you should have to think more about formatting than you do for other files

ikreymer: As discussed, updated ignores files and with `crawls`, `test-crawls`, `scratch` dir for misc scripts that are not usually checked in and should not be linted/prettified.

ikreymer:
ack, will add updates to ignore files in separate PR.

PR #433, opened 2023-11-12
Fix potential for pending list never being processed

ikreymer:
Due to an optimization, numPending() call assumed that queueSize() would be called to update cached queue size. However, in the current worker code, this is not the case. Remove cacheing queue size and just check queue size in numPending(), to ensure pending list is always processed.

Perhaps want to port this to a 0.12.2 release, as this addresses an edge case after exclusions where pending URLs remain but are not reprocessed.

### DIFF
--- a/src/util/state.ts
+++ b/src/util/state.ts
@@ -117,7 +117,6 @@ declare module "ioredis" {
 export class RedisCrawlState {
   redis: Redis;
   maxRetryPending = 1;
-  _lastSize = 0;
 
   uid: string;
   key: string;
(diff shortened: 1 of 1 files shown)


tw4l (approved):
Nice catch!

tw4l:
And I agree, porting to 0.12.2 seems like a good idea :)

ikreymer:
> And I agree, porting to 0.12.2 seems like a good idea :)

Will open a follow-up PR for 0.12.2

PR #435, opened 2023-11-12
Add types + validation for log context options

ikreymer:
- add LogContext type and enumerate all log contexts
- also add LOG_CONTEXT_TYPES array to validate --context arg
- make errJSON() convert unknown to proper type
- make logger info/error/debug accept unknown as well, to avoid explicit 'any' typing in all catch handlers

### DIFF
--- a/src/util/logger.ts
+++ b/src/util/logger.ts
@@ -11,7 +11,7 @@ Object.defineProperty(RegExp.prototype, "toJSON", {
 
 // ===========================================================================
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function errJSON(e: unknown): Record<string, any> {
+export function formatErr(e: unknown): Record<string, any> {
   if (e instanceof Error) {
     return { type: "exception", message: e.message, stack: e.stack || "" };
   } else if (typeof e === "object") {
@@ -34,7 +34,7 @@ export const LOG_CONTEXT_TYPES = [
   "exclusion",
   "screenshots",
   "screencast",
-  "originoverride",
+  "originOverride",
   "healthcheck",
   "browser",
   "blocking",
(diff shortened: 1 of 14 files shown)


src/util/logger.ts
```diff
@@ -10,16 +10,47 @@ Object.defineProperty(RegExp.prototype, "toJSON", {
...
+  "recorder",
+  "writer",
+  "state",
+  "redis",
+  "storage",
+  "text",
+  "exclusion",
+  "screenshots",
+  "screencast",
+  "originoverride",
```
tw4l, line 37: Should we make this `originOverride` to be more consistent with the camel casing used elsewhere in `LOG_CONTEXT_TYPES`?

src/util/logger.ts
```diff
@@ -58,20 +89,14 @@ class Logger {
...
     context: string,
     logLevel = "info",
   ) {
-    if (data instanceof Error) {
-      data = errJSON(data);
-    } else if (typeof data !== "object") {
-      data = { message: data.toString() };
-    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const data: Record<string, any> = errJSON(dataUnknown);
```
tw4l, line 99: Wondering if we should rename `errJSON()` now that it's being used to format messages for all log levels? Otherwise this is a nice solution for getting a consistent type from an unknown (and I got to learn about `unknown`, so thanks :) )

tw4l (approved):

src/util/logger.ts
ikreymer: This was referring to formatting the exception rather than the log level specifically

ikreymer: The name came from it checking `if (e instanceof Error)` and converting that to json.. not sure if there's a better name now - toJSON() seems overly generic

ikreymer: Converted to `formatErr`

src/util/logger.ts
ikreymer: done

PR #436, opened 2023-11-13
[Docs] Better regex examples

### LINKED ISSUE #416: [Docs] More exclusion examples?
Perhaps a dedicated wikipage? I tried to use the exclusion rules ?updated-min=$ &updated-min=$ and it didn't work. But the exclusion rule /share-post.g$ works. I assume this is due to some incompatibility issue with the special characters. Since this is not documented, I'll have to find out on my own which one is causing this through testing (but it's probably =)

pato-pan:
Fixes #416 

Please look at the following details

1. The complex example
2. Suggested way of verifying if the regex works

thanks


ikreymer:
Thanks for the additions, I think we probably want to tweak that text a bit. And you're right, we should add a log option to indicate when a page URL is skipped, otherwise looking through the logs isn't as helpful

pato-pan:
> Thanks for the additions, I think we probably want to tweak that text a bit. And you're right, we should add a log option to indicate when a page URL is skipped, otherwise looking through the logs isn't as helpful

anything you have in mind that I can do on my end?

README.md
```diff
@@ -388,7 +388,7 @@ that will take precedence over the `--extraHops`.
 
 #### Scope Rule Examples
 
-For example, the following seed will start on `https://example.com/startpage.html` and crawl all pages on the `https://example.com/` domain, except pages that match the regexes `example.com/skip.*` or `example.com/search.*`
+The seed below will start on `https://example.com/startpage.html` and crawl all pages on the `https://example.com/` domain. With the .* regular expression, any character after will be excluded except pages that match the regexes `example.com/skip.*` or `example.com/search.*`. It will also match characters before .* if placed on the other end of the string, this means that example.com/skip/postfeed will also be excluded.
```
Shrinks99, line 391:
```suggestion
The seed below will start on `https://example.com/startpage.html` and crawl all pages on the `https://example.com/` domain. With the `.*` regular expression, any character after will be excluded except pages that match the regexes `example.com/skip.*` or `example.com/search.*`. It will also match characters before `.*` if placed on the other end of the string, this means that example.com/skip/postfeed will also be excluded.
```

README.md
```diff
@@ -397,11 +397,11 @@ seeds:
     exclude:
       - example.com/skip.*
       - example.com/search.*
+      - .*postfeed&
 
 ```
 
-In the following example, the scope include regexes will crawl all page URLs that match `example.com/(crawl-this|crawl-that)`,
-but skip URLs that end with 'skip-me'. For example, `https://example.com/crawl-this/page.html` would be crawled, but `https://example.com/crawl-this/pages/skip` would not be.
+In the following example, the scope includes regexes that will crawl all page URLs that match `example.com/(crawl-this|crawl-that)` since the | indicates an "or". URLs that match skip$ will be excluded but skip-me wouldn't. For example, `https://example.com/crawl-this/page.html` and `https://example.com/crawl-this/pages/skipme/not` would be crawled, but `https://example.com/crawl-this/pages/skip` would not be.
```
Shrinks99, line 404:
```suggestion
In the following example, the scope includes regexes that will crawl all page URLs that match `example.com/(crawl-this|crawl-that)` since the | indicates an "or". URLs that match `skip$` will be excluded but `skip-me` wouldn't. For example, `https://example.com/crawl-this/page.html` and `https://example.com/crawl-this/pages/skipme/not` would be crawled, but `https://example.com/crawl-this/pages/skip` would not be.
```

README.md
```diff
@@ -411,12 +411,26 @@ seeds:
       - skip$
 ```
 
+It's possible to do more complex regular expressions. It will only target characters and numbers after search until the word ID= shows, and any numbers after the word ID will be matched. This means that https://example.com/search/6vH8R4TmID=5819 won't be crawled, but https://example.com/search/ID=5819, https://example.com/search/6vH8R4Tm and https://example.com/search/2o3Jq89cID=5ag8h19 will be crawled
```
Shrinks99, line 414:
```suggestion
For a more complex regular expression, one might only target characters and numbers after `search` until the string `ID=` shows, and any numbers after the string `ID` will be matched. This means that https://example.com/search/6vH8R4TmID=5819 won't be crawled, but https://example.com/search/ID=5819, https://example.com/search/6vH8R4Tm, and https://example.com/search/2o3Jq89cID=5ag8h19 will be crawled
```

README.md
```diff
@@ -411,12 +411,26 @@ seeds:
...
+      - *.example.com/search/[A-Za-z0-9]+ID=[0-9]+
+```
+
 The `include`, `exclude`, `scopeType` and `depth` settings can be configured per seed, or globally, for the entire crawl.
 
 The per-seed settings override the per-crawl settings, if any.
 
 The test suite [tests/scopes.test.js](tests/scopes.test.js) for additional examples of configuring scope inclusion and exclusion rules.
 
+It's important to note that the include and exclude rules will always have regular expressions in effect. This means that for your rules to match, you may have to escape special characters that commonly appear in urls like "?", "+", or ".". You just have to place a \ before the character. For example youtube.com/watch\?rdwz7QiG0lk
```
Shrinks99, line 430: Can probably tighten this up slightly, though my suggestion might not be ideal for highlighting the backslash in the YouTube url??

```suggestion
It's important to note that the include and exclude rules will always have regular expressions in effect. For rules to match, you may have to escape special characters that commonly appear in urls like `?`, `+`, or `.`. by placing a `\` before the character. For example: `youtube.com/watch\?rdwz7QiG0lk`
```

Shrinks99 (commented):
Made some minor comments for formatting.  Thanks for the additional examples!

README.md
pato-pan: I think you can still highlight it by putting it in bold? I am not sure if it was highlighted originally, but if it was I probably did it by accident :]

It's a nice addition to highlight it, but it's not necessary and I think having a consistent single quote throughout every example is a better compromise.

> commonly appear in urls like `?`, `+`, or `.`**.** by placing a `\` before the character

I think the dot after the list of special characters to escape should be removed. I don't know how to edit your suggestion. 

I appreciate the suggestions, it's more concise and to the point with better pacing.

tw4l:
Thank you for this! At long last I've incorporated it into the docs. Instead of merging this PR, I've added these changes (tested and modified slightly) to PR https://github.com/webrecorder/browsertrix-crawler/pull/436, which moves the crawler docs out of the README and into a separate proper documentation site.

PR #450, opened 2023-12-13
detect invalid custom behaviors on load:

ikreymer:
- on first page, attempt to evaluate the behavior class to ensure it compiles
- if fails to compile, log exception with fatal and exit
- update behavior gathering code to keep track of behavior filename
- tests: add test for invalid behavior which causes crawl to exit with fatal exit code (17)


tw4l (approved):
Tested locally and working well, thanks!

Issue #456, opened 2024-01-08
RCE Vulnerability in puppeter-core


Feribv:
Hello,

Testing the Application's security, I've noticed that puppeter-core 20.7.4 should be updated to at least 2.8.2 as it is vulnerable
![image](https://github.com/webrecorder/browsertrix-crawler/assets/53820850/debbdecd-69bf-46dd-a541-4250ad493efa)

Could somebody update this in the next release?
Thank you!

tw4l:
PR with fix is in: https://github.com/webrecorder/browsertrix-crawler/pull/459

PR #459, opened 2024-01-15
Bump puppeteer-core to ^20.8.2 to patch vulnerability

tw4l:
Fixes https://github.com/webrecorder/browsertrix-crawler/issues/456

We'll want to do the same for the 1.0.0-beta branch.

It looks like bumping straight to the latest major release of puppeteer-core (21.x.x) resulted in some test failures around text extraction and blockrules, so I'm just bumping to a version that contains the patch for now until we have more time to investigate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "ioredis": "^4.27.1",
     "js-yaml": "^4.1.0",
     "minio": "7.0.26",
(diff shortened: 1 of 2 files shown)


tw4l:
Hm, seeing some test failures, looking into it.

PR #460, opened 2024-01-17
skipping resources: ensure HEAD, OPTIONS, 204, 206, and 304 response/request pairs are not written to WARC

ikreymer:
Allows for skipping network traffic that doesn't need to be stored, as it is not necessary/will result in incorrect replay (eg. 304 instead of a 200).


tw4l:
Kind of an edge case but I think 204 responses can sometimes be sent in response to a successful GET, POST, or DELETE request - might be worth keeping those request/response pairs?

PR #465, opened 2024-02-16
Page Resources: Include Cached Resources

ikreymer:
Ensure cached resources (that are not written to WARC) are still included in the `url:pageinfo:...` records. This will make it easier to track which resources are actually *loaded* from a given page.

To test, run crawl for a site with cached resources, such as https://webrecorder.net/ and verify that each page pageinfo includes all the resources.

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -186,7 +186,12 @@ export class Recorder {
         requestId: params.requestId,
         ...this.logDetails,
       });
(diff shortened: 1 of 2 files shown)


tw4l (approved):
Tested with a https://webrecorder.net crawl - pages like /about have cached resources written into the `pageinfo` record after the change.

ikreymer:
Also added tests to verify `urn:pageinfo:` records

PR #469, opened 2024-02-20
QA Crawl Support (Beta)

ikreymer:
Initial support for QA crawling, crawling over an existing replay to generate QA/comparison data.
Can be deployed with `webrecorder/browsertrix-crawler qa` entrypoint.

Requires `--qaSource`, pointing to WACZ or multi-WACZ json that will be QAd.

Also supports `--qaRedisKey` where QA comparison data will be pushed, if specified.
Supports `--qaDebugImageDiff` for outputting crawl / replay/ diff images.

The data pushed to redis is `{"url": <page url>", "comparison": <...>"}` where comparison is:
```
  comparison: {
    screenshotMatch?: number;
    textMatch?: number;
    resourceCounts: {
      crawlGood?: number;
      crawlBad?: number;
      replayGood?: number;
      replayBad?: number;
    };
  };
  ```

### CODE REVIEW

File: src/replaycrawler.ts [hunk 1]
```diff
@@ -0,0 +1,671 @@
...
+    logger.info("Resource counts", { url, ...resourceCounts }, "replay");
+
+    // if (crawlGood !== replayGood) {
+    //   console.log("*** ORIG");
+    //   console.log(origResData);
+    // }
+
+    // //console.log(origResData);
+    // console.log("*** REPLAY");
+    // console.log(pageInfo);
```
tw4l, line 515:
```suggestion
```

src/replaycrawler.ts
```diff
@@ -0,0 +1,671 @@
...
+    // optionally reload
+    // await page.reload();
+
+    await sleep(10);
+
+    // console.log("Frames");
+    // for (const frame of page.frames()) {
+    //   console.log(`${frame.name()} - ${frame.url()}`);
+    // }
+
```
tw4l, line 338:
```suggestion
```

tw4l: Or keep, but as debug?

tw4l (approved):
This is looking good! Tested with local WACZs as well as serving files via an http server.

I've left some comments and suggestions (suggestions are mostly adding some logging and removing some commented out bits that are no longer necessary).

We could use some tests - possibly an integration test to start that runs a crawl, then QAs the crawl, and checks the outputs?

tw4l:
Could we also add the page id to the data pushed to Redis, just to help with matching in Browsertrix?

src/util/state.ts
```diff
@@ -690,4 +701,8 @@ return 0;
   async writeToPagesQueue(value: string) {
     return await this.redis.lpush(this.pageskey, value);
   }
+
+  async writeToComparisonQueue(compareQKey: string, value: string) {
```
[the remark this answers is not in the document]
ikreymer: Has been replaced with just including comparison data in the page queue.

ikreymer:
> Could we also add the page id to the data pushed to Redis, just to help with matching in Browsertrix?

The QA data is now merged with the page data, so should already be in one place.

PR #470, opened 2024-02-21
Set warc prefix via WARC_PREFIX env var

ikreymer:
In addition to `--warcPrefix` flag, also support WARC_PREFIX env var, which takes precedence.
Bump to 1.0.0-beta.4


tw4l (approved):
Looks good!

PR #473, opened 2024-02-22
Include WARC prefix for screenshots and text WARCs

ikreymer:
Ensure the env var / cli warc prefix is also applied to `screenshots.warc.gz` and `text.warc.gz`


tw4l (approved):
Tested, looks good!

ikreymer:
Per discussion, also added crawlId if custom prefix is used to be consistent with main crawl WARCs.

PR #476, opened 2024-02-24
new seed on redirect + error page check:

### LINKED ISSUE #475: Handle seed redirects
A seed URL may redirect to a different page via a 3xx. For prefix/domain scope crawls, this may result in the page after redirect being out of scope, resulting in crawl ending early.
eg. https://nytimes.com/ redirects to https://www.nytimes.com/, so if the former is specified as a seed with scopeType prefix, the crawl will end after the first page.

To be as flexible as possible, the crawler should check if seed page URL is different from original URL, and add the new URL as a seed as well with same scoping rules.

ikreymer:
- if a seed page redirects (page response != seed url), then add the final url as a new seed with same scope
- add newScopeSeed() to ScopedSeed to duplicate seed with different URL, store original includes / excludes
- also add check for 'chrome-error://' URLs for the page, and ensure page is marked as failed if page.url() starts with chrome-error://
- fixes #475

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1521,7 +1521,7 @@ self.__bx_behaviors.selectMainBehavior();
     data: PageState,
     selectorOptsList = DEFAULT_SELECTORS,
   ) {
-    const { url, seedId, depth } = data;
+    const { url, depth } = data;
 
     const logDetails = data.logDetails;
 
@@ -1556,11 +1556,27 @@ self.__bx_behaviors.selectMainBehavior();
         throw new Error("page response missing");
       }
 
+      const respUrl = resp.url();
+      const isChromeError = page.url().startsWith("chrome-error://");
+
+      if (respUrl !== url && depth === 0 && !isChromeError) {
+        const seed = this.params.scopedSeeds[data.seedId];
+        this.params.scopedSeeds.push(seed.newScopedSeed(respUrl));
+        data.seedId = this.params.scopedSeeds.length - 1;
+        logger.info("Seed page redirected, adding redirected seed", {
+          origUrl: url,
+          newUrl: respUrl,
+          seedId: data.seedId,
+        });
+      }
+
       // Handle 4xx or 5xx response as a page load error
       const statusCode = resp.status();
+      const statusString = statusCode.toString();
       if (
-        statusCode.toString().startsWith("4") ||
-        statusCode.toString().startsWith("5")
+        statusString.startsWith("4") ||
+        statusString.startsWith("5") ||
+        isChromeError
       ) {
         if (failCrawlOnError) {
           logger.fatal("Seed Page Load Error, failing crawl", {
@@ -1568,10 +1584,15 @@ self.__bx_behaviors.selectMainBehavior();
             ...logDetails,
           });
         } else {
-          logger.error("Non-200 Status Code, skipping page", {
-            statusCode,
-            ...logDetails,
-          });
+          logger.error(
+            isChromeError
+              ? "Page Crashed on Load"
+              : "Non-200 Status Code, skipping page",
+            {
+              statusCode,
+              ...logDetails,
+            },
+          );
           throw new Error("logged");
         }
       }
@@ -1648,6 +1669,8 @@ self.__bx_behaviors.selectMainBehavior();
       logger.debug("Skipping link extraction for non-HTML page", logDetails);
       return;
     }
+
+    const { seedId } = data;
 
     const seed = this.params.scopedSeeds[seedId];
 
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1559,7 +1559,7 @@ self.__bx_behaviors.selectMainBehavior();
       const respUrl = resp.url();
       const isChromeError = page.url().startsWith("chrome-error://");
 
-      if (respUrl !== url && depth === 0 && !isChromeError) {
+      if (depth === 0 && !isChromeError && respUrl !== url) {
         const seed = this.params.scopedSeeds[data.seedId];
         this.params.scopedSeeds.push(seed.newScopedSeed(respUrl));
         data.seedId = this.params.scopedSeeds.length - 1;
--- a/src/util/seeds.ts
+++ b/src/util/seeds.ts
@@ -14,20 +14,22 @@ export class ScopedSeed {
   url: string;
   scopeType: ScopeType;
   include: RegExp[];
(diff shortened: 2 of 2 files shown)


benoit74:
I love it, this would greatly simplify openZIM `zimit` scraper, we do the same logic in Python and was a bit of a concern for us. FYI, our concern is that our current solution means two UAs, and even when you fake the UA header some WAF achieve to detect there is two UA thanks to some "TLS black magic".

So clearly, the sooner this is merged, the better for us ^^

src/crawler.ts
```diff
@@ -1556,22 +1556,43 @@ self.__bx_behaviors.selectMainBehavior();
         throw new Error("page response missing");
       }
 
+      const respUrl = resp.url();
+      const isChromeError = page.url().startsWith("chrome-error://");
+
+      if (respUrl !== url && depth === 0 && !isChromeError) {
```
benoit74, line 1562: Very minor suggestion, I would rewrite it `if (depth === 0 && !isChromeError && respUrl !== url ) {` for clarity

benoit74:
Would it be easy to "inform" the user in a programmatic way about the new URL? We could detect it based on the record depth, but I don't think that the depth is saved in the WARC records. We probably need to have access to this new "fixed seed" URL at some point in our `zimit` scraper to know what is the main entry point. This would also help [replayer](https://replayweb.page/) to know where he should start indeed.

tw4l:
> Would it be easy to "inform" the user in a programmatic way about the new URL? We could detect it based on the record depth, but I don't think that the depth is saved in the WARC records. We probably need to have access to this new "fixed seed" URL at some point in our `zimit` scraper to know what is the main entry point. This would also help [replayer](https://replayweb.page/) to know where he should start indeed.

Hm, this is a good point. Currently I think one would have to dig into the logs to find this out. Perhaps we could add a new `pages/seeds.jsonl` file or similar?

benoit74:
> Perhaps we could add a new pages/seeds.jsonl file or similar?
This would be ok for us as well (anything else than parsing the logs is probably ok indeed 😅, just choose what suit your "vision" of the product, we can adapt 😉)

ikreymer:
> > Perhaps we could add a new pages/seeds.jsonl file or similar?
> 
> This would be ok for us as well (anything else than parsing the logs is probably ok indeed 😅, just choose what suit your "vision" of the product, we can adapt 😉)

Hmm, could perhaps add it to the `--statsFilename` log in some way, if that's easier, although there could be more than one seed and more than one seed that redirects... I think this flag is mostly for Zimit use so can add something there. What would be most useful for Zimit? To print the actual seed used post-redirect? To indicate something like `seedRedirected: true`?

ikreymer:
One remaining question: how should we handle a seed that redirects to a page that has already been crawled? Perhaps not as common for single-seed, but could happen with multiple seed crawls, eg.

https://someseed.example.com/
https://someotherseed.example.com/ -> https://someseed.example.com/

Should we add another `isInScope()` for the response URL, similar to: https://github.com/webrecorder/browsertrix-crawler/blob/dev-1.0.0/src/util/worker.ts#L391

benoit74:
> Hmm, could perhaps add it to the --statsFilename log in some way, if that's easier, although there could be more than one seed and more than one seed that redirects... I think this flag is mostly for Zimit use so can add something there. What would be most useful for Zimit? To print the actual seed used post-redirect? To indicate something like seedRedirected: true?

For Zimit, `--statsFilename` is clearly the easiest solution since we already watch this file. I think we probably need the actual seed used post-redirect ; could be a list of exactly what you push in the log (origUrl, newUrl, seedID), which would handle nicely the case where multiple seeds have been passed.

> One remaining question: how should we handle a seed that redirects to a page that has already been crawled? Perhaps not as common for single-seed, but could happen with multiple seed crawls

Anyway, there will be a WARC record  in the archive for the seed that redirects, isn't it? If yes, then I would just skip adding the new seed in such a case.

ikreymer:
> > Hmm, could perhaps add it to the --statsFilename log in some way, if that's easier, although there could be more than one seed and more than one seed that redirects... I think this flag is mostly for Zimit use so can add something there. What would be most useful for Zimit? To print the actual seed used post-redirect? To indicate something like seedRedirected: true?
> 
> For Zimit, `--statsFilename` is clearly the easiest solution since we already watch this file. I think we probably need the actual seed used post-redirect ; could be a list of exactly what you push in the log (origUrl, newUrl, seedID), which would handle nicely the case where multiple seeds have been passed.
> 
> > One remaining question: how should we handle a seed that redirects to a page that has already been crawled? Perhaps not as common for single-seed, but could happen with multiple seed crawls
> 
> Anyway, there will be a WARC record in the archive for the seed that redirects, isn't it? If yes, then I would just skip adding the new seed in such a case.

Yes, there will be, so when parsing the WARC you can just follow the redirect chain for the seed until you find the non-redirect. Maybe a bit more work but should be straightforward since you're verifying that the WARC has the seed already. That might be the simplest option, since the statsFilename output is the stats on the current state of the crawl, not about the seed.

ikreymer:
> One remaining question: how should we handle a seed that redirects to a page that has already been crawled? Perhaps not as common for single-seed, but could happen with multiple seed crawls, eg.
> 
> https://someseed.example.com/ https://someotherseed.example.com/ -> https://someseed.example.com/
> 
> Should we add another `isInScope()` for the response URL, similar to: https://github.com/webrecorder/browsertrix-crawler/blob/dev-1.0.0/src/util/worker.ts#L391
> 
> If the page has already been crawled, we just skip the page? That would result in behaviors / screenshots being skipped, so may be confusing. Or, maybe we just commit to full capture of any page that is specified as a seed.
> 
> Thoughts?

Decided to keep current behavior, as user supplied seeds are guaranteed to be loaded and processed fully. Can perhaps as a separate seed dedup option later if necessary.

src/crawler.ts
ikreymer:
```suggestion
      if (depth === 0 && !isChromeError && respUrl !== url) {
```

benoit74:
> Yes, there will be, so when parsing the WARC you can just follow the redirect chain for the seed until you find the non-redirect. Maybe a bit more work but should be straightforward since you're verifying that the WARC has the seed already. That might be the simplest option, since the statsFilename output is the stats on the current state of the crawl, not about the seed.

This is enough indeed, no need for more I think (this PR is already a great step forward for us)

PR #477, opened 2024-02-28
store page statusCode if not 200

ikreymer:
don't treat non-200 pages as errors, still extract text, take screenshots, and run behaviors
only consider actual page load errors, eg. chrome-error:// page url, as errors

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -91,6 +91,7 @@ type PageEntry = {
   text?: string;
   favIconUrl?: string;
   ts?: string;
+  statusCode?: number;
 };
 
 // ============================================================================
@@ -786,6 +787,10 @@ self.__bx_behaviors.selectMainBehavior();
     }
 
     data.loadState = LoadState.EXTRACTION_DONE;
+
+    if (data.statusCode >= 400) {
+      return;
+    }
 
     if (this.params.behaviorOpts) {
       if (!data.isHTMLPage) {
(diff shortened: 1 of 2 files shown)


tw4l:
In testing I'm seeing some pages in `pages.jsonl` with a `0` status code, e.g.:

```
{"id":"fdf6458e-ee38-4386-8301-b6ebf4d430a6","url":"https://docnow.io/","loadState":0,"statusCode":0}
```

tw4l:
Fetched non-HTML resources such as images all seem to also have a status code of 0 in `pages.jsonl`.

src/crawler.ts
```diff
@@ -1582,27 +1587,18 @@ self.__bx_behaviors.selectMainBehavior();
 
       // Handle 4xx or 5xx response as a page load error
       const statusCode = resp.status();
-      const statusString = statusCode.toString();
-      if (
-        statusString.startsWith("4") ||
-        statusString.startsWith("5") ||
-        isChromeError
-      ) {
+      data.statusCode = statusCode;
```
tw4l, line 1590: I think we need to add this for directly fetched non-HTML pages above as well to avoid the status code of 0 for fetched images, videos, etc.

src/crawler.ts
```diff
@@ -1974,6 +1971,10 @@ self.__bx_behaviors.selectMainBehavior();
       row.mime = mime;
     }
 
+    if (statusCode !== 200) {
+      row.statusCode = statusCode;
+    }
```
tw4l, line 1976: I wonder if it's better to write the status code to redis even if it's 200, and simply have the Browsertrix backend return 200 as a default instead of None if it's not set for a page, rather than have all 200s be `None` in the database? Could go either way though honestly.

tw4l (approved):

ikreymer:
> In testing with extra hops set to 1 I'm seeing some pages in `pages.jsonl` with a `0` load state and status code, e.g.:
> 
> ```
> {"id":"fdf6458e-ee38-4386-8301-b6ebf4d430a6","url":"https://docnow.io/","loadState":0,"statusCode":0}
> ```

loadState 0 means in didn't load, so would make sense for statusCode to also be 0/unset, but maybe we don't include it in that case.

src/crawler.ts
ikreymer: The PageOut now returns 200 if its unset in the db, since previous crawls will have it unset. I guess its a minor space saving, but also backwards compatible with previous usage - eg. if loadState > 0, and no statusCode, assume 200

ikreymer:
Pushed a change to set statusCode to 200 for direct fetch responses, since they're always 200 do to the filter used (if not a 200, then we load in the browser). Also not writing statusCode 0.

I think the general assumption we can have is:
- if loadState == 0, page didn't even load so no status code
- if loadState > 0, if statusCode is not set, assume 200, otherwise assume statusCode as set.

src/crawler.ts
tw4l: Up to you, but if there ever any other consumers for this data from the crawler/redis, I think it might be clearer to be explicit about the 200 status code rather than relying on knowledge that its absence means 200? Maybe I've just done too many system-to-system migrations where things like this gave me a bit of a headache haha.

ikreymer:
Thinking maybe it is better to just set the 200 to avoid any ambiguity.

PR #478, opened 2024-02-28
Ensure links added via behaviors also get processed

ikreymer:
Requires webrecorder/browsertrix-behaviors#69 / browsertrix-behaviors 0.5.3, which will add support for behaviors to add links.

Simplify adding links by simply adding the links directly, instead of batching to 500 links. Errors are already being logged in queueing a new URL fails.


tw4l (approved):
Might want to revise the PR description to remove the reference to flushLinks when squashing + merging, otherwise looks good! Nice simplification.

PR #480, opened 2024-03-04
Fail on status code option + requeue fix

ikreymer:
Add fail on status code option, --failOnInvalidStatus to treat non-200 responses as failures. Can be useful especially when combined with --failOnFailedSeed or --failOnFailedLimit

requeue: ensure requeued urls are requeued with same depth/priority, not 0

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1586,20 +1586,30 @@ self.__bx_behaviors.selectMainBehavior();
         });
       }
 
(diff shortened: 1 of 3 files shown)


tw4l (approved):
My initial reaction was to not want to add another arg required to make `--failOnFailedSeed` work, but after thinking about it a moment, what you have here is nice and flexible.

PR #481, opened 2024-03-04
warc: add Network.resourceType (https://chromedevtools.github.io/devt…

### LINKED ISSUE #451: Add request initiator to WARC?
To improve replay systems, it would be beneficial to know whether a request is JS-emmited (xhr, fetch) or not.
This information is not available in the WARC but is available at crawling time in the browser.
Would it be possible to add this information to a dedicated WARC Header?

See https://github.com/openzim/warc2zim/issues/140

ikreymer:
Add resourcesType value from https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-ResourceType as `WARC-Resource-Type` header, fixes #451

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -203,7 +203,7 @@ export class Recorder {
       });
       const reqresp = this.pendingReqResp(params.requestId, true);
       if (reqresp) {
-        reqresp.resourceType = params.this.addPageRecord(reqresp);
+        this.addPageRecord(reqresp);
 
         this.removeReqResp(params.requestId);
       }
@@ -333,7 +333,9 @@ export class Recorder {
     }
 
     const { url } = reqresp;
-    reqresp.resourceType = type;
+    if (type) {
+      reqresp.resourceType = type.toLowerCase();
+    }
 
     switch (errorText) {
       case "net::ERR_BLOCKED_BY_CLIENT":
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -1444,6 +1444,10 @@ function createResponse(
     "WARC-Page-ID": pageid,
   };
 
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Tested, looks good

ikreymer:
Only question for me is if: https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-ResourceType is standard enough.

There is the extension-focused: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType which is slightly more standardized (still differences between Chrome and Firefox) but its not a one-to-one mapping with the CDP resourceType that we have access to

tw4l:
> Only question for me is if: https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-ResourceType is standard enough.
> 
> There is the extension-focused: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType which is slightly more standardized (still differences between Chrome and Firefox) but its not a one-to-one mapping with the CDP resourceType that we have access to

It's a good question, but if the CDP resourceType is what we have access to and we have a reference we can point to, I think that's probably sufficient at this stage, especially since we are only crawling using Chromium-based browsers. Unless there's a very clear standard to map to cleanly, which I don't know that there is in this case.

ikreymer:
This is probably fine for now, but also opened an issue in WARC spec about possibly standardizing how this is approached. iipc/warc-specifications#96

ikreymer:
Per discussion, it appears that Puppeteer / Playwright both use resourceType as all lowercase. To match their convention, also setting WARC-Resource-Type header to be all lowercase and updating the `pageinfo` record to store the lowercase version as well.

PR #495, opened 2024-03-15
Fix Save/Load State

### LINKED ISSUE #491: Crawl resumed from saved state revisits already done pages
If I run a crawl and then send the node.js crawl process a SIGINT it writes a state YAML file into the `crawls/` directory. The README states that:

> The idea is that this crawl state YAML file can then be used as --config option to restart the crawl from where it was left of previously.

However the state file only seems to contain the list of queued URLs and does not include any that are already done:

```
state:
  done: 2
  queued:
    - '{"added":"2024-03-11T05:36:42.324Z","url":"https://site.example/page3","seedId":0,"depth":1}'
    - '{"added":"2024-03-11T05:36:42.324Z","url":"https://site.example/page4","seedId":0,"depth":1}'
  pending: []
  failed: []
  errors: []
```

When passing a state file to the --config option browsertrix-crawler seems to recrawl the entire site in a slightly different order. As far as I can tell the second run doesn't know which pages were done in the first run so it just queues them up again as soon as it encounters a link to them.

My expectation was that stopping and resuming from a state file should be roughly equivalent in terms of captured data to a crawl that was just never stopped.

**Example:**

```
$ podman run -it --rm -v $PWD:/crawls/ webrecorder/browsertrix-crawler:1.0.0-beta.7 crawl --id test -c test --combinewarc --generatecdx --url https://www.meshy.org/
```

In a different terminal stop the crawl by sending SIGINT to the crawl process. (Pressing CTRL+C doesn't exit gracefully as it kills the browser. Maybe that's podman/docker difference.)

```
$ pkill -INT -f /bin/crawl
```

Resume from state file

```
$ cat collections/test/crawls/crawl-20240311062844-test.yaml
state:
  done: 3
  queued:
    - '{"added":"2024-03-11T06:28:34.511Z","url":"https://www.meshy.org/blog/pywb-migration/","seedId":0,"depth":1,"extraHops":0}'
  pending: []
  failed: []
  errors: []
$ podman run -it --rm -v $PWD:/crawls/ webrecorder/browsertrix-crawler:1.0.0-beta.7 crawl --id test -c test --combinewarc --generatecdx --url https://www.meshy.org/ --config collections/test/crawls/crawl-20240311062844-test.yaml
```

Confirm that the same page was captured twice, once in each run:

```
$ grep '^org,meshy)/blog/outbackcdx-replication ' collections/test/indexes/index.cdxj
org,meshy)/blog/outbackcdx-replication 20240311062915 {"url": "https://www.meshy.org/blog/outbackcdx-replication/", "mime": "text/html", "status": "200", "digest": "sha256:3c91aad9db5f772528c32ffae302fc059d3e31de78c9235ce149d93bceac3c38", "length": "2155", "offset": "28193", "filename": "rec-b8254df4d6ed-20240311062902253-0.warc.gz"}
org,meshy)/blog/outbackcdx-replication 20240311062840 {"url": "https://www.meshy.org/blog/outbackcdx-replication/", "mime": "text/html", "status": "200", "digest": "sha256:3c91aad9db5f772528c32ffae302fc059d3e31de78c9235ce149d93bceac3c38", "length": "2160", "offset": "7558", "filename": "rec-dc8c7a02b0f6-20240311062832904-0.warc.gz"}
```

ikreymer:
- Fixes state serialization, which was missing the done list. Instead, adds a 'finished' list computed from the seen list, minus failed and queued URLs.
- Also adds serialization support for 'extraSeeds', seeds added dynamically from a redirect (via #475). Extra seeds are added to Redis and also included in the serialization.

Fixes #491

### DIFF
--- a/tests/saved-state.test.js
+++ b/tests/saved-state.test.js
@@ -95,6 +95,11 @@ test("check parsing saved state + page done + queue present", () => {
   expect(numQueued > 0).toEqual(true);
   expect(numDone + numQueued).toEqual(10);
 
+  // ensure extra seeds also set
+  expect(state.extraSeeds).toEqual([
+    `{"origSeedId":0,"newUrl":"https://www.iana.org/"}`,
+  ]);
+
   finished = state.finished;
 });
 
--- a/tests/saved-state.test.js
+++ b/tests/saved-state.test.js
@@ -9,23 +9,34 @@ function sleep(ms) {
 }
 
 async function waitContainer(containerId) {
-  execSync(`docker kill -s SIGINT ${containerId}`);
+  try {
+    execSync(`docker kill -s SIGINT ${containerId}`);
+  } catch (e) {
+    return;
+  }
+
+  // containerId is initially the full id, but docker ps
+  // only prints the short id (first 12 characters)
+  containerId = containerId.slice(0, 12);
 
   while (true) {
-    const res = execSync("docker ps -q", { encoding: "utf-8" });
-    if (res.indexOf(containerId) < 0) {
-      return;
+    try {
+      const res = execSync("docker ps -q", { encoding: "utf-8" });
+      if (res.indexOf(containerId) < 0) {
+        return;
+      }
+    } catch (e) {
+      console.error(e);
     }
     await sleep(500);
   }
 }
 
-var savedStateFile;
-var state;
-var numDone;
-var numQueued;
-var finished;
-var redis;
+let savedStateFile;
+let state;
+let numDone;
+let numQueued;
+let finished;
 
 test("check crawl interrupted + saved state written", async () => {
   let containerId = null;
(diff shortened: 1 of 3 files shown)


tests/saved-state.test.js
```diff
@@ -1,37 +1,51 @@
...
-  const wait = waitForProcess();
+  let containerId = null;
 
   try {
-    proc = exec(
-      "docker run -v $PWD/test-crawls:/crawls -v $PWD/tests/fixtures:/tests/fixtures webrecorder/browsertrix-crawler crawl --collection int-state-test --url https://webrecorder.net/ --limit 20",
-      { shell: "/bin/bash" },
-      wait.callback,
+    containerId = execSync(
+      "docker run -d -v $PWD/test-crawls:/crawls -v $PWD/tests/fixtures:/tests/fixtures webrecorder/browsertrix-crawler crawl --collection int-state-test --url https://iana.org/ --limit 10",
```
tw4l, line 46: Test looks good but does go back to relying on externally-hosted sites, which could be a problem down the road?

tw4l (approved):
Very nice, tested locally and working well. Thanks for the tests and cleanup here too.

tw4l:
I guess only other question is, could we not go back to using `done` rather than introducing a new `finished`, since we're already handling it as an int or array or strings for backwards compatibility's sake? No strong opinion here, but it seems we could just undo the breaking change rather than renaming.

tests/saved-state.test.js
ikreymer: I took another look, and we actually have a redirect on https://www.webrecorder.net/ -> https://webrecorder.net/ so can switch to that!

ikreymer:
> I guess only other question is, could we not go back to using `done` rather than introducing a new `finished`, since we're already handling it as an int or array or strings for backwards compatibility's sake? No strong opinion here, but it seems we could just undo the breaking change rather than renaming.

Yeah, I thought about that, technically could populate `done` from the pages list (it is supposed to be a full json dict rather than a single URL), but that requires either 1) reading and parsing pages.jsonl, and in case there's an issue (though there shouldn't be, have seen it if text is added) or 2) re-adding the done key in redis which would only be used for this.
But, that previous format still required that done be a JSON string but really we just need a single URL so this is more efficient, but perhaps later can switch to reading directly from pages.jsonl as well.

So overall, this is a better format, and chose a new key name to avoid confusion.

tests/saved-state.test.js
ikreymer: Updated!

PR #497, opened 2024-03-17
SAX-based sitemap parser

### LINKED ISSUE #496: Implement more efficient sitemap parsing.
The current sitemap parser reads the full XML into memory to parse it, and also doesn't handle sitemap indexes / nested sitemaps.
There is an alternative: https://www.npmjs.com/package/sitemap-stream-parser which uses SAX-based parsing and is much more efficient, but this parser has not been updated in a while. We should implement our own SAX-based parser.

ikreymer:
Adds a new SAX-based sitemap parser, inspired by: https://www.npmjs.com/package/sitemap-stream-parser

Supports:
- recursively parsing sitemap indexes, using p-queue to process N at a time (currently 5)
- `from` and `to` filter dates, to only include URLs between the given dates
- async parsing, continue parsing in the background after 100 URLs
- timeout for initial fetch / first 100 URLs set to 30 seconds to avoid slowing down the crawl
- save/load state integration: mark if sitemaps have already been parsed in redis, serialize to save state, to avoid reparsing again. (Will reparse if parsing did not fully finish)
- Aware of `pageLimit`, don't add URLs pass the page limit, interrupt further parsing when at limit.

Fixes #496 

TODO: Still need tests

### DIFF
--- a/src/util/sitemapper.ts
+++ b/src/util/sitemapper.ts
@@ -24,7 +24,7 @@ export class SitemapReader extends EventEmitter {
   fromDate?: Date;
   toDate?: Date;
 
-  q: PQueue;
+  queue: PQueue;
 
   seenSitemapSet: Set<string>;
   pending: Set<string>;
@@ -36,7 +36,7 @@ export class SitemapReader extends EventEmitter {
     super();
     this.headers = opts.headers;
 
-    this.q = new PQueue({ concurrency: SITEMAP_CONCURRENCY });
+    this.queue = new PQueue({ concurrency: SITEMAP_CONCURRENCY });
 
     this.fromDate = opts.fromDate;
     this.toDate = opts.toDate;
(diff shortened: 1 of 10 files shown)


tw4l:
Proper review coming, but in the meantime, this might also be a nice time to add a section on sitemap parsing to the crawler docs.

src/util/sitemapper.ts
```diff
@@ -0,0 +1,294 @@
...
+          parsingSitemap = false;
+          break;
+
+        default:
+          otherTags--;
+      }
+    });
+
+    parserStream.on("text", (text: string) => {
+      //console.log("text", text);
```
tw4l, line 182: Probably best to remove before merging

```suggestion
```

src/util/sitemapper.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    parserStream.on("end", async () => {
+      if (this.isRoot) {
+        await this.q.onIdle();
+      }
+      this.emit("end");
+    });
+
+    parserStream.on("opentag", (node: sax.Tag) => {
+      //console.log("open", node);
```
tw4l, line 106:
```suggestion
```

src/util/sitemapper.ts
```diff
@@ -0,0 +1,294 @@
...
+  limit?: number;
+};
+
+export class SitemapReader extends EventEmitter {
+  headers?: Record<string, string>;
+
+  fromDate?: Date;
+  toDate?: Date;
+
+  q: PQueue;
```
tw4l, line 33: Might be slightly better practice to name this `queue`? The type annotations really help clarify intention though, so feel free to ignore.

src/util/state.ts
```diff
@@ -793,4 +807,12 @@ return 0;
     }
     return seeds;
   }
+
+  async isSitemapDone() {
+    return (await this.redis.get(this.key + ":sitemapDone")) == "1";
```
tw4l, line 812: Maybe let's set the key as a const above, since it's repeated twice here?

tw4l (approved):
Nicely done! Left a few nitpicky comments, but I've tested this now on a good number of sitemaps, including some sitemap indices/nested sitemaps, and it's working great!

I pushed a commit with a docs update. Once we get some tests here, I think it's good to go :)

src/util/sitemapper.ts
ikreymer: done!

PR #500, opened 2024-03-18
profiles: handle terminate signals directly

ikreymer:
- add our own signal handling to create-login-profile to ensure fast exit in k8s
- print crawler version info string on startup


tw4l (approved):
Nice!

PR #503, opened 2024-03-21
Improved support for running as non-root

### LINKED ISSUE #502: /usr/bin/crawl Permission denied
I just happened to pull the latest image (1.0.0) and using the usual crawl command i got the error:

```
argv0-ignore: line 1: /usr/bin/crawl: Permission denied
```

ikreymer:
This PR provides improved support for running crawler as non-root, matching the user to the uid/gid of the crawl volume.

This fixes #502 initial regression from 0.12.4, where `chmod u+x` was used instead of `chmod a+x` on the node binary files.

However, that was not enough to fully support equivalent signal handling / graceful shutdown as when running with the same user. To make the running as different user path work the same way:
- need to switch to `gosu` instead of `su` (added in Brave 1.64.109 image)
- run all child processes as detached (redis-server, socat, wacz, etc..) to avoid them automatically being killed via SIGINT/SIGTERM
- running detached is controlled via `DETACHED_CHILD_PROC=1` env variable, set to 1 by default in the Dockerfile (to allow for overrides just in case)

A test has been added which runs one of the tests with a non-root `test-crawls` directory to test the different user path. The test (saved-state.test.js) includes sending interrupt signals and graceful shutdown and allows testing of those features for a non-root gosu execution.

Also bumping crawler version to 1.0.1

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,5 +44,7 @@ jobs:
         run: docker-compose build
       - name: mkdir test crawl dir
         run: mkdir test-crawls
+      - name: run stat
+        run: stat test-crawls
       - name: run jest
         run: sudo yarn test
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,10 @@ jobs:
         run: yarn run tsc
       - name: build docker
         run: docker-compose build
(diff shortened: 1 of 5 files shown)


tw4l (approved):
Very nice! Appreciate the extra test case, and graceful stopping is working much better now.

PR #505, opened 2024-03-22
Switch to using JS WACZ

### LINKED ISSUE #484: Use js-wacz to create WACZ files
Improvements for 1.0.0 branch of crawler:

- Switch from using py-wacz to [js-wacz ](https://github.com/harvard-lil/js-wacz) for WACZ generation
- Pass in indexes from `/tmp-cdx` rather than reindexing from WARCS
- Support creating indices with `--generateCDX` from `temp-cdx/` rather than having to reindex from the WARCs
- Delete `/tmp-cdx` after no longer needed

ikreymer:
Replaces dependencies on py-wacz with importing js-wacz natively.
Writes pages to either pages.jsonl (if seed) or extraPages.jsonl (if non-seed)
Uses streams for writing pages
Replaces --generateCDX with just moving tmp-cdx -> indexes
Removes any dependencies on python

Fixes #484 

Pending more testing and js-wacz release, using @tw4l branch for now!


src/crawler.ts
```diff
@@ -1477,29 +1479,28 @@ self.__bx_behaviors.selectMainBehavior();
     return false;
   }
 
-  awaitProcess(proc: ChildProcess) {
-    const stdout: string[] = [];
-    const stderr: string[] = [];
+  // todo: replace with js-wacz impl eventually
+  async _addCDXJ(wacz: WACZ) {
+    const dirPath = path.join(this.collDir, "tmp-cdx");
```
tw4l, line 1484: if `this.params.generateCDX`is set, the CDX files may already be in `indexes` instead of `tmp-cdx` here

src/crawler.ts
```diff
@@ -1295,36 +1300,50 @@ self.__bx_behaviors.selectMainBehavior();
...
-      } else {
-        logger.error("Error indexing and generating CDX", {
-          "status code": indexResult,
-        });
+      await this.crawlState.setStatus("generate-cdx");
+      await fsp.mkdir(dest, { recursive: true });
+
+      const tmpCdxList = await fsp.readdir(src);
+      for (const filename of tmpCdxList) {
+        await fsp.copyFile(path.join(src, filename), path.join(dest, filename));
```
tw4l, line 1346: This is a fairly significant change in outputs, from a single `index.cdjx` file or it's gzipped equivalent to a number of CDXJ files named `<warc-file>.warc.gz.cdx`.

1. We should consider whether there are ramifications for this, including if the CDXJ files are large enough that they should be gzipped, and whether it'll be a breaking change for any users.
2. Since the tmp-cdx files being copied over are actually CDXJ, we should use the more accurate `.cdxj` extension for these when writing them

tw4l: Resolved by using warcio-js's `CDXIndexer` to reindex from WARCS if `--generateCDX` is passed.

tw4l:
TODO:

- Add WACZ validation (not yet supported in js-wacz)
- Make CDXJ handling more memory-efficient in js-wacz (currently keeps all pages in memory, may OOM with large crawls)
- Possibly move CDXJ line handling in js-wacz from bin/cli.js into WACZ class

src/crawler.ts
```diff
@@ -1555,69 +1568,45 @@ self.__bx_behaviors.selectMainBehavior();
...
+    const waczLogger = new WACZLogger(logger);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const waczOpts: Record<string, any> = {
+      input: warcFileList.map((x) => path.join(this.archivesDir, x)),
+      output: waczPath,
+      pages: this.pagesDir,
+      detectPages: false,
+      indexFromWARCs: false,
+      logDirectory: this.logDir,
```
tw4l, line 1580: Seems like `logDirectory` may not be working as expected, not seeing log files in the resulting WACZs.

tw4l: Ah, it's not working because this isn't supported in js-wacz yet!

ikreymer:
Other existing difference: the warcio.js cdx contains status code as number instead of as string

tw4l:
> Other existing difference: the warcio.js cdx contains status code as number instead of as string, caught by current test failures.

I'm wondering if the solution here isn't just to change the tests to expect a number. Looking at the CDXJ specification, it looks like examples also use an int for status code, e.g.: https://specs.webrecorder.net/cdxj/0.1.0/#example

tw4l:
Closing in favor of #673 (WACZ generation approach has been changed, as documented in #674)

Also worth noting that as of https://github.com/webrecorder/warcio.js/pull/75, CDXJ created by warcio.js now uses strings consistently for `status`, `offset`, and `length`

PR #506, opened 2024-03-22
service worker capture fix: disable by default for now

ikreymer:
Due to issues with capturing top-level pages, make bypassing service workers the default for now. Previously, it was only disabled when using profiles. (This is also consistent with ArchiveWeb.page behavior).
Includes:
- add --serviceWorker option which can be `disabled`, disabled-if-profile (previous default) and `enabled`
- ensure page timestamp is set for direct fetch
- warn if page timestamp is missing on serialization, then set to now before serializing

bump version to 1.0.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "browsertrix-crawler",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "main": "browsertrix-crawler",
   "type": "module",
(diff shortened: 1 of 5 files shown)


tw4l (approved):
Works great! Tested with https://bitarchivist.net, and now seeing the full request/response records in the WARCs and timestamps for all files!

PR #507, opened 2024-03-23
Unify WARC writing + CDXJ indexing into single class

ikreymer:
Previously, we had the main WARCWriter as well as utility WARCResourceWriter that was used for screenshots, text, pageinfo and only generated resource records. This separate WARC writing path did not generate CDX, and used appendFile() to append new WARC records to an existing WARC.

This PR refactors into a single WARCWriter, which uses a writable stream to append records, and can also generate CDX on the fly. This PR is a pre-requisite to the js-wacz conversion (#484) since all WARCs need to have generated CDX.

However, as this is a big enough change, it made sense to refactor into its own PR for now.


src/util/warcwriter.ts
```diff
@@ -47,6 +55,8 @@ export class WARCWriter implements IndexerOffsetLength {
   }) {
     this.archivesDir = archivesDir;
     this.tempCdxDir = tempCdxDir;
+    // for now, disabling CDX
+    this.tempCdxDir = undefined;
```
tw4l, line 59: Maybe should undo this/enable CDX before merging? Or is there a reason to disable for now?

src/crawler.ts
```diff
@@ -2363,15 +2379,54 @@ self.__bx_behaviors.selectMainBehavior();
     }
   }
 
+  getWarcPrefix(defaultValue = "") {
```
tw4l, line 2382: Maybe we should set `rec` as the default value here rather than in the calling below?

src/crawler.ts
```diff
@@ -2363,15 +2379,54 @@ self.__bx_behaviors.selectMainBehavior();
...
+      logDetails,
+    });
+  }
+
   createRecorder(id: number): Recorder | null {
     if (!this.recording) {
       return null;
     }
 
+    const filenameBase = `${this.getWarcPrefix("rec")}$ts-${id}`;
```
tw4l, line 2421: See above re: default value, think it belongs better in the method signature.

src/crawler.ts
```diff
@@ -2363,15 +2379,54 @@ self.__bx_behaviors.selectMainBehavior();
...
+
+    const writer = this.createWarcWriter(filenameBase, true, { id: id + "" });
```
tw4l, line 2423:
```suggestion
    const writer = this.createWarcWriter(filenameBase, true, { id: id.toString() });
```

Might be a little clearer as to intention?

src/util/worker.ts
```diff
@@ -403,7 +403,6 @@ export async function runWorkers(
 ) {
   logger.info(`Creating ${numWorkers} workers`, {}, "worker");
 
-  const workers = [];
```
tw4l, line 406: Nice catch

tw4l (commented):
Tested and working well, just a few comments/questions

src/util/warcwriter.ts
ikreymer: oh, this was for merging into main/next release, just to be consistent in what we had before, where we disabled tmp-cdx.
It'll be reenabled in the js-warc branch.

src/crawler.ts
ikreymer: sure, yes, for the auxiliary writers, they could pass in empty string.

src/util/worker.ts
ikreymer: yep, already quickfix merged this into main before as well.

src/crawler.ts
ikreymer: though, this will result in:
```
this.getWarcPrefix() meaning "rec"
this.getWarcPrefix("") meaning ""
```
not sure if its clearer this way -- could also remove the default value altogether.

tw4l: Ah right, eh, I think it's fine to leave as-is given the auxiliary writers, you make a good point.

tw4l (approved):

PR #509, opened 2024-03-24
fixes redirected seed (from #475) being counted againt page limit:

### LINKED ISSUE #508: Seed redirect causes one less URL to be crawled due to limit
A follow up to #475, currently adding extra redirect seeds are not subtracted from the limit, resulting in one less page being crawled if one of the pages redirects, eg. with 1 seed and limit of 10, if that seed redirects, only 9 pages total will be crawled.

The solution is to subtract the number of 'extra seeds' from the seen list when computing the limit

ikreymer:
- subtract extraSeeds when computing limit
- don't include redirect seeds in seen list when serializing
- tests: adjust saved-state-test to also check total pages when crawl is done

fixes #508
(for 1.0.3 release)

### DIFF
--- a/src/util/state.ts
+++ b/src/util/state.ts
@@ -91,6 +91,7 @@ declare module "ioredis" {
       pkey: string,
       qkey: string,
       skey: string,
+      esKey: string,
       url: string,
       score: number,
       data: string,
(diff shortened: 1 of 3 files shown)


src/util/state.ts
```diff
@@ -702,9 +705,9 @@ return 0;
     return parseInt(done || "0");
   }
 
-  async numSeen() {
-    return await this.redis.scard(this.skey);
-  }
+  // async numSeen() {
+  //   return await this.redis.scard(this.skey);
+  // }
```
tw4l, line 710: I guess this isn't used anymore, but maybe we just remove it? Pretty straightforward to add back if needed.

tw4l (approved):
These changes look good! The exclusions test failure seems to be an issue with with `1.0.3-release` branch rather than these changes, can see it in https://github.com/webrecorder/browsertrix-crawler/pull/511 as well. Should fix that before release.

PR #514, opened 2024-03-26
Use RFC2606 invalid domain names

vnznznz:
`invalid.dev` can potentially be registered and used. `.invalid` is guaranteed to never be valid. See also: https://www.rfc-editor.org/rfc/rfc2606.html

### DIFF
--- a/src/util/browser.ts
+++ b/src/util/browser.ts
@@ -593,8 +593,8 @@ export const defaultArgs = [
   "--no-service-autorun",
   "--export-tagged-pdf",
   "--apps-keep-chrome-alive-in-tests",
-  "--apps-gallery-url=https://invalid.webstore.example.com/",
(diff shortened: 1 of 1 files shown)


tw4l (approved):
Thank you!

tw4l:
Might want to change the base branch of this to `1.0.3-release` and then merge into main from there.

ikreymer:
> Might want to change the base branch of this to `1.0.3-release` and then merge into main from there.

Eh, do you think this is urgent enough for 1.0.3?

tw4l:
> > Might want to change the base branch of this to `1.0.3-release` and then merge into main from there.
> 
> Eh, do you think this is urgent enough for 1.0.3?

Eh maybe not actually, given that the related issue seems unrelated to the URL being used.

Issue #446, opened 2023-12-05
Cloudflare interstitial wait isn't working


vbanos:
`browsertrix-crawler` is supposed to detect Cloudflare with this function:
https://github.com/webrecorder/browsertrix-crawler/blob/c3b98e5047ea219336883b0b1969da425fc43456/crawler.js#L1376
but it doesn't work correctly. Example site: https://www.defence-point.gr
See the final capture:
![image](https://github.com/webrecorder/browsertrix-crawler/assets/5803877/179d401c-0989-4dcc-b5a9-6ebbada848b0)

tw4l:
Interesting, I can repro but I see an additional message in the replay:

<img width="981" alt="Screen Shot 2023-12-13 at 11 16 22 AM" src="https://github.com/webrecorder/browsertrix-crawler/assets/6758804/58891ff9-1f94-486c-8b38-44dae02dc5cd">

tw4l:
Not seeing the same issue navigating to the site with the same version of Brave Browser manually, seems to be something with either our handling of the Cloudflare check or new behavior on their end. We are investigating.

tw4l:
@ikreymer has been doing research into this and we should be able to get a fix but it will require some R&D so may take a little time

wvengen:
I saw this once (but on a re-crawl it was not a problem). Just in case it is useful for anyone, here is the WACZ for the crawl where it got stuck on the CF page (renamed to zip to allow uploading): [20240314104054731-f9133c99-4da-0.zip](https://github.com/webrecorder/browsertrix-crawler/files/14615095/20240314104054731-f9133c99-4da-0.zip)

ikreymer:
Good news, it should be possible to at least load this page in a profile, and then save the profile with cookies to avoid this.
Previously, it was even intercepting the profile loading.
The change requires filtering out some browser targets from being captured by puppeteer, and disabling a flag that is generally used, so may need more work to work automatically.
But at least it should be possible with the profile workflow.

ikreymer:
I think this is probably as much as we can do with this for now - the profile option seems like a good solution, and still works well. I believe Cloudflare is able to detect puppeteer use somehow through timing when it attaches to each target, which we can filter out in profile creation mode.
Short of removing puppeteer altogether, I think this is best solution for the moment as it also avoids the Cloudflare page altogether if first creating a profile.

PR #518, opened 2024-03-28
avoid cloudflare detection of puppeteer when using browser profiles:

ikreymer:
- filter out 'other' / no url targets from puppeteer attachment
- disable '--disable-site-isolation-trials' for profiles
- workaround for #446 with profiles
- also fixes `pageExtraDelay` not working for non-200 responses - may be useful for detecting captcha blocked pages.
- connect VNC right away instead of waiting for page to fully finish loading, hopefully resulting in faster profile start-up time.

Can test profile creation with: https://defence-point.gr/ and https://nopecha.com/demo/turnstile to ensure that they now work.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -891,11 +891,7 @@ self.__bx_behaviors.selectMainBehavior();
 
     data.loadState = LoadState.EXTRACTION_DONE;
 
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Tested and working well, thanks!

PR #520, opened 2024-03-28
add an extra --postLoadDelay param to specify how many seconds to wait after page-load

### LINKED ISSUE #519: Content-Length: 0 in text.warc instead of text like in rec.warc
in Btrix Version 1.0.2
In the regular warc File the content is their as expected;
But the generated text.warc contains no content only the message content-length 0
And: This problem occurs with Instagram, not with Facebook! With the same settings!
Addiditonal-hint: I also messed around with --text option and tried it several times, but none of the --text options was forcing a text.warc that had any content (should only --text work also? it's not generating text.warc and not adding text in the page.json)

This crawl i use to validate my browser profiles are still loged in, an it is very beneficial to only grep on the text.warc.gz file and not the originally rec-xxx.warc.gz file

Please find attached one try; (The archive result is always the same)

And finally the docker start command:
docker run -d --name ONB_Btrix_validate_insta_profile_20240328105141 -e NODE_OPTIONS="--max-old-space-size=32768" -p 9438:9438 -p 17412:17412  -v /home/netarchive/browsertrix/crawls/:/crawls/ webrecorder/browsertrix-crawler:1.0.2 crawl --screencastPort 9438 --seedFile /crawls/config/validate_insta_profile_seeds.txt  --profile /crawls/profiles/insta.tar.gz --scopeType prefix --depth 0 --extraHops 0 --workers 1 --healthCheckPort 17412 --limit 2 --sizeLimit 20971520 --timeLimit 300 --delay 15 --waitUntil networkidle0  --saveState always   --behaviorTimeout 10 --pageLoadTimeout 10  --text to-warc --logging stats,info --warcInfo ONB_CRAWL_validate_insta_profile_Depth_0_20240328105141 --userAgentSuffix +ONB_Bot_Btrix_1.0.2, webarchiv@onb.ac.at --crawlId id_ONB_CRAWL_validate_insta_profile_Depth_0_20240328105141 --collection validate_insta_profile_20240328105141

[validate_insta_profile_20240328105141.zip](https://github.com/webrecorder/browsertrix-crawler/files/14787576/validate_insta_profile_20240328105141.zip)

ikreymer:
but before running link extraction, text extraction, screenshots and behaviors.

Useful for sites that load quickly but perform async loading / init afterwards, fixes #519

A simple workaround for when it's tricky to detect when a page has actually fully loaded.
Useful for sites such as Instagram.

### DIFF
--- a/docs/docs/user-guide/common-options.md
+++ b/docs/docs/user-guide/common-options.md
@@ -10,6 +10,13 @@ See [page.goto waitUntil options](https://pptr.dev/api/puppeteer.page.goto#remar
 
 The `--pageLoadTimeout`/`--timeout` option sets the timeout in seconds for page load, defaulting to 90 seconds. Behaviors will run on the page once either the page load condition or the page load timeout is met, whichever happens first.
 
+### Additional Wait
+
+Occasionally, a page may seem to have loaded, but performs dynamic initialization / additional loading. This is can be hard to detect, and the `--postLoadDelay` flag
+can be used to specify additional seconds to wait after the page appears to have loaded, before moving on to post-processing actions, such as link extraction, screenshotting and text extraction (see below).
+
+(On the other hand, the `--pageExtraDelay`/`--delay` adds an extra after all post-load actions have taken place, and can be useful for rate-limiting.)
+
 ## Ad Blocking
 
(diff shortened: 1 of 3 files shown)


tw4l (approved):

tw4l:
May want to add a mention in the docs? Otherwise looks great!

PR #523, opened 2024-04-02
Make /app world-readable to better support non-root usage

vnznznz:
I suspect https://github.com/webrecorder/browsertrix-cloud/pull/1625 is failing in CI because browsertrix-crawler sometimes needs to read from `/app` after launching. This causes errors with the non-root deployment.


ikreymer:
Hm, we do set permissions on some of the files already, but I guess doesn't hurt to try. Can try it in the next beta release.

ikreymer:
Actually, as is, this is slowing down the build, by 1.5min, as the chmod is expensive. We may need to find out which files need to be world-readable if this is the issue.

PR #524, opened 2024-04-02
behaviors: ensure autofetch and autoplay behaviors start as soon as p…

ikreymer:
…age is loaded, same as 0.12.x behavior

this results in proper wait time for initial page load, fixes issues such as #519


tw4l:
Gave notes offline but in my testing this is not resolving the issue in #519, with or without a profile

ikreymer:
Yeah, this may have been a fluke initially, not worth merging for release w/o consistent repro :/

PR #525, opened 2024-04-02
refactor handling of max size for html/js/css

### LINKED ISSUE #522: Website fails to fully load properly since 1.0.0
I'm doing some tests on a demo Confluence website: https://octo-help.refined.com (not sure that the fact it is a demo Confluence has some importance).

The website seems to fail to load properly since 1.0.0 release.

This command succeeds to extract multiple pages:
```
docker run --rm webrecorder/browsertrix-crawler:0.12.3 crawl --url "https://octo-help.refined.com"
```

This commands loads only seed page and then stops, obviously failing to wait for page to be fully loaded
```
docker run --rm webrecorder/browsertrix-crawler:1.0.0 crawl --url "https://octo-help.refined.com"
```

I experimented with `--waitUntil load,networkidle0` without any success.

I experimented with various 1.0.0 beta images and 1.1.0 beta images, without any success as well.

I created a login profile with 1.1.0-beta.3 and the website is fully loaded in the interactive VNC display, so it is clearly working. Passing the login profile to the crawl does not help.

ikreymer:
- due to a typo (and lack of type-checking!) incorrectly passed in matchFetchSize instead of maxFetchSize, resulting in text/css/js for >5MB instead of >25MB not properly streamed back to the browser
- add type checking to AsyncFetcherOptions to avoid this in the future.
- refactor to avoid checking size altogether for 'essential resources', html(document), js and css, instead always fetch them fully and continue in the browser. Only apply rewriting if <25MB.
- fixes #522

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -606,7 +606,7 @@ export class Recorder {
       : MAX_BROWSER_DEFAULT_FETCH_SIZE;
 
     if (contentLen < 0 || contentLen > maxFetchSize) {
-      logger.warn(
+      logger.debug(
         "Large/unknown resource fetch",
         { url, contentLen, contentType, resourceType },
         "recorder",
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -596,13 +596,11 @@ export class Recorder {
 
     let streamingConsume = false;
 
-    const contentType = this._getContentType(responseHeaders);
-
     // if contentLength is large or unknown, do streaming, unless its an essential resource
     // in which case, need to do a full fetch either way
     if (
       (contentLen < 0 || contentLen > MAX_BROWSER_DEFAULT_FETCH_SIZE) &&
-      !this.isEssentialResource(reqresp.resourceType, contentType)
+      !this.isEssentialResource(reqresp.resourceType)
     ) {
       const opts: ResponseStreamAsyncFetchOptions = {
         tempdir: this.tempdir,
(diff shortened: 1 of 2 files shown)


src/util/recorder.ts
```diff
@@ -577,23 +596,30 @@ export class Recorder {
...
     // set max fetch size higher for HTML responses for current page
-    const matchFetchSize = this.allowLargeContent(contentType)
+    const maxFetchSize = this.allowLargeContent(resourceType, contentType)
       ? MAX_BROWSER_TEXT_FETCH_SIZE
       : MAX_BROWSER_DEFAULT_FETCH_SIZE;
 
-    if (contentLen < 0 || contentLen > matchFetchSize) {
-      const opts = {
+    if (contentLen < 0 || contentLen > maxFetchSize) {
+      logger.warn(
```
tw4l, line 609: I'm seeing a lot of these in the logs, maybe should be `debug`? It gets quite verbose as-is.

tw4l:
Confirmed it resolves the issue with the example in #522. I'd recommend setting a `--limit` on the crawl when testing.

src/util/recorder.ts
ikreymer: Yes, not really warning

ikreymer, line 609:
```suggestion
      logger.debug(
```

tw4l:
Changes make sense and tested that it's still working well with example site now. Looks good to merge to me.

tw4l (approved):

PR #528, opened 2024-04-04
ensure all warcwriter write operations go through a queue.

ikreymer:
Currently, only the recorder's WARCWriter writes records through a queue, resulting in other WARCs potentially suffering from concurrent write attempts. This fixes that by:
- adding the concurrent queue to WARCWriter itself
- all writeRecord, writeRecordPair, writeNewResourceRecord calls are first added to the PQueue, which ensures writes happen in order and one-at-a-time
- flush() also ensures queue is empty/idle
- should avoid any issues with concurrent writes to any WARC

### DIFF
--- a/src/replaycrawler.ts
+++ b/src/replaycrawler.ts
@@ -675,7 +675,7 @@ export class ReplayCrawler extends Crawler {
         (state as ComparisonPageState).comparison = comparison;
       }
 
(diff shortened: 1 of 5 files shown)


tw4l (approved):
Thanks for catching this! Tested and code looks good

PR #534, opened 2024-04-11
Adblock support

ikreymer:
Now that RWP 2.0.0 with adblock support has been released (webrecorder/replayweb.page#307), this enables adblock on the QA mode RWP embed, to get more accurate screenshots.
Fetches the adblock.gz directly from RWP (though could also fetch it separately from Easylist)
Updates to 1.1.0-beta.5

To test: QA a crawl that was crawled with ad-blocking (eg. a site like cnn.com) and verify that ads are being blocked in screenshots / via screencasting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,9 +48,10 @@ ADD config/ /app/
 
 ADD html/ /app/html/
 
-ARG RWP_VERSION=1.8.15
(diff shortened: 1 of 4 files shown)


tw4l (approved):
Confirmed this is working to block ads in QA replay, as evidenced by screenshots.

PR #542, opened 2024-04-13
Fix for --rolloverSize for individual WARCs in 1.x

### LINKED ISSUE #533: --rolloverSize Default seems not to apply on warc.gz
Hi, in version 1.0.2 in the archive folder we find only one warc.gz file As example it looks like this:
image
But I also got bigger once, however when I read about the rolloverSize parameter i was interpretating it the following way:
If the warc.gz file is bigger then 1000000000 Byte (~953.67431640625 MB) a second file should be generated like logrotating 

From the Docs (9.4.2024):
  --rolloverSize                        If set, declare the rollover size [number] [default: 1000000000]

Hint: --combineWarc was not setted!
Here is the entire starting command:
docker run -d --name ONB_Btrix_exxpress_entirely_daily_20240329081001 -e NODE_OPTIONS="--max-old-space-size=32768" -p 9240:9240 -p 14360:14360  -v /home/netarchive/browsertrix/crawls/:/crawls/ webrecorder/browsertrix-crawler:1.0.2 crawl --screencastPort 9240 --seedFile /crawls/config/exxpress_entirely_daily_seeds.txt   --scopeType prefix --depth 50 --extraHops 0 --workers 1 --healthCheckPort 14360  --delay 1 --waitUntil networkidle0  --saveState always   --limit 24500  --timeLimit 86400  --logging stats,info --warcInfo ONB_CRAWL_exxpress_entirely_daily_Depth_50_20240329081001 --userAgentSuffix +ONB_Bot_Btrix_1.0.2, webarchiv@onb.ac.at --crawlId id_ONB_CRAWL_exxpress_entirely_daily_Depth_50_20240329081001 --collection exxpress_entirely_daily_20240329081001

I want also to stress this, that if there is an error where the warc is corrupted the damaged data is much bigger if no rotation takes place

Maybe i am missinterpreting this setting, i tried to use it with different values, but Version 1.0.4 nor 1.0.2 never generated more then One Warc File

ikreymer:
Fixes #533 

Fixes rollover in WARCWriter, separate from combined WARC rollover size:
- check rolloverSize and close previous WARCs when size exceeds
- add timestamp to resource WARC filenames to support rollover, eg. screenshots-{ts}.warc.gz
- use append mode for all write streams, just in case
- tests: add test for rollover of individual WARCs with 500K size limit
- tests: update screenshot tests to account for WARCs now being named screenshots-{ts}.warc.gz instead of just screenshots.warc.gz

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -454,7 +454,7 @@ export class Crawler {
     await fsp.mkdir(this.tempdir, { recursive: true });
     await fsp.mkdir(this.tempCdxDir, { recursive: true });
 
-    this.logFH = fs.createWriteStream(this.logFilename);
+    this.logFH = fs.createWriteStream(this.logFilename, { flags: "a" });
     logger.setExternalLogStream(this.logFH);
 
     this.infoString = await getInfoString();
@@ -2484,7 +2484,7 @@ self.__bx_behaviors.selectMainBehavior();
   }
 
   createExtraResourceWarcWriter(resourceName: string, gzip = true) {
-    const filenameBase = `${this.getWarcPrefix()}${resourceName}`;
+    const filenameBase = `${this.getWarcPrefix()}${resourceName}-$ts`;
 
     return this.createWarcWriter(filenameBase, gzip, { resourceName });
   }
--- a/src/util/warcwriter.ts
+++ b/src/util/warcwriter.ts
@@ -85,14 +85,15 @@ export class WARCWriter implements IndexerOffsetLength {
   private async initFH() {
     if (this.filename && this.offset >= this.rolloverSize) {
       const oldFilename = this.filename;
+      const size = this.offset;
       this.filename = this._initNewFile();
       logger.info(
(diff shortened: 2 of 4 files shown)


tw4l (approved):
Looks good and testing well!

Only thought is that because we're only rolling over once the size has been exceeded, occasionally a file gets quite a bit over the rollover size before a new one is started. Maybe eventually we could look into optimizing that a little, but I think for now this is a good improvement.

ikreymer:
> Looks good and testing well!
> 
> Only thought is that because we're only rolling over once the size has been exceeded, occasionally a file gets quite a bit over the rollover size before a new one is started. Maybe eventually we could look into optimizing that a little, but I think for now this is a good improvement.

Yeah, we could potential adjust this to account for size after next records are written, but that's a bit more work, and generally this should be pretty good. For small rollover limits, there's always chance of a file exceeding it, eg. if a 100MB video is downloaded and limit is 10MB, there will still be a 100MB WARC

Issue #515, opened 2024-03-26
S3 upload and region error


cmillet2127:
Hello,
I made some tests with S3 upload. It works very well with a bucket stored in us-east-1, but doesn't work for other regions, in my case eu-west-3.
I try to analyze the code, and it seems it failed in MinIO call in util/storage.ts line 90:
```
    await this.client.fPutObject(
      this.bucketName,
      this.objectPrefix + targetFilename,
      srcFilename,
    );
```
where the region can optionnally be provided, otherwise us-east-1 is used by default.

The region can be extracted from the STORE_ENDPOINT_URL. example : https://\<bucket\>.s3.\<region\>.amazonaws.com/\<bucket\>/

Best regards

ikreymer:
We are planning to switch to AWS CLI SDK, I think this will be addressed with that change. See: #479

tw4l:
Hi @cmillet2127, based on [a discussion in the minio-js repo](https://github.com/minio/minio-js/issues/619#issuecomment-326158139) I think the crawler should work as-is and minio-js will autodiscover the bucket if you use `s3.amazonaws.com` as the STORE_ENDPOINT_URL. Want to try that out and let us know if it works?

RomanSmolka:
The change in #543 has broken it for me in the 'eu-central-1' region. Now I'm getting an error:

`S3Error: The authorization header is malformed; the region 'auto' is wrong; expecting 'eu-central-1'`

After going back to 1.1.0 Beta 5, it's working fine. This is my env configuration:

```
environment: [
  { name: 'STORE_ENDPOINT_URL', value: 'https://s3.amazonaws.com/' + process.env.S3_BUCKET },
  { name: 'STORE_PATH', value: '/' },
  { name: 'STORE_FILENAME', value: item.s3_key },
  { name: 'STORE_ACCESS_KEY', value: process.env.S3_ACCESS_KEY },
  { name: 'STORE_SECRET_KEY', value: process.env.S3_SECRET },
  { name: 'CRAWL_ID', value: crawlId },
  { name: 'WEBHOOK_URL', value: process.env.WEBRECORDER_HOOK }
]
```

Perhaps just adding a STORE_REGION variable would be a better solution? @ikreymer @tw4l

ikreymer:
[...]

Issue #479, opened 2024-03-04
Failure uploading large files (handling slowDown)


wvengen:
During a large crawl (2GB+), I get stuck in the "Uploading WACZ" stage (using OpenStack SWIFT S3 for storage). The log shows

```jsonl
{"timestamp":"2024-03-04T12:07:55.250Z","logLevel":"debug","context":"general","message":"WACZ successfully generated and saved to: /crawls/collections/thecrawl/thecrawl.wacz","details":{}}
{"timestamp":"2024-03-04T12:07:55.255Z","logLevel":"info","context":"s3Upload","message":"S3 file upload information","details":{"bucket":"x","crawlId":"x","prefix":"x/"}}
{"timestamp":"2024-03-04T12:08:03.027Z","logLevel":"error","context":"general","message":"Crawl failed","details":{"type":"exception","message":"","stack":"S3Error\n    at Object.parseError (/app/node_modules/minio/dist/main/xml-parsers.js:79:11)\n    at /app/node_modules/minio/dist/main/transformers.js:165:22\n    at DestroyableTransform._flush (/app/node_modules/minio/dist/main/transformers.js:89:10)\n    at DestroyableTransform.prefinish (/app/node_modules/readable-stream/lib/_stream_transform.js:123:10)\n    at DestroyableTransform.emit (node:events:514:28)\n    at prefinish (/app/node_modules/readable-stream/lib/_stream_writable.js:569:14)\n    at finishMaybe (/app/node_modules/readable-stream/lib/_stream_writable.js:576:5)\n    at endWritable (/app/node_modules/readable-stream/lib/_stream_writable.js:594:3)\n    at Writable.end (/app/node_modules/readable-stream/lib/_stream_writable.js:535:22)\n    at IncomingMessage.onend (node:internal/streams/readable:705:10)"}}
{"timestamp":"2024-03-04T12:08:03.036Z","logLevel":"info","context":"general","message":"Exiting, Crawl status: failing","details":{}}
```

and the error message mentioned is
```
S3Error
    at Object.parseError (/app/node_modules/minio/dist/main/xml-parsers.js:79:11)
    at /app/node_modules/minio/dist/main/transformers.js:165:22
    at DestroyableTransform._flush (/app/node_modules/minio/dist/main/transformers.js:89:10)
    at DestroyableTransform.prefinish (/app/node_modules/readable-stream/lib/_stream_transform.js:123:10)
    at DestroyableTransform.emit (node:events:514:28)
[...]

PR #543, opened 2024-04-13
allow minio to connect to other regions

mguella:
This should address the issue of connecting to buckets stored outside us-east-1 (https://github.com/webrecorder/browsertrix-crawler/issues/515) while the switch from Minio client to AWS SDK is being worked on (https://github.com/webrecorder/browsertrix-crawler/issues/479)

### DIFF
--- a/src/util/storage.ts
+++ b/src/util/storage.ts
@@ -65,6 +65,7 @@ export class S3StorageSync {
       accessKey,
       secretKey,
       partSize: 100 * 1024 * 1024,
+      region: "auto",
     });
 
     this.bucketName = url.pathname.slice(1).split("/")[0];


tw4l:
Hi @mguella, see my comment here: https://github.com/webrecorder/browsertrix-crawler/issues/515#issuecomment-2057841290

Based on the minio-js developer's response, I think this change may be unnecessary if one uses `s3.amazonaws.com` as the STORE_ENDPOINT_URL, as minio-js claims to auto-discover the region. But it could use some testing.

mguella:
@tw4l unfortunately that doesn't work with non AWS providers (I've tried both Cloudflare and Tigris) and I think it would be nice to support other S3 compatible storage other than AWS and Minio itself

PR #551, opened 2024-04-18
Add crawler QA docs

### LINKED ISSUE #550: Add documentation for QA!

tw4l:
Fixes #550

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -29,7 +29,7 @@ Browsertrix Crawler is a command line application responsible for the core featu
 - Extensible Puppeteer driver script for customizing behavior per crawl or page.
 - Ability to create and reuse browser profiles interactively or via automated user/password login using an embedded browser.
 - Multi-platform support — prebuilt Docker images available for Intel/AMD and Apple Silicon (M1/M2) CPUs.
-- Quality Assurance Crawling - crawl the replay of existing crawls (via WACZ) and produce comparison stats.
+- Quality Assurance (QA) crawling - crawl the replay of existing crawls (via WACZ) and produce stats comparing what the browser encountered on a website during crawling against the replay of the crawl WACZ.
 
 ## Documentation
 
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -29,6 +29,7 @@ Browsertrix Crawler is a command line application responsible for the core featu
 - Extensible Puppeteer driver script for customizing behavior per crawl or page.
 - Ability to create and reuse browser profiles interactively or via automated user/password login using an embedded browser.
 - Multi-platform support — prebuilt Docker images available for Intel/AMD and Apple Silicon (M1/M2) CPUs.
+- Quality Assurance Crawling - crawl the replay of existing crawls (via WACZ) and produce comparison stats.
 
 ## Documentation
 
--- a/docs/docs/user-guide/qa.md
+++ b/docs/docs/user-guide/qa.md
@@ -1,0 +1,4 @@
+# Quality Assurance Analysis Crawling
+
+Browsertrix Crawler also has the capability to crawl or 're-crawl' over an existing crawl, to produce comparsion
+resource records long with comparison data for the crawl vs replay along several dimensions.
--- a/docs/docs/user-guide/qa.md
+++ b/docs/docs/user-guide/qa.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-Browsertrix Crawler has the capability to "re-crawl" an existing crawl to compare what the browser encountered on a website during crawling against the replay of the crawl WACZ. The WACZ produced by this analysis run includes additional comparison data (stored as WARC `resource` records) for the pages found during crawling against their replay in ReplayWeb.page along several dimensions, including screenshot, text comparisons and resource compar
+Browsertrix Crawler has the capability to "re-crawl" an existing crawl to compare what the browser encountered on a website during crawling against the replay of the crawl WACZ. The WACZ produced by this analysis run includes additional comparison data (stored as WARC `resource` records) for the pages found during crawling against their replay in ReplayWeb.page along several dimensions, including screenshot, text comparisons and resource comparison.
 
 QA features described on this page are available in Browsertrix Crawler releases 1.1.0 and later.
 
--- a/docs/docs/user-guide/qa.md
+++ b/docs/docs/user-guide/qa.md
@@ -8,7 +8,7 @@ QA features described on this page are available in Browsertrix Crawler releases
 
 ## Getting started
 
-To analyze an existing crawl, call Browsertrix Crawler with the `qa` endpoint, passing the original crawl WACZ as the `qaSource`:
+To analyze an existing crawl, call Browsertrix Crawler with the `qa` entrypoint, passing the original crawl WACZ as the `qaSource`:
 
 ```sh
 docker run -v $PWD/crawls/:/crawls/ -it webrecorder/browsertrix-crawler qa --qaSource /crawls/collections/example-crawl/example-crawl.wacz --collection example-qa
--- a/docs/docs/user-guide/qa.md
+++ b/docs/docs/user-guide/qa.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-Browsertrix Crawler has the capability to "re-crawl" an existing crawl to compare what the browser encountered on a website during crawling against the replay of the crawl WACZ. The WACZ produced by this analysis run includes additional comparison data (stored as WARC `resource` records) for the pages found during crawling against their replay in ReplayWeb.page along several dimensions, including screenshot, text comparisons and resource comparison.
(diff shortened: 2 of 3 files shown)


docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `qaSource` can be:
+- A local WACZ file path or a URL
+- A single WACZ or a JSON file containing a list of WACZ files in the `resources` json (Multi-WACZ)
+
+This assumes an existing crawl that was created in the `example-crawl` collection.
+
+A new WACZ for the analysis run will be created in the resulting `example-qa` collection.
+
+By default, the QA crawl will visit all of the pages (as read from the source WACZ file(s)), however pages can further be limited by adding `--include` and `--exclude` regexes. The `--limit` flag will also limit how pages are tested.
```
tw4l, line 28:
```suggestion
By default, the QA crawl will visit all of the pages (as read from the source WACZ file(s)), however pages can further be limited by adding `--include` and `--exclude` regexes. The `--limit` flag will also limit how many pages are tested.
```

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,84 @@
+# Quality Assurance
+
+## Overview
+
+Browsertrix Crawler has the capability to "re-crawl" an existing crawl to compare what the browser encountered on a website during crawling against the replay of the crawl WACZ. The WACZ produced by this analysis run includes additional comparison data (stored as WARC `resource` records) for the pages found during crawling against their replay in ReplayWeb.page along several dimensions, including screenshot, extracted text, and page resource comparisons.
```
Shrinks99, line 5: Splits up sentence, a bit long otherwise.  Uses "analysis" instead of "re-crawling", consistency with Browsertrix

```suggestion
Browsertrix Crawler can analyze an existing crawl to compare what the browser encountered on a website during crawling against the replay of the crawl WACZ. The WACZ produced by this analysis run includes additional comparison data (stored as WARC `resource` records) for the pages found during crawling against their replay in ReplayWeb.page. This works along several dimensions, including screenshot, extracted text, and page resource comparisons.
```

docs/docs/index.md
```diff
@@ -29,6 +29,7 @@ Browsertrix Crawler is a command line application responsible for the core featu
 - Extensible Puppeteer driver script for customizing behavior per crawl or page.
 - Ability to create and reuse browser profiles interactively or via automated user/password login using an embedded browser.
 - Multi-platform support — prebuilt Docker images available for Intel/AMD and Apple Silicon (M1/M2) CPUs.
+- Quality Assurance (QA) crawling - crawl the replay of existing crawls (via WACZ) and produce stats comparing what the browser encountered on a website during crawling against the replay of the crawl WACZ.
```
Shrinks99, line 32:
```suggestion
- Quality Assurance (QA) crawling — analyze the replay of existing crawls (via WACZ) and produce stats comparing what the browser encountered on a website during crawling against the replay of the crawl WACZ.
```

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,84 @@
...
+
+!!! note
+
+    QA features described on this page are available in Browsertrix Crawler releases 1.1.0 and later.
```
Shrinks99, line 9: I guess these are good because we don't version our docs?  No harm in them really, though as we add new features and reference them on pages like this, they run the risk of going out of date?

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+## Getting started
+
+To be able to run QA on a crawl, we must first have an existing crawl, for example:
```
Shrinks99, line 13: Not mentioned in styleguide, but ([stolen from Google's documentation guidelines](https://developers.google.com/style/pronouns#personal-pronouns)) "we" should refer to the author IE Webrecorder.  I think I need to update the section on "you" statements, which was really about brevity and not a hard rule :\

```suggestion
To be able to run QA on a crawl, you must first have an existing crawl, for example:
```

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,84 @@
...
+
+```sh
+docker run -v $PWD/crawls:/crawls/ -it webrecorder/browsertrix-crawler crawl --url https://webrecorder.net/ --collection example-crawl --text to-warc --screenshot view --generateWACZ```
+
+Note that this crawl must be run with `--generateWACZ` flag as QA requires a WACZ to work with, and also ideally the `--text to-warc` and `--screenshot view` flags as well (see below for more details on comparison dimensions).
+
+To analyze this crawl, call Browsertrix Crawler can be run with the `qa` entrypoint, passing the original crawl WACZ as the `qaSource`:
```
Shrinks99, line 20:
```suggestion
To analyze this crawl, call Browsertrix Crawler with the `qa` entrypoint, passing the original crawl WACZ as the `qaSource`:
```

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,84 @@
...
+
+The `qaSource` can be:
+- A local WACZ file path or a URL
+- A single WACZ or a JSON file containing a list of WACZ files in the `resources` json (Multi-WACZ)
+
+This assumes an existing crawl that was created in the `example-crawl` collection.
+
+A new WACZ for the analysis run will be created in the resulting `example-qa` collection.
+
+By default, the QA crawl will visit all of the pages (as read from the source WACZ file(s)), however pages can further be limited by adding `--include` and `--exclude` regexes. The `--limit` flag will also limit how many pages are tested.
```
Shrinks99, line 34:
```suggestion
By default, the analysis crawl will visit all of the pages (as read from the source WACZ file(s)), however pages can further be limited by adding `--include` and `--exclude` regexes. The `--limit` flag will also limit how many pages are tested.
```

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,84 @@
...
+
+The QA crawl will skip over any non-HTML pages.
```
Shrinks99, line 36: Would expand a bit on why this is?  Feel free to edit if anything I've written is inaccurate!

```suggestion
The analysis crawl will skip over any non-HTML pages such as PDFs which can be relied upon to be bit-for-bit identical as long as the resource was fully fetched.
```

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Comparison Dimensions
+
+### Screenshot Match
+
+One way to compare crawl and replay is to compare the screenshots of a page while it is being crawled with when it is being replayed. To make this simple, the initial viewport screenshots of each page from the crawl and replay are compared on the basis of matching pixel count. This results in a score between 0 and 1.0 representing the percentage match between the crawl and replay screenshots for each page. The screenshots are stored in `urn:view:<url>` WARC resource records.
```
Shrinks99, line 42: Pretty sure this is a little more complex than the amount of pixels being the same!  Changed to pixel value similarity, referencing the RGB values.

```suggestion
One way to compare crawl and replay is to compare the screenshots of a page while it is being crawled with when it is being replayed. The initial viewport screenshots of each page from the crawl and replay are compared on the basis of pixel value similarity. This results in a score between 0 and 1.0 representing the percentage match between the crawl and replay screenshots for each page. The screenshots are stored in `urn:view:<url>` WARC resource records.
```

docs/docs/user-guide/qa.md
```diff
@@ -0,0 +1,79 @@
...
+
+To enable comparison on this dimension, the crawl must be run with at least the `--screenshot view` option. (Additional screenshot options can be added as well).
+
+### Text Match
+
+Another way to compare the crawl and replay results is to compare the extracted text that appeears on the page, extracted from the HTML. This is currently done by comparing the extracted text from crawl and replay on the basis of [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance). This results in a score between 0 and 1.0 representing the percentage match between the crawl and replay text for each page. The extracted text is stored in `urn:text:<url>` WARC resource records.
```
Shrinks99, line 48: Nice wiki link!

```suggestion
Another way to compare the crawl and replay results is to use the extracted text from the HTML. This is currently done by comparing the extracted text from crawl and replay on the basis of [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance). This results in a score between 0 and 1.0 representing the percentage match between the crawl and replay text for each page. The extracted text is stored in `urn:text:<url>` WARC resource records.
```

Shrinks99 (changes requested):
Nice!!

docs/docs/user-guide/qa.md
tw4l: Yeah, probably don't wanna go crazy with this but it's also worth flagging since this is a big feature that doesn't exist in older releases I think.

tw4l: You're right that better docs versioning would definitely help!

Shrinks99 (approved):
Looks great!  Thanks for putting this together!

PR #555, opened 2024-04-18
Support site-specific wait via browsertrix-behaviors

ikreymer:
The 0.6.0 release of Browsertrix Behaviors / webrecorder/browsertrix-behaviors#70 introduces support for site-specific behaviors to implement an `awaitPageLoad()` function which allows for waiting for specific resources on the page load.

This PR just adds a call to this function directly after page load. This is to support custom loading for Instagram, and other sites in the future.

Testing:
- Build browsertrix-behaviors from branch / use `dist/behaviors.js` from PR
- Copy to behaviors.js
- Uncomment the COPY behaviors line in Dockerfile
- Build
- Run crawl with an instagram page.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1835,6 +1835,17 @@ self.__bx_behaviors.selectMainBehavior();
 
     await this.netIdle(page, logDetails);
 
+    logger.debug(
+      "Waiting for custom page load via behavior",
+      logDetails,
+      "behavior",
+    );
+    try {
+      await page.evaluate("self.__bx_behaviors.initialPageLoad();");
+    } catch (e) {
+      logger.warn("Waiting for custom page load failed", e, "behavior");
+    }
+
     if (this.params.postLoadDelay) {
       logger.info("Awaiting post load delay", {
         seconds: this.params.postLoadDelay,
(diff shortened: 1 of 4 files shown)


tw4l:
Makes sense, nice to be able to support it per-page. Guess there's a bit of overlap with `--postLoadDelay`, but I think that's okay.

src/crawler.ts
```diff
@@ -1835,6 +1835,17 @@ self.__bx_behaviors.selectMainBehavior();
 
     await this.netIdle(page, logDetails);
 
+    logger.debug(
+      "Waiting for custom page load via behavior",
+      logDetails,
+      "behavior",
+    );
+    try {
+      await page.evaluate("self.__bx_behaviors.initialPageLoad();");
```
tw4l, line 1844: Guess this already has its own timeout, yeah?

tw4l (approved):

src/crawler.ts
ikreymer: Yes, it does, so wanted to avoid yet another timeout here, though could add. Of course, its also falls under the total page timeout time.

tw4l: Yup I think with the overall page timeout and the behavior timeout we're already well covered here :)

PR #556, opened 2024-04-19
Always add warcinfo records to all WARCs

### LINKED ISSUE #553: `.warc` output, always write the warcinfo record ?
I noticed that the warcinfo record is not written in the files that that are in `collections/*/archive/*.warc.gz`. It only gets written in the combined archive written when `--combineWARC` is specified.

Is there any reason not to write it in every `collection/*/archive/*.warc.gz` file ? That way those can simply be combined at will by `cat`ing them (or not if we want to keep the files smaller). Rather than having to rely on the `--combineWARC` option. If I [understand well](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.1/#warcinfo) there's no issue with multiple `warcinfo`s, it simply applies to all subsequent records.

Also not sure it matters much but I noticed that the record is written with version `1.0` while the rest is written with version 1.1 (and I'm not sure that `format:` field matters much, it's written in all the records :-).

```
WARC/1.0
WARC-Filename: example.warc.gz
WARC-Date: 2024-04-18T21:49:24Z
WARC-Type: warcinfo
WARC-Record-ID: <urn:uuid:26769025-1e40-4b40-8ba9-99214b637fe4>
Content-Type: application/warc-fields
Content-Length: 90

software: Browsertrix-Crawler 1.0.4 (with warcio.js 2.2.1)
format: WARC File Format 1.0


WARC/1.1
WARC-Page-ID: 1e929081-aeaf-4814-bb55-54ef27b1a27c
WARC-Resource-Type: document
```

ikreymer:
Fixes #553 

Includes `warcinfo` records at the beginning of new WARCs, as well as the combined WARC.
Makes the warcinfo record also WARC/1.1 to match the rest of the WARC records.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -16,8 +16,6 @@ import { parseArgs } from "./util/argParser.js";
 
 import yaml from "js-yaml";
 
-import * as warcio from "warcio";
-
 import { HealthChecker } from "./util/healthcheck.js";
 import { TextExtractViaSnapshot } from "./util/textextract.js";
 import {
@@ -60,7 +58,7 @@ import { CDPSession, Frame, HTTPRequest, Page, Protocol } from "puppeteer-core";
 import { Recorder } from "./util/recorder.js";
 import { SitemapReader } from "./util/sitemapper.js";
 import { ScopedSeed } from "./util/seeds.js";
-import { WARCWriter } from "./util/warcwriter.js";
+import { WARCWriter, createWARCInfo, setWARCInfo } from "./util/warcwriter.js";
 
 const HTTPS_AGENT = new HTTPSAgent({
   rejectUnauthorized: false,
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Very nice! Tested and working well

ikreymer:
As discussed, one thing we may want to add is additional data to the warcinfo records, perhaps:

screenshots: `type: screenshot`
text: `type: text`
info (for QA crawls): `type: pageinfo`
regular warcs + combined warc: `type: combined` or `type: web`?

tw4l:
> regular warcs + combined warc: `type: combined` or `type: web`?

+1 for `web` as best term we've come up with so far for general WARC records capturing web traffic

ikreymer:
Will create separate issue / PR for additional type field

PR #562, opened 2024-05-06
improved handling of requests from workers:

ikreymer:
on sites with regular workers, requests from workers were being skipped as there was no match for the worker frameId
- add recorder.hasFrame() frameId to match not just service-worker frameIds but also other frame ids already tracked in the in frameIdToExecId map

Tests:
Fixes 'Skipping URL from unknown frame' on sites such as https://www.stadttunnel-zug-nein.ch/
Without this fix, many URLs result in this error, and this should eliminate that.

### DIFF
--- a/src/util/browser.ts
+++ b/src/util/browser.ts
@@ -504,14 +504,14 @@ export class Browser {
           recorder.swFrameIds.add(frameId);
         }
 
-        if (recorder.swFrameIds && recorder.swFrameIds.has(frameId)) {
+        if (recorder.hasFrame(frameId)) {
           foundRecorder = recorder;
           break;
         }
       }
 
       if (!foundRecorder) {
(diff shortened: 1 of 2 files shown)


src/util/browser.ts
```diff
@@ -504,14 +504,14 @@ export class Browser {
...
 
-        if (recorder.swFrameIds && recorder.swFrameIds.has(frameId)) {
+        if (recorder.hasFrame(frameId)) {
           foundRecorder = recorder;
           break;
         }
       }
 
       if (!foundRecorder) {
-        logger.debug(
```
ikreymer, line 514: not sure if this should be a warning.. means something is getting skipped, but maybe debug is ok for that?

tw4l (approved):
Tested and working, thanks very much!

PR #564, opened 2024-05-07
Skip Checking Empty Frame + eval timeout

ikreymer:
Don't run frame.evaluate() on an empty frame, also add a timeout just in case to frame.evaluate().

Testing:
- Fixes issue on crawling https://memex.garden/ where crawler frame.evaluate() on an empty frame doesn't return!


tw4l (approved):
Nice! Thanks for the good test case example.

PR #565, opened 2024-05-07
add STORE_REGION env var to be able to specify region

### LINKED ISSUE #515: S3 upload and region error

ikreymer:
defaults to us-east-1 for minio compatibility
fixes #515


tw4l (approved):

tw4l:
Might be worth adding to the documentation? Otherwise looks good!

PR #566, opened 2024-05-07
Adding support for HTTP Basic Auth

### LINKED ISSUE #168: HTTP Basic Auth profile?
Is it possible to crawl a site that has had HTTP Basic Authentication turned on? When I try to create a profile I get an error:

```
docker run -p 9222:9222 -p 9223:9223 -v $PWD/profiles:/output/ -it webrecorder/browsertrix-crawler create-login-profile --interactive --url "http://www.buffon.cnrs.fr"
Launching XVFB
loading
/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Frame.js:225
                    ? new Error(`${response.errorText} at ${url}`)
                      ^

Error: net::ERR_INVALID_AUTH_CREDENTIALS at http://www.buffon.cnrs.fr
    at navigate (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Frame.js:225:23)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Frame.goto (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Frame.js:195:21)
    at async Page.goto (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Page.js:1155:16)
    at async main (/app/create-login-profile.js:161:3)
```

Then I tried to create a profile by starting at one place and then navigating to the desired location. I see in the devtools console that a HTTP 401 response has been sent, but the popup to log in does not appear.

Screen Shot 2022-09-22 at 11 35 27 AM

edsu:
You can use the `--httpBasicAuth` flag to pass colon separated username and password credentials to the browser used for crawling.

For some reason I couldn't get `page.authenticate()` to work as advertised in the puppeteer docs. I could get it to work outside of browsertrix-crawler's browser, so maybe there's some interference from another setting? But setting the Authenticate header directly worked nicely. Thanks for the tip @vnznznz!

Fixes #168

### DIFF
--- a/tests/http_basicauth.test.js
+++ b/tests/http_basicauth.test.js
@@ -1,0 +1,15 @@
+import child_process from "child_process";
+import fs from "fs";
+
+test("test that http basic auth works", async () => {
+  child_process.execSync(
+    'docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://httpbin.org/basic-auth/foo/bar --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json',
+  );
+});
+
+test("check that the page didn't fail", () => {
+  const data = fs.readFileSync("test-crawls/stats.json", "utf8");
+  const dataJSON = JSON.parse(data);
+  expect(dataJSON.crawled).toEqual(1);
+  expect(dataJSON.failed).toEqual(0);
+});
--- a/tests/http_basicauth.test.js
+++ b/tests/http_basicauth.test.js
@@ -1,15 +1,29 @@
 import child_process from "child_process";
 import fs from "fs";
+import http from 'http';
 
-test("test that http basic auth works", async () => {
-  child_process.execSync(
-    'docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://httpbin.org/basic-auth/foo/bar --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json',
-  );
+const server = http.createServer((req, res) => {
+  res.writeHead(200);
+  res.write("Hi!");
+  res.end();
 });
 
-test("check that the page didn't fail", () => {
+beforeAll(() => {
+  server.listen(9998, '0.0.0.0');
+});
+
+test("test that http basic auth works", async () => {
+
+  child_process.execSync(
+    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://host.docker.internal:9998/ --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json",
+  );
+
   const data = fs.readFileSync("test-crawls/stats.json", "utf8");
   const dataJSON = JSON.parse(data);
   expect(dataJSON.crawled).toEqual(1);
   expect(dataJSON.failed).toEqual(0);
 });
+
+afterAll(() => {
+  server.close();
+})
--- a/tests/http_basicauth.test.js
+++ b/tests/http_basicauth.test.js
@@ -15,7 +15,7 @@ beforeAll(() => {
 test("test that http basic auth works", async () => {
 
   child_process.execSync(
-    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://host.docker.internal:9998/ --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json",
+    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://host.docker.internal:9998 --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json",
   );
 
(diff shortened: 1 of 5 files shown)


ikreymer:
Hm, it's probably because we are also overriding Fetch.requestPaused, but not [Fetch.authRequired](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#event-authRequired) for recording, though they're two separate events there.
The downside of this approach is that it sends the auth header always with any request. Is that a security risk potentially?
I think that will probably send it to third party URLs also.

ikreymer:
I wonder if the auth should be associated with a particular seed, and then only sent to requests to the seed domain. That'd be a bit more work, though, and slight refactoring of seeds... I suppose you would probably never have two seeds with two different HTTP auth passwords?

edsu:
I was wondering about that too. I guess it begs the question of how it would show up in the browsertrix user interface?

I don't think it would be easy to have url specific auth from the command line? But certainly from YAML configuration it would work?

```yaml
seeds:
  - url: https://webrecorder.net/
    depth: 1
    scopeType: "prefix"
    httpBasicAuth: alice:abc123
```

tests/http_basicauth.test.js
```diff
@@ -0,0 +1,15 @@
+import child_process from "child_process";
+import fs from "fs";
+
+test("test that http basic auth works", async () => {
+  child_process.execSync(
+    'docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://httpbin.org/basic-auth/foo/bar --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json',
```
vnznznz, line 6: httpbin.org doesn't have a responding maintainer at the moment. I think it's better for test stability to boot https://github.com/psf/httpbin with docker instead of relying on external hosting.

tw4l: Agreed but might be made complicated by https://github.com/psf/httpbin/issues/27

edsu: Yeah, I noticed the Docker image wasn't currently working.

I was going to suggest https://github.com/mccutchen/go-httpbin (I'm not sure if it does http auth). But I wasn't entirely sure how to slot in running the container as part of the current test suite.

vnznznz: Couldn't find support for http auth in go-httpbin. Another option would be to boot a webserver inside the test suite like here: https://github.com/webrecorder/browsertrix-crawler/blob/main/src/util/replayserver.ts#L28-L64

edsu: I pushed up a revised test that starts a web server and then ensures browsertrix-crawler can access it. Unfortunately it doesn't seem to work as written (even when I'm not yet checking username/password). 

The crawler gets reports:

```
{"timestamp":"2024-05-16T17:50:10.725Z","logLevel":"error","context":"general","message":"Page Load Timeout, skipping page","details":{"msg":"net::ERR_CONNECTION_REFUSED at http://host.docker.internal:9998/","page":"http://host.docker.internal:9998/","workerid":0}}
```

But I can see when running the server and the docker container independently (independent of the test) that things work ok with `http://host.docker.internal:9998`. So I'm a bit stumped what I'm doing wrong here.

tw4l: @edsu I think beforeAll is expecting a promise to be returned, so might be a timing/async issue as you suspect. Might be easier to just start the server in-line prior to running the crawler and close after it's done, all within the test?

tests/http_basicauth.test.js
```diff
@@ -0,0 +1,29 @@
...
+});
+
+beforeAll(() => {
+  server.listen(9998, '0.0.0.0');
+});
+
+test("test that http basic auth works", async () => {
+
+  child_process.execSync(
+    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://host.docker.internal:9998/ --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json",
```
tw4l, line 18: Try without the trailing slash after the port?

```suggestion
    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url http://host.docker.internal:9998 --httpBasicAuth foo:bar --collection basicauth-test --statsFilename stats.json",
```

ikreymer:
@edsu Can you try #616, I believe this should do what you want, while keeping the http-auth per seed, which allows for it to be specified directly as `https://user:pass@example.com/` (and should work with Browsertrix app as is)

tests/http_basicauth.test.js
ikreymer: FWIW, I believe it has to be the docker IP 172.17.0.1 while host.docker.internal is more for mac/windows.

ikreymer:
Closing in favor of #616 - I suppose could add CLI flags as well to that PR, but that introduces more risk of leaking credentials with multiple seeds, and requires more work to integrate into Browsertrix app and so is less flexible

edsu:
Thanks for pushing this forward. It makes sense why you would want to have the auth attached to the seed.

PR #571, opened 2024-05-14
PDF loading status code fix

### LINKED ISSUE #570: WARC record HTTP status code is 0 instead of 200
On some occasion, we have WARC archives where the HTTP status code is 0 instead of 200.

When looking inside the WARC, we see that we indeed have a HTTP Header whose value is `HTTP/1.1 0 OK`

Sample command to quickly reproduce the problem (only 53 pages are fetched):

```
docker run -v $PWD/output:/output --name zimit2 --rm  webrecorder/browsertrix-crawler:1.1.1 crawl --failOnFailedSeed --behaviors "autoplay,autofetch,autoscroll" --url "https://journals.openedition.org/bibnum/889" --scopeType host --mobileDevice "Pixel 2" --cwd /output --combineWARC --depth 1
```

Details about problematic WARC record:

```
### REC Headers ###
WARC/1.1
WARC-Page-ID: 5910275c-800e-4687-b4f1-04bcc965406a
WARC-Resource-Type: document
WARC-JSON-Metadata: {"ipType":"Public","cert":{"issuer":"R3","ctc":"0"}}
WARC-Target-URI: https://journals.openedition.org/bibnum/pdf/889
WARC-Date: 2024-05-14T14:39:43.291Z
WARC-Type: response
WARC-Record-ID: <urn:uuid:c484334e-3c77-43dc-b49d-a70df8f72ddf>
Content-Type: application/http; msgtype=response
WARC-Payload-Digest: sha256:3318b4cb23862b565cf745ba42415824d6bc477b0645525df764f5e580e28ca7
WARC-Block-Digest: sha256:c30ff1dd4f4c7063abadef817a65a1ce7ed5d3daf389bd44ffacf59e71c50d0d
Content-Length: 620680

### HTTP Headers ###
HTTP/1.1 0 OK
Accept-Ranges: bytes
Age: 0
Connection: keep-alive
Content-Length: 620314
Content-Type: application/pdf
Content-disposition: attachment; filename="bibnum-889.pdf"
Content-transfer-encoding: binary
Date: Tue, 14 May 2024 14:39:43 GMT
Via: 1.1 varnish (Varnish/6.5)
X-Backend: journals3
X-Served-By: frontwebjournals1
X-Varnish: 687682549
```

Doing a curl on the same URL `https://journals.openedition.org/bibnum/pdf/889` does not gives a 0 status code. 

At first, I thought it was linked to the fact that it is a direct fetch, but this is not the case in fact, it is considered as a real page. Maybe a problem linked to the fact that it is a PDF and hence rendered differently by the browser (usually downloaded instead of being displayed)?

Is this normal / expected behavior or a bug?

ikreymer:
when loading a PDF as a page, the browser returns a 'false positive' net::ERR_ABORTED even though the PDF is loaded.
- this is already handled, but status code was still being cleared, ensure status code is not reset to 0 on response
- ensure page status and mime are also recorded if this failure is ignored (in shouldIgnoreAbort)

fixes #570

Note: as follow up, should ensure PDFs are directly fetched, even if cookies are being set

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -391,10 +391,8 @@ export class Recorder {
         // check if this is a false positive -- a valid download that's already been fetched
         // the abort is just for page, but download will succeed
         if (type === "Document" && reqresp.isValidBinary()) {
-          this.serializeToWARC(reqresp);
-          this.addPageRecord(reqresp);
           this.removeReqResp(requestId);
-          return;
+          return this.serializeToWARC(reqresp);
         } else if (
           url &&
           reqresp.requestHeaders &&
@@ -426,8 +424,8 @@ export class Recorder {
     }
     reqresp.status = 0;
     reqresp.errorText = errorText;
-
     this.addPageRecord(reqresp);
+
     this.removeReqResp(requestId);
   }
 
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -392,7 +392,9 @@ export class Recorder {
         // the abort is just for page, but download will succeed
         if (type === "Document" && reqresp.isValidBinary()) {
           this.serializeToWARC(reqresp);
-          //} else if (url) {
+          this.addPageRecord(reqresp);
+          this.removeReqResp(requestId);
+          return;
         } else if (
           url &&
           reqresp.requestHeaders &&
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1697,7 +1697,7 @@ self.__bx_behaviors.selectMainBehavior();
     // Detect if ERR_ABORTED is actually caused by trying to load a non-page (eg. downloadable PDF),
     // if so, don't report as an error
     page.once("requestfailed", (req: HTTPRequest) => {
-      ignoreAbort = shouldIgnoreAbort(req);
+      ignoreAbort = shouldIgnoreAbort(req, data);
     });
 
     let isHTMLPage = data.isHTMLPage;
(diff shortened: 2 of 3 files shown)


src/util/recorder.ts
```diff
@@ -392,7 +392,9 @@ export class Recorder {
         // the abort is just for page, but download will succeed
         if (type === "Document" && reqresp.isValidBinary()) {
           this.serializeToWARC(reqresp);
-          //} else if (url) {
+          this.addPageRecord(reqresp);
+          this.removeReqResp(requestId);
+          return;
```
tw4l, line 397: Now that we have this return here, the else if statement below could be moved out of the conditional and the now unnecessary `break` removed.

tw4l (approved):
Looks good! Would be nice to add a test case for this.

src/util/recorder.ts
ikreymer: Hmm, there are other conditions around this, don't think that quite works? Removing the `break` results in lint error.

ikreymer: Did change it to:
```
          this.removeReqResp(requestId);
          return this.serializeToWARC(reqresp);
```
which is consistent with other usage (and the addPageRecord is already part of serializeToWARC) and that way the promise from serializeToWARC can be awaited.

tw4l: cool, that works!

tw4l:
Very nice! Thanks for the tests! :)

Issue #463, opened 2024-02-08
Brave Default Settings


Shrinks99:
### Context

Some of Brave's settings _aren't_ good defaults for crawling.  

- Embedded LinkedIn posts are blocked by default to preserve privacy — probably with good intention, but it also means they won't be captured by the crawler!
- AMP pages are redirected automatically — _unsure_ if this is good or bad, seems like possible added hassle when figuring out scoping?
-  Tracker and ad blocking is set to `Standard`, unsure what this does
- Cross site cookies are blocked  — likely fine?
- Widevine is off by default — likely not an issue, crawling DRM content doesn't work.
- IPFS is set to "Ask" and requires user input, won't work when crawling
- Brave's analytics are on, would probably be good to turn these off?  We're not real users :P

### User Story

As a user, I'd like my crawler's default settings to offer the best chances of success!

tw4l:
Thank you for doing this investigation! I agree that it'd be wise to tweak the default profile to fit user expectations a bit better.

We could also potentially have a selection of "base profiles" that users could extend from or ignore as they wish, e.g.:

1. All Shields Disabled
2. Webrecorder default
3. Aggressive Ad + Cookie Blocking

PR #579, opened 2024-05-21
Add group policies, limit browser access to container filesystem

vnznznz:
Add some default policy settings to disable unneeded Brave features. Helps a bit with #463, but Brave unfortunately doesn't provide all mentioned settings as policy options.

Most important changes are in `config/policies/lockdown-profilebrowser.json` it limits access to the container filesystem especially during interactive profile browser creation.

### DIFF
--- a/config/policies/lockdown-profilebrowser.json
+++ b/config/policies/lockdown-profilebrowser.json
@@ -1,0 +1,8 @@
+{
+    "IncognitoModeAvailability": 1,
+    "TorDisabled": true,
+    "AllowFileSelectionDialogs": false,
+    "URLBlocklist": [
+        "file://*"
+    ]
+}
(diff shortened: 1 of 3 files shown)


config/policies/lockdown-profilebrowser.json
```diff
@@ -0,0 +1,8 @@
+{
+    "IncognitoModeAvailability": 1,
+    "TorDisabled": true,
```
ikreymer, line 3: Only question about this one, maybe we want to leave option to use Tor for future use? I guess could also change.

ikreymer:
Looks good!
Is there also a policy to enable kiosk mode, or is that not available here?

vnznznz:
> Looks good! Is there also a policy to enable kiosk mode, or is that not available here?

There's no single policy for kiosk mode. Kiosk mode also often means hiding the URL bar, which I'd also avoid for now.

PR #583, opened 2024-05-24
Load non-HTML resources directly whenever possible

ikreymer:
Optimize the direct loading of non-HTML pages. Currently, the behavior is:
- make a HEAD request first
- make a direct fetch request only if HEAD request is a non-HTML and 200
- only use fetch request if non-HTML and 200 and doesn't set any cookies

This changes the behavior to:
- get cookies from browser for page URL
- make a direct fetch request with cookies, if provided
- only use fetch request if non-HTML and 200
Also:
- ensures pageinfo is properly set with timestamp for direct fetch.
- remove obsolete Agent handling that is no longer used in default (fetch)

If fetch request results in HTML, the response is aborted and browser loading is used.

Note: initially attempted to handle redirects, but gets a bit complicated since many need to follow the redirect check to determine if request is non-HTML, and would need to buffer those responses, and not write them in case they need to be dropped in favor of browser loading. Erring on side of caution of requiring 4xx/5xx responses to still be loaded through browser, just in case.

Other optimization: it is possible to turn on http/2 loading for fetch, as mentioned in: https://github.com/nodejs/undici/issues/2750#issuecomment-1941009554 (the obsolete Agent setup from http/https modules has been removed)

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -44,27 +44,18 @@ import { Browser } from "./util/browser.js";
 import {
   ADD_LINK_FUNC,
   BEHAVIOR_LOG_FUNC,
-  HTML_TYPES,
   DEFAULT_SELECTORS,
 } from "./util/constants.js";
 
 import { AdBlockRules, BlockRules } from "./util/blockrules.js";
 import { OriginOverride } from "./util/originoverride.js";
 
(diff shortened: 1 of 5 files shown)


tw4l (approved):
Very nice! Tested well, and the reorganization is a nice touch.

PR #586, opened 2024-05-28
Consider disk usage of collDir instead of default /crawls

### LINKED ISSUE #585: Crawler is not checking the proper disk usage
https://github.com/webrecorder/browsertrix-crawler/blob/a7d279cfbd3f6ca8375e564d1e0ef751af7e345f/src/util/storage.ts#L212

When checking disk usage, the crawler always checks the disk usage on `/crawls` directory (with `df /crawls`) while it should in fact take into account the `--cwd` parameter.

I suggest to even take into account only the `collDir` property (which is a subfolder or `cwd`) since this is the real place where stuff will be stored and someone might tweak its mounting points to mount a dedicated partition only for `collDir`.

I will propose a PR right now.

benoit74:
Fix #585 

Changes:
- compute disk usage based on crawler `collDir` property instead of always computing it on `/crawls` directory

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1114,7 +1114,11 @@ self.__bx_behaviors.selectMainBehavior();
 
     if (this.params.diskUtilization) {
       // Check that disk usage isn't already or soon to be above threshold
-      const diskUtil = await checkDiskUtilization(this.params, size);
+      const diskUtil = await checkDiskUtilization(
+        this.collDir,
+        this.params,
+        size,
+      );
       if (diskUtil.stop === true) {
         interrupt = true;
       }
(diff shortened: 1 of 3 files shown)


benoit74:
@tw4l @ikreymer this is ready for review from my PoV, I don't get why there are three failing tests which seems very unrelated from my experience (but still very limited compared to yours)

tw4l:
Thanks for this @benoit74! Looks like the tests are okay now, maybe that was just a fluke/bad run. Will plan to review this week.

ikreymer (approved):
thanks, yep, i think that makes sense! I think we set it to /crawls in case multiple crawls are running using same volume locally, but that should still work fine and allows for more granular checking

PR #589, opened 2024-05-30
proxy: support setting proxy via --proxyServer, PROXY_SERVER env var or PROXY_HOST + PROXY_PORT env vars

### LINKED ISSUE #587: Add SOCKS5 + HTTP Proxy support (and add backwards compatible support for 0.12.x support in Browsertrix Crawler 1.x)
We've had a user report that specifying browser proxy information via the `PROXY_HOST` and `PROXY_PASS` environment variables is not working in the Browsertrix 1.x releases.

We should fix this and add tests to catch any future regressions.

Also adding SOCKS5 support as well as HTTP proxy support, set via the above env vars.

ikreymer:
fixes #587 

The proxy env vars PROXY_HOST and PROXY_PORT were being ignored, as they were hardcoded to obsolete values in the Dockerfile.

Proxy settings can now be set, in order of precedence via:
- --proxyServer cli flag
- PROXY_SERVER env var
- PROXY_HOST and PROXY_PORT env vars, which set an HTTP proxy server only (for backwards compatibility with 0.12.x)

The --proxyServer / PROXY_SERVER settings are passed to the browser via the --proxy-server flag.
AsyncFetcher / direct fetch also supports HTTP and SOCKS5 proxying.
Supported proxies are: HTTP no auth, SOCKS5 no auth, SOCKS5 with auth (supported in Brave, but not Chrome!)


tw4l:
Might involve a bit of effort to spin up a local proxy server, but would be great to add some tests with this PR to ensure the feature is working correctly moving forward.

vnznznz:
> Might involve a bit of effort to spin up a local proxy server, but would be great to add some tests with this PR to ensure the feature is working correctly moving forward.

The socks proxy feature for browsertrix will also depend on this

vnznznz:
The AsyncFetcher in recorder.ts doesn't handle proxy settings as well. As nodejs doesn't seem to support setting a global proxy, it might make sense to just add support for socks proxy like we had with pywb?

ikreymer:
> The AsyncFetcher in recorder.ts doesn't handle proxy settings as well. As nodejs doesn't seem to support setting a global proxy, it might make sense to just add support for socks proxy like we had with pywb?

Now fixed, supporting http and socks proxy settings for both browser and direct fetch via async fetcher.
The SOCKS5 proxy also supports auth, since it is supported in Brave (though not Chrome!) with `--proxy-server` flag.
The HTTP proxy does not support auth, as that's not supported in `--proxy-server` flag (always results in an interactive prompt)

ikreymer:
> Might involve a bit of effort to spin up a local proxy server, but would be great to add some tests with this PR to ensure the feature is working correctly moving forward.

Tests added in `proxy.test.js` using https://hub.docker.com/r/tarampampam/3proxy/ which supports both HTTP and SOCKS5 proxying!

vnznznz:
Got one of these errors when crawling via a socks proxy
```
TypeError [ERR_INVALID_ARG_TYPE]: The "readableStream" argument must be an instance of ReadableStream. Received an instance of PassThrough
    at Object.newStreamReadableFromReadableStream (node:internal/webstreams/adapters:482:11)
    at Readable.fromWeb (node:internal/streams/readable:1740:27)
    at AsyncFetcher._doFetch (file:///app/dist/util/recorder.js:918:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async AsyncFetcher.load (file:///app/dist/util/recorder.js:800:26)
    at async Recorder.directFetchCapture (file:///app/dist/util/recorder.js:741:21)
    at async Crawler.crawlPage (file:///app/dist/crawler.js:512:47)
    at async PageWorker.crawlPage (file:///app/dist/util/worker.js:153:21)
    at async PageWorker.timedCrawlPage (file:///app/dist/util/worker.js:169:13)
    at async PageWorker.runLoop (file:///app/dist/util/worker.js:224:17)
 ```

File: src/util/proxy.ts [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+
+export function initProxy(proxy?: string): string {
+  if (!proxy) {
+    proxy = getEnvProxyUrl();
+  }
+  if (proxy) {
+    const dispatcher = createDispatcher(proxy);
+    if (dispatcher) {
+      setGlobalDispatcher(dispatcher);
+      return proxy;
```
tw4l, line 27: Might be worth also logging the proxy being used here as an `info` message so it's clear from the logs that the traffic was proxied?

tw4l (approved):
Code looks good, just one minor logging suggestion.

Tested locally with 3proxy and with a remote paid socks5 proxy (as part of a service I'm already using), and worked as expected with both. Nice work! Really appreciate the thorough test coverage too.

src/util/proxy.ts
ikreymer: There is a log line here that's logged once:
https://github.com/webrecorder/browsertrix-crawler/pull/589/files#diff-44d0707556db333e2610b1c8d5a2a6da872141d6ea9a52ac978271ef83841df3R239
that's basically the same thing, just in browser.ts

tw4l: Ah great, missed that!

PR #590, opened 2024-05-30
Get rid of Python dependencies now that pywb is gone

benoit74:
As far as I've understood, pywb is not used anymore now.

We however still have few Python dependencies which AFAIK are not used by the crawler anymore. I propose to get rid of them since this will make life easier to migrate to Ubuntu Noble


benoit74:
This is ready for review

benoit74:
Hum, looks like Python is still used for cdxj indexer ... so this will have to be fixed indeed. Not ready for review then, sorry.

benoit74:
Closing this in favor of https://github.com/webrecorder/browsertrix-crawler/pull/591

tw4l:
> Hum, looks like Python is still used for cdxj indexer ... so this will have to be fixed indeed. Not ready for review then, sorry.

Yes at this point it's still used to create WACZ files. That should change soon with a switch to js-wacz however :) Thanks for the PRs!

PR #591, opened 2024-05-30
Use isolated Python venv for dependencies installation

benoit74:
Debian distro now requires the use of virtual environments to not mess with dependencies installed by official apt packages

This commit also removes tldextract update now that pywb is not in use anymore.

These changes are required by https://github.com/webrecorder/browsertrix-browser-base/pull/20

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,12 @@ EXPOSE 9222 9223 6080
 WORKDIR /app
 
 ADD requirements.txt /app/
-RUN pip install -U setuptools; pip install -r requirements.txt
+RUN apt update && apt -y upgrade && apt -y install python3-venv
+RUN python3 -m venv /app/python-venv && \
+    /app/python-venv/bin/pip install -U setuptools && \
+    /app/python-venv/bin/pip install -r requirements.txt && \
+    ln -s /app/python-venv/bin/wacz /usr/bin/wacz && \
+    ln -s /app/python-venv/bin/cdxj-indexer /usr/bin/cdxj-indexer
 
 ADD package.json yarn.lock /app/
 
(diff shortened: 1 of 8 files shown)


benoit74:
Nota: build is failing for now, but it should pass (at least it does on my local environment) once https://github.com/webrecorder/browsertrix-crawler/pull/591 is merged / has published new images

tw4l:
> Nota: build is failing for now, but it should pass (at least it does on my local environment) once #591 is merged / has published new images

The CI is building the image locally and I'm seeing the same build failure locally, but I think it's just that we need to install `python3-venv`, which will require updating and upgrading apt as well. I'll add a suggestion for the line to add, with that it's building well for me locally.

Dockerfile
```diff
@@ -21,16 +21,17 @@ ENV PROXY_HOST=localhost \
 WORKDIR /app
 
 ADD requirements.txt /app/
-RUN pip install -U setuptools; pip install -r requirements.txt
+RUN python3 -m venv /app/python-venv && \
```
tw4l, line 24:
```suggestion
RUN apt update && apt -y upgrade && apt -y install python3-venv
RUN python3 -m venv /app/python-venv && \
```

tw4l:
> Nota: build is failing for now, but it should pass (at least it does on my local environment) once #591 is merged / has published new images

Ah, my mistake, now I see you meant to point to https://github.com/webrecorder/browsertrix-browser-base/pull/20! Thanks for that. Can revert the extra commit here when that's merged :)

benoit74:
> Ah, my mistake, now I see you meant to point to https://github.com/webrecorder/browsertrix-browser-base/pull/20! Thanks for that. Can revert the extra commit here when that's merged :)

No worries, but yeah, I thought it would make more sense to install python3-venv in base image rather than doing again an apt-update + install + cleanup in every sub-image.

PR #592, opened 2024-05-30
base image version bump to brave 1.66.115

ikreymer:


tw4l:
Only question is if we want to use this as an opportunity to merge https://github.com/webrecorder/browsertrix-browser-base/pull/20 and https://github.com/webrecorder/browsertrix-crawler/pull/591, or to do that on a later Brave release.

ikreymer:
> Only question is if we want to use this as an opportunity to merge [webrecorder/browsertrix-browser-base#20](https://github.com/webrecorder/browsertrix-browser-base/pull/20) and #591, or to do that on a later Brave release.

I was thinking both of those can wait until we have js-wacz, and then we can drop python altogether. There's already a newer brave release, so I imagine can revisit if needed sooner?

tw4l:
> I was thinking both of those can wait until we have js-wacz, and then we can drop python altogether.

Makes sense to me!

tw4l (approved):

PR #594, opened 2024-06-06
add --dryRun flag and mode

### LINKED ISSUE #593: Dry Run Mode
A 'dry run' mode (is that the best name?) can be used to run a crawl without storing any archive data.  It can be used to examine the scope of crawl via logs / saved state, or, to delegate handling via a remote proxy, when used in conjunction with external proxies (see #587). The dry run mode should still fetch everything + run behaviors, but not write any local data.
Text extraction and screenshots should also be skipped.

ikreymer:
- if set, runs the crawl but doesn't store any archive data (WARCS, WACZ, CDXJ, pages) while logs are still written, and saved state can be generated (per the --saveState options).
- adds test to ensure only 'logs' dir is generated with --dryRun
- screenshot, text extraction are skipped altogether in dryRun mode, warning is printed that storage and archiving-related options may be ignored
- fixes #593


src/crawler.ts
```diff
@@ -2068,6 +2079,10 @@ self.__bx_behaviors.selectMainBehavior();
   async initPages(filename: string, title: string) {
     let fh = null;
 
+    if (this.params.dryRun) {
+      return;
+    }
```
tw4l, line 2084: Given how much useful information there is in the pages.jsonl files, I'm tempted to let the dry runs write these too.

tw4l:
Nice!

I left an inline comment for it but I think the pages.jsonl files might be good to write for dry runs as well. They're really useful and not that large typically.

Previous test failures fixed by rebasing on PDF test fix.

tw4l (approved):

src/crawler.ts
ikreymer: Done, added pages to dry run

PR #596, opened 2024-06-11
tests: reduce logging

ikreymer:
remove logging of crawl logs by default for clearer output from tests, only log in case of error.
Cleaner test results, eg: https://github.com/webrecorder/browsertrix-crawler/actions/runs/9671553142/job/26682374524


tw4l (approved):
Nice! Much cleaner output

PR #600, opened 2024-06-12
Change some logged errors to warns

### LINKED ISSUE #599: Make timeout logging messages warns, not errors
When page or behavior timeouts are hit the crawler currently logs these as errors, which can alarm users. In fact, this is almost expected behavior in many cases, so I would suggest we change these messages from `error` to `warn` in logging.

tw4l:
Fixes #599 

This PR modifies the following logging messages from `error` to `warn`, to avoid alarming users for fairly expected behavior:

- Behavior timeout
- Page worker timeout
- Streaming fetch error (will already trigger another page error message)

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -953,12 +953,6 @@ self.__bx_behaviors.selectMainBehavior();
         this.healthChecker.resetErrors();
       }
     } else {
-      logger.warn(
-        "Page Load Failed",
-        { loadState, ...logDetails },
-        "pageStatus",
-      );
-
       await this.crawlState.markFailed(data.url);
 
       if (this.healthChecker) {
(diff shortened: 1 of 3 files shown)


ikreymer:
Nice!

Also in case of dns failure, we have:
```
{"timestamp":"2024-06-12T18:27:46.091Z","logLevel":"error","context":"general","message":"Page Load Timeout, skipping page","details":{"msg":"net::ERR_NAME_NOT_RESOLVED at https://example.invalidx/","page":"https://example.invalidx/","workerid":0}}
{"timestamp":"2024-06-12T18:27:46.299Z","logLevel":"warn","context":"pageStatus","message":"Page Load Failed","details":{"loadState":0,"page":"https://example.invalidx/","workerid":0}}
```

if anything, maybe should be reversed, or consolidated, as it prints the reason for the error

tw4l:
Tested locally, seems good now. I think ready to merge if you agree, @ikreymer

PR #602, opened 2024-06-13
recorder: add missing shouldSkip() to responseReceived callback

### LINKED ISSUE #601: Ensure responseReceived callback also checks shouldSkip()
The responseReceived callback has a missing shouldSkip() check, resulting in a resource that should otherwise be skipped added to the fetch queue (eg. one with a chrome-extension:// URL that is skipped elsewhere).

ikreymer:
Fixes #601, fixes issue with extra wait on PDF pages, where browser seems to be waiting for a chrome-extension:// URL.
These should have already be getting skipped, but missed here.

Testing:
- Crawl a redirecting PDF, which currently will load in the browser (due to redirect): https://dpconline.org/about/governance/articles-of-association-2
- Verify no chrome-extension:// URLs are being waited on.

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -238,7 +238,7 @@ export class Recorder {
   handleResponseReceived(params: Protocol.Network.ResponseReceivedEvent) {
     const { requestId, response, type } = params;
 
-    const { mimeType, url } = response;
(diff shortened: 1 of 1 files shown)


tw4l (approved):
Working well!

PR #605, opened 2024-06-13
Fix synching extraSeeds state with multiple crawler instances

### LINKED ISSUE #604: Fix seed redirect with multiple crawl instances (eg. scale > 1)
When running with multiple crawler instances sharing state, the extra seeds added on redirect do not propagate to other crawler instances, resulting in a possible error when encountering a page from a different seed.

The seed added with addExtraSeed() needs to propagate to other crawler instances, and the isInScope() check needs to sync the local seeds array with any extra seeds from Redis.

ikreymer:
Fixes #604 

Ensures that extra seeds are propagated to all crawler instances.
Adds a new redis hashmap key to store the extraSeed mappings url->extraSeeds index, to ensure the extra seeds are added in the same order on other instances, even if encountered in different order.
Add a new redis lua primitive 'addnewseed' which combines several operations: check if extra seed already exists and returning existing index, add new seed to extraSeed list, also add to regular URL seed list.

The end result is to ensure that the extra seeds added to the end are always in the same order:

Regular Seeds (from config)   Extra Seeds (stored in Redis)
[0... N)                                      [N, N+M)

Redis stores these in a list [0, M) and the mapping url->index [0, M) is also stored in redis to ensure the same ordering.

Testing:
- Run a crawl with a redirecting site, eg. www.webrecorder.net, with scale >1. Crawler instances should not fail.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -133,6 +133,9 @@ export class Crawler {
 
   maxPageTime: number;
 
+  seeds: ScopedSeed[];
+  numOriginalSeeds = 0;
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   emulateDevice: any = {};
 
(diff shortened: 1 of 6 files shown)


tw4l (approved):
Dang, I'm impressed by how quickly you got this in! Looks good and working well in testing. Nice work!

PR #615, opened 2024-06-17
clearer scope check

ikreymer:
split isInScope into a protected sync getScope() used for link extraction (no need for async as we know seed is already set) and which returns url / isOOS count.
and a simpler, public async isInScope() which just returns a bool, but also ensures the seed exists.

### DIFF
--- a/src/util/seeds.ts
+++ b/src/util/seeds.ts
@@ -213,7 +213,12 @@ export class ScopedSeed {
     return depth >= this.maxDepth;
   }
 
-  isIncluded(url: string, depth: number, extraHops = 0, logDetails = {}) {
+  isIncluded(
+    url: string,
+    depth: number,
+    extraHops = 0,
+    logDetails = {},
+  ): { url: string; isOOS: boolean } | false {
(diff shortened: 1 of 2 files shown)


src/util/seeds.ts
```diff
@@ -213,7 +213,12 @@ export class ScopedSeed {
     return depth >= this.maxDepth;
   }
 
-  isIncluded(url: string, depth: number, extraHops = 0, logDetails = {}) {
+  isIncluded(
+    url: string,
+    depth: number,
+    extraHops = 0,
+    logDetails = {},
+  ): { url: string; isOOS: boolean } | false {
```
tw4l, line 221: This return type works but I wonder if we want to try to be more consistent?

tw4l (approved):
Approved, one question on return types that you can feel free to address or ignore at your discretion.

src/util/seeds.ts
ikreymer: well, typescript has helped make this more consistent! I feel like this works for now and is type-safe, so maybe lets leave in as is for now.

PR #616, opened 2024-06-17
http auth support per seed (supersedes #566):

ikreymer:
- parse URL username/password, store in 'auth' field in seed, remove from seed
- add Authorization header directly if auth is set
- also fix isInScope type to only return false

tests: add test for crawling with auth using http-server using local docs build

Tests:
- Test using http-server with auth serving the local docs directory, ensure crawler only succeeds if correct auth credentials are passed, otherwise fails with failOnFailedSeed

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -796,7 +796,19 @@ self.__bx_behaviors.selectMainBehavior();
     const { page, cdp, data, workerid, callbacks, directFetchCapture } = opts;
     data.callbacks = callbacks;
 
-    const { url } = data;
+    const { url, seedId } = data;
+
+    const auth = this.seeds[seedId].auth;
+
+    if (auth) {
+      logger.debug("Setting auth for seed", {
+        seedId,
+        seedUrl: this.seeds[seedId].url,
+      });
+      await page.setExtraHTTPHeaders({ Authorization: "Basic " + auth });
+    } else {
+      await page.setExtraHTTPHeaders({});
+    }
 
     const logDetails = { page: url, workerid };
     data.logDetails = logDetails;
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -798,14 +798,14 @@ self.__bx_behaviors.selectMainBehavior();
 
     const { url, seedId } = data;
 
-    const auth = this.seeds[seedId].auth;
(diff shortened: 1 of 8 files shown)


src/crawler.ts
```diff
@@ -796,7 +796,20 @@ self.__bx_behaviors.selectMainBehavior();
...
+    if (auth) {
+      logger.debug("Setting HTTP basic auth for seed", {
+        seedId,
+        seedUrl: this.seeds[seedId].url,
+      });
+      await page.setExtraHTTPHeaders({ Authorization: "Basic " + auth });
+      opts.isAuthSet = true;
+    } else if (opts.isAuthSet) {
+      await page.setExtraHTTPHeaders({});
+    }
```
tw4l, line 812: Can you talk me through what's happening here? I don't see `options.isAuthSet` being used in the worker (unless for logging?) after being set, and unclear why the need for the else if/adding empty headers.

ikreymer: oh, this is just an optimization to avoid calling `page.setExtraHTTPHeaders({})` if they were never set before. Sort of unnecessary, can juts always call it - its only being set / checked here.

tests/http-auth.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+test("run crawl with auth", () => {
+  let status = 0;
+  try {
+    execSync(`docker run --rm webrecorder/browsertrix-crawler crawl --url http://user:pass@${process.env.SERVER_HOST}:31501 --limit 2 --failOnFailedSeed`, {stdio: "inherit"});
+  } catch (e) {
+    status = e.status;
+  }
+  expect(status).toBe(0);
+});
```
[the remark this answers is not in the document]
ikreymer: yeah, that's a good point, should ensure that we actually crawled two pages. will add.

edsu:
I tested and noticed that `auth` in the config yaml didn't work the way I expected:

https://github.com/webrecorder/browsertrix-crawler/pull/618

ikreymer:
> I tested and noticed that `auth` in the config yaml didn't work the way I expected: See #618

Now fixed! Can now specify auth in the `auth` field per seed, if not in the URL itself.

PR #618, opened 2024-06-20
btoa() the auth from yaml config too

edsu:
I think this small change is needed for specifying the username and password using the `auth` property in the seed configuration?

```yaml
seeds:
  - url: http://www.example.com
    auth: "user:pass"
```

Refs #616


ikreymer:
Thanks! The `auth` flag was actually intended for copying an existing seed for seed redirects, but support can work like this as well.

PR #628, opened 2024-06-25
Don't filter saving redirect if no response body.

### LINKED ISSUE #627: Crawler is not returning full seed page URL in WARC `WARC-Target-URI`
Seed URL: https://www.playmobil.com/fr-fr/tiny-house/71509.html?gad_source=1&gclid=CjwKCAjwuJ2xBhA3EiwAMVjkVK41oNKfKsuOcp6oXd4I1lLYXhgnB4PE3Yg8zSBMPb7jHvZEZbMdBRoCizIQAvD_BwE

Command:
```
docker run  -v $PWD/output:/output --name crawlme --rm webrecorder/browsertrix-crawler:1.2.0 crawl  --failOnFailedSeed --behaviors "autoplay,autofetch,autoscroll" --url "https://www.playmobil.com/fr-fr/tiny-house/71509.html?gad_source=1&gclid=CjwKCAjwuJ2xBhA3EiwAMVjkVK41oNKfKsuOcp6oXd4I1lLYXhgnB4PE3Yg8zSBMPb7jHvZEZbMdBRoCizIQAvD_BwE" --mobileDevice "Pixel 2" --cwd /output --combineWARC
```

Only one page (the seed) is crawled by this command (a bit weird site to crawl ... but why not !).

Unfortunately, the WARC file contains no entry with a `WARC-Target-URI` matching the seed URL. Looking at `pages.jsonl`, I've found the proper `WARC-Page-ID` and I've found that the WARC record has `https://www.playmobil.com/fr-fr/tiny-house/71509.html?gad_source=1` as `WARC-Target-URI`.

This causes us an issue because it means warc2zim cannot just process the WARC file and search for the entry whose `WARC-Target-URI` is matching the seed URL. We could imagine to process `pages.jsonl` like I did manually, but this is not really convenient since it means we cannot rely only on the WARC file(s).

Is this behavior intentional or a bug / side-effect of something?

Other than this problem, the WARC is perfect and the transformation to ZIM fully OK if I hack the seed URL before calling warc2zim.

ikreymer:
It's possible for a redirect, especially a browser-generated one to have headers and no body (eg. Brave removing tracking url query). Don't filter these out from being written to WARC, just send payload to empty buffer.

fixes #627 where Brave-generated redirect response was not stored.

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -1059,13 +1059,17 @@ export class Recorder {
       );
       return;
     } else if (reqresp.shouldSkipSave()) {
-      logNetwork("Skipping writing request/response", {
-        requestId,
-        url,
-        method,
-        status,
-        payloadLength: (payload && payload.length) || 0,
-      });
+      logger.debug(
+        "Skipping writing request/response",
+        {
+          requestId,
+          url,
+          method,
+          status,
+          payloadLength: (payload && payload.length) || 0,
+        },
+        "recorder",
+      );
       return;
     }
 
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Looks good, thanks!

ikreymer:
Also added test for `gclid=` query being removed, in case that changes in the future somehow.

PR #629, opened 2024-06-25
Always download PDF + non HTML page cleanup + enterprise policy cleanup

ikreymer:
Adds enterprise policy to always download PDF and sets download dir to /dev/null
Moves policies to chromium.json and brave.json for clarity
Further cleanup of non-HTML loading path:
 - sets downloadResponse when page load is aborted but response is actually download
 - sets firstResponse when first response finishes, but page doesn't fully load
 - logs that non-HTML pages skip all post-crawl behaviors


src/crawler.ts
```diff
@@ -878,12 +885,12 @@ self.__bx_behaviors.selectMainBehavior();
     const { page, cdp, data, workerid } = opts;
     const { url } = data;
 
+    if (!data.isHTMLPage) {
```
tw4l, line 888: I think returning here means we won't wait `pageExtraDelay` after non-HTML pages

src/crawler.ts
```diff
@@ -1704,109 +1705,69 @@ self.__bx_behaviors.selectMainBehavior();
...
+      // got firstResponse and content loaded, not a failure
+      if (firstResponse && data.loadState == LoadState.CONTENT_LOADED) {
         // if timeout error, and at least got to content loaded, continue on
-        if (
-          e.name === "TimeoutError" &&
-          data.loadState == LoadState.CONTENT_LOADED
-        ) {
-          logger.warn("Page Loading Slowly, skipping behaviors", {
+        logger.warn(
+          "Page Load Timedout, loading but slowly, skipping behaviors",
```
tw4l, line 1762:
```suggestion
          "Page load timed out, loading but slowly, skipping behaviors",
```

ikreymer:
> Overall looks good! One issue with skipping post-page delay for non-HTML pages (may still want to apply the wait between requests for politeness reasons/avoiding rate limits), otherwise testing well

Thanks! Added a fix for keeping the page extra delay for non-HTML pages, which is applied at the end of everything, but *not* the post-load delay (which happens after page load but before behaviors are run) and doesn't make sense for non-HTML.

PR #635, opened 2024-07-05
Improved handling of pages that redirect back to the same page.

### LINKED ISSUE #634: Better handling of redirect chains to same page
Certain pages may create a redirect chain to a third party auth URL, eg.

https://example.com 307 -> https://auth.example.com/auth 307 -> https://example.com 200

Currently, the dupe checking will only store the first `https://example.com/` and not the subsequent 200 response.

Instead, the dupe check should include status and the page timestamp should also be updated to prefer the 200 over a 3xx response.

ikreymer:
In the case of a page https://example.com/ which results in a redirect chain: 307 https://example.com/ -> 307 https://auth.example.com/ -> 200 https://example.com/
- Includes status in dupe checks, ensures that `307 https://example.com/` and `200 https://example.com/` are both recorded to WARC
- When setting page timestamp, update the timestamp to the lower status code if above 300, eg. first setting to `307 https://example.com/` and then to `200 https://example.com/`

Fixes #634 

Testing:
- The site: https://www.journaldemontreal.com/ includes such as redirect chain which results in the 200 response not being captured, before this fix. (Note: for replay, a separate fix is needed in wabac.js in case the 307 and 200 responses are the same timestamp, to ensure the 200 one is chosen, this is fixed via: webrecorder/wabac.js#188)

### DIFF
--- a/src/replaycrawler.ts
+++ b/src/replaycrawler.ts
@@ -412,6 +412,7 @@ export class ReplayCrawler extends Crawler {
       ts: date,
       comparison: { resourceCounts: {} },
       counts: { jsErrors: 0 },
+      tsStatus: 999,
     };
     this.pageInfos.set(page, pageInfo);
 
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Nice work!

PR #636, opened 2024-07-08
Fix Pending Request causing timeout

ikreymer:
Occasionally, it seems some pending requests that only go through `Network.requestWillBeSent` but never receive a finished/failed/blocked message persist in the pending request list.

Instead of waiting for them to finish until timeout, just drop these requests. Requests that have never been intercepted (`fetch` is set) and are not being loaded asynchronously (`asyncLoading` is set) should be discarded when the page is done. Large requests that are streaming the response body, etc.. will have one of these flags set.

(Chrome devtools also handles these by marking their state to 'unknown')..

Test:
- Run a crawl of a few pages of: https://www.cbc.ca/news/canada with debug logging
- Observe a few (possibly blocked) tracking requests are stuck in "Finishing pending requests for page" without this fix.
- With this fix, they should be dropped and `Removing pending request that was never fetched` is logged.

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -323,18 +323,19 @@ export class Recorder {
   }
 
   handleRequestWillBeSent(params: Protocol.Network.RequestWillBeSentEvent) {
-    // only handling redirect here, committing last response in redirect chain
-    // request data stored from requestPaused
     const { redirectResponse, requestId, request, type } = params;
 
     const { headers, method, url } = request;
 
     logNetwork("Network.requestWillBeSent", {
       requestId,
+      url,
       redirectResponse,
       ...this.logDetails,
     });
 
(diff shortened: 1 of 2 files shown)


src/util/recorder.ts
```diff
@@ -323,18 +323,19 @@ export class Recorder {
...
   handleRequestWillBeSent(params: Protocol.Network.RequestWillBeSentEvent) {
-    // only handling redirect here, committing last response in redirect chain
-    // request data stored from requestPaused
     const { redirectResponse, requestId, request, type } = params;
 
     const { headers, method, url } = request;
 
     logNetwork("Network.requestWillBeSent", {
       requestId,
+      url,
```
tw4l, line 332: Nice addition

tw4l (approved):
Great improvement, and thanks for the URL to test against :)

PR #638, opened 2024-07-11
Loosen selectors for login fields in automated profile creation

### LINKED ISSUE #637: "Login form could not be found"
What is the proper syntax, if any, for specifying the login field names in the YAML config file?

This is the command I'm running
```
sudo docker run -v $PWD/crawls/profiles:/crawls/profiles -it webrecorder/browsertrix-crawler create-login-profile --headless --url "<URL>" --automated
```
After entering the credentials, I get
```
{"timestamp":"2024-07-09T09:40:39.787Z","logLevel":"debug","context":"browser","message":"Target Filter","details":{"url":"","type":"tab","attach":true}}
{"timestamp":"2024-07-09T09:40:39.797Z","logLevel":"debug","context":"browser","message":"Target Filter","details":{"url":"about:blank?_browsertrix5ovm89w1q7l","type":"page","attach":true}}
{"timestamp":"2024-07-09T09:40:39.826Z","logLevel":"info","context":"general","message":"Loading page: <URL>","details":{}}
{"timestamp":"2024-07-09T09:40:40.134Z","logLevel":"debug","context":"general","message":"Looking for username and password entry fields on page...","details":{}}
{"timestamp":"2024-07-09T09:41:10.147Z","logLevel":"debug","context":"general","message":"Login form could not be found","details":{}}
```

The login URL is correct, as are the credentials.

OS: Debian 12

tw4l:
Fixes #637 

- Username will match if name attribute is one of: user, username, email
- Password will match if type is password and name attribute is one of: pass, password

This loosens the rules sufficiently to solve the issue with the URL in the linked issue without requiring users to pass custom CSS selectors at this point.

It looks like we were also using XPath methods like contains whereas puppeteer expects CSS selectors, hence the syntax change.

### DIFF
--- a/src/create-login-profile.ts
+++ b/src/create-login-profile.ts
@@ -266,10 +266,10 @@ async function automatedProfile(
 
   try {
     u = await page.waitForSelector(
-      "//input[contains(@name, 'user') or contains(@name, 'email')]",
+      "input[name='user'],input[name='username'],input[name='email']",
     );
     p = await page.waitForSelector(
-      "//input[contains(@name, 'pass') and @type='password']",
+      "input[type='password'].input[name='pass'],input[name='password']",
     );
   } catch (e) {
     if (params.debugScreenshot) {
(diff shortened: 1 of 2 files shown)


ikreymer:
Makes sense! (I think the xpath syntax may not have been correct as is, was migrated from earlier versions probably).

PR #646, opened 2024-07-17
Fix 206 response + general video handling

### LINKED ISSUE #645: Fix skipping of 206 responses
Currently, 206 responses are being skipped, even full range ones. This can lead to certain videos being omitted from the crawl.
Full 206 responses should be stored, while partial 206 responses should be queued for retry.

ikreymer:
Refactors handling of 206 responses:
- If a 206 response is encountered, and its actually the full range, convert to 200 and rewrite range and content-range headers to x-range and x-orig-range. This is to support rewriting of 206 responses for DASH manifests
- If a partial 206 response starting with `0-`, do a full async fetch separately.
- If a partial 206 response not starting with 0-, just ignore (very likely a duplicate picked up when handling the 0- response)
- Don't stream content-types that can be rewritten, since streaming prevents rewriting. Fixes rewriting on DASH/HLS manifests which have no content-length and don't get properly rewritten.
- Overall, adds missing rewriting of DASH/HLS manifests that have no content-length and are served as 206.
- Update to latest wabac.js which fixes rewriting of DASH manifest to avoid duplicate '<?xml' prefix, webrecorder/wabac.js#192
- Fixes #645 

Testing:
Combined with RWP 2.1.2 (released), this fixes the capture and replay of videos on such sites as:
- https://www.rferl.org/a/weeks-best-rferl-videos-stories-27-2024/33030874.html
- https://www.bbc.com/news/videos/cy7985ly17yo (replay doesn't yet work 100% but probably not related)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ ADD config/ /app/
 
 ADD html/ /app/html/
 
-ARG RWP_VERSION=2.1.1
+ARG RWP_VERSION=2.1.2
 ADD https://cdn.jsdelivr.net/npm/replaywebpage@${RWP_VERSION}/ui.js /app/html/rwp/
 ADD https://cdn.jsdelivr.net/npm/replaywebpage@${RWP_VERSION}/sw.js /app/html/rwp/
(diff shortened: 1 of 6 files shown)


tw4l (approved):
Looks good, verified fix is working for example videos

PR #649, opened 2024-07-19
Make it clear that profile argument can be an HTTP(S) URL

benoit74:
Small documentation enhancement to make it clear that browser profile can be passed as HTTP(S) URL as well.

Nota: I simplified the doc by not specifying anymore that the tar.gz will be extracted, I think it is a detail which most user shouldn't be concerned about.


benoit74:
@ikreymer @tw4l this is ready for review.

ikreymer:
Thanks! Made some tiny tweaks to the text

Issue #486, opened 2024-03-07
Make screenshot after custom behaviors


cmillet2127:
Currently it seems screenshot are made before custom behaviors.

It could be very interesting to be able a post-custom behaviors screenshot. For example to capture screenshot after removing the "accept cookies" modals.

ikreymer:
We are using Brave, and the accept cookies modals are actually removed by the browser before our custom behaviors are run, so the screenshot should actually reflect that, I believe. But, your point stands that it could be interesting to take a screenshot after autoscroll, etc..

cmillet2127:
In my situation, I'm encountering an unusual behavior while running Docker on Windows with the most recent image release. When utilizing the browsertrix-crawler, the 'accept cookies' modal persists. However, when navigating manually with Brave browser, the modal does not appear. Initially, I suspected that my image was still employing Chrome, but your confirmation of its use of Brave has led me to reconsider.

Example:
docker run -v c:\tmp\crawls\:/crawls/ -it webrecorder/browsertrix-crawler crawl --url https://www.abarth.fr --generateWACZ final-to-warc --text --wait-until domcontentloaded --screenshot thumbnail,view,fullPage --scopeType page --blockAds

cmillet2127:
> We are using Brave, and the accept cookies modals are actually removed by the browser before our custom behaviors are run, so the screenshot should actually reflect that, I believe. But, your point stands that it could be interesting to take a screenshot after autoscroll, etc..

Indeed, an additional suggestion might involve capturing a screenshot through a custom behavior using a 'utils' method. This approach would allow us to incorporate it into the WARC file, aligning with the methodology used for other screenshots.

ikreymer:
If you run `webrecorder/browsertrix-crawler` it will use `webrecorder/browsertrix-crawler:latest`, which currently still points to the non-Brave version, unless you check out the repo and build it locally. You can try the latest beta release with `webrecorder/browsertrix-crawler:1.0.0-beta.7`.
We hope to release the 1.0.0 version soon and then it will be latest.

fservida:
@ikreymer is this still of interest? it would be extremely useful for us as some websites load images dynamically during scrolling, and therefore are missing if doing a fullpage screenshot before custom behaviours.
I am quite lost in the code as unfamiliar with js, if pointed to right place of screenshot logic I can try something out and provide a PR

PR #656, opened 2024-07-29
Implemented option for FullPage screenshot after the behaviours have run

fservida:
This PR implements a new option to take the FullPage screenshot **after** the behaviours have run instead of before.
This allows to capture elements that might not be fully loaded until the whole page is visited.

I only implemented for fullpage as I think it is the only screenshot where this is useful and the only case where I would need it, as it can include content not currently in frame. Feel free to expand on that as needed.

Related to #486

Tested with custom docker, though I had to ensure setuptools v71 as else it would not build.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1031,6 +1031,21 @@ self.__bx_behaviors.selectMainBehavior();
         if (textextract && this.params.text.includes("final-to-warc")) {
           await textextract.extractAndStoreText("textFinal", true, true);
         }
+
+        if (this.params.screenshot && this.screenshotWriter) {
+          await page.evaluate(() => {
+            window.scrollTo(0, 0);
+          });
+          const screenshots = new Screenshots({
+            browser: this.browser,
+            page,
+            url,
+            writer: this.screenshotWriter,
+          });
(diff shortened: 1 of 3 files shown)


ikreymer:
Thanks for this - i think it would be clearer if it was part of the existing `--screenshot` flag, perhaps called `fullPageAfterBehaviors` - do you mind refactoring it to use that. That way can combine `--screenshot fullPage,fullPageAfterBehaviors` etc...

fservida:
Reopening after adding changes

fservida:
@ikreymer I've reimplemented the changes as suggested, hope this can be useful

tw4l (approved):
Making a new minor changes but overall looks good! Thank you for this.

tw4l:
@fservida Looks like we need to run the auto-formatter before merging to resolve linting issues. Would you mind pushing a commit to this branch after running `yarn run format:fix`? Thanks! :)

ikreymer:
Thinking should rename it to fullPageFinal instead of fullPageAfterBehaviors, since behaviors may not run in some circumstances, and this is more consistent with `final-to-warc` setting with have for text

ikreymer (approved):

PR #667, opened 2024-08-06
QA: Ensure empty string is propagated for comparison

### LINKED ISSUE #666: [Bug]: Pages with no text to extract should be shown as a 100% match
### Browsertrix Version

v1.11.1-973c10c

When visiting https://jamaps.github.io/toronto_building_heights_map/ (a page with only an image on it) there is no text and extracted text shows up as "no data".  To stay consistent with our heuristics, we should mark this as a 100% match in the graph instead of "incomplete" as we do currently.

### Reproduction instructions

1. Run a crawl workflow with a website that includes HTML pages with no text on them
2. Run analysis on that workflow
3. Wait for analysis to fully complete
4. Observe "incomplete" extracted text results for pages that have no text on them.

### Screenshots / Video

Screenshot_2024-08-05_at_12 35 33_PM

Link to archived item on dev: https://dev.browsertrix.com/orgs/demo-org/items/crawl/manual-20240805145814-ce281b23-a83?collectionId=1b7bf4f3-c777-455c-888d-5765ceffd81e#qa

tw4l:
Fixes #666 

The root causes of this issue were that: 1. text (specifically replay text in my testing) was sometimes `undefined` rather than an empty string, and 2. when the max length of both original and replay text strings was 0, we had a divide by zero error.

This PR ensures that both are strings before comparison and that we never divide by zero, which means that if no text was able to be extracted in original or replay the `textMatch` score will now be `1.0` rather than `null`.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -947,7 +947,7 @@ self.__bx_behaviors.selectMainBehavior();
         this.params.text.includes("to-warc"),
       );
 
-      if (text && (this.textInPages || saveOutput)) {
+      if (text !== null && (this.textInPages || saveOutput)) {
         data.text = text;
       }
     }
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,7 +49,7 @@ jobs:
         run: yarn run tsc
 
       - name: build docker
-        run: docker-compose build
+        run: docker compose build
 
       - name: install python deps for docs
         run: pip install mkdocs-material
--- a/src/replaycrawler.ts
+++ b/src/replaycrawler.ts
@@ -565,7 +565,11 @@ export class ReplayCrawler extends Crawler {
 
     const dist = levenshtein(origText, replayText);
     const maxLen = Math.max(origText.length, replayText.length);
-    const matchPercent = (maxLen - dist) / maxLen;
+
+    let matchPercent = 1.0;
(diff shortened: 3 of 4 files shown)


src/crawler.ts
```diff
@@ -947,7 +947,7 @@ self.__bx_behaviors.selectMainBehavior();
         this.params.text.includes("to-warc"),
       );
 
-      if (text && (this.textInPages || saveOutput)) {
+      if (text !== null && (this.textInPages || saveOutput)) {
```
ikreymer, line 950: I think this was the issue with empty string not being set properly

ikreymer:
We definitely want to avoid the div by 0, nice catch.

For the other part, we do actually want to differentiate between `undefined` and empty string. If the text is `undefined`, that means there was some error retrieving either the capture time or QA time text, and we want to flag that, and not compare against the empty string. If both capture and QA text failed, we don't want to show that as a 100% match, so the existing logic there makes sense. We just need to ensure that an empty string text is being passed on correctly - looks like it is in `fetchOrigText()`, but the QA text was not - this is now fixed in: https://github.com/webrecorder/browsertrix-crawler/pull/667/files#diff-e732a18fcf123ca614f71f3527a5fce65b04af5991ae691f36012b78d1fe35acR950 and reverted the other changes.

PR #668, opened 2024-08-13
1.2.8 updates:

ikreymer:
- rewriting: update wabac.js, use getCustomRewriter(), don't truncate POST request bodies for URLs that use a custom rewriter
- browser: disable --enable-automation, setting webdriver = true, so no need for override
- deps: update puppeteer-core, necessary changes for latest puppeteer


src/util/recorder.ts
```diff
@@ -1003,10 +1000,9 @@ export class Recorder {
       case "text/javascript":
       case "application/javascript":
       case "application/x-javascript": {
-        const rules = contentType === "text/html" ? htmlDSRules : baseDSRules;
-        const rw = rules.getRewriter(url);
+        const rw = getCustomRewriter(url, isHTMLMime(contentType));
```
tw4l, line 1003: This is a nice improvement!

tw4l (approved):
LGTM! Tested screenshots to make sure the buffer was working as expected, all seems good.

PR #673, opened 2024-08-26
Streaming in-place WACZ creation + CDXJ indexing

### LINKED ISSUE #674: Use in-place streaming to generate WACZ files
The current method of generating WACZ files in the crawler is less than ideal because it:

- Relies on the py-wacz library, despite having removed other Python dependencies in the crawler; and perhaps more importantly
- Requires the crawler to have at least 2x as much storage available as the total size of the WARC files

We can optimize WACZ creation by generating it in-place, without having to copy the WARC file data or read data multiple times.

Details of this approach:

- WACZ contents are streamed to remote upload (or to disk) from existing files on disk
- CDXJ indices are merged using the linux 'sort' command, and compressed to ZipNum if >50K (or always if using --`generateCDX`)

This should result in significant speed / disk usage improvements: previously the WARC was written once, then read again (for CDXJ indexing), read again (for adding to new WACZ ZIP), written to disk (into new WACZ ZIP), and read again (if uploaded to remote endpoint). Now, WARCs are written once along with the CDXJ, CDXJ is merged, and then all data is read once to either generate WACZ on-disk or upload the WACZ to remote storage (i.e. S3 bucket).

This issue supercedes https://github.com/webrecorder/browsertrix-crawler/issues/484, as it offers additional advantages over switching WACZ generation to the js-wacz library.

ikreymer:
Fixes #674 

This PR supersedes #505, and instead of using js-wacz for optimized WACZ creation:
- generates an 'in-place' or 'streaming' WACZ in the crawler, without having to copy the data again.
- WACZ contents are streamed to remote upload (or to disk) from existing files on disk
- CDXJ indices are merged using the linux 'sort' command, and compressed to ZipNum if >50K (or always if using --generateCDX)
- All data in the WACZ is read / written once
- Should result in significant speed / disk usage improvements: previously WARC was written once, then read again (for CDXJ indexing), read again (for adding to new WACZ ZIP), written to disk (into new WACZ ZIP), read again (if upload to remote endpoint). Now, WARCs are written once, along with the per-WARC CDXJ, CDXJ is then merged on-disk, and all data is read once to either generate WACZ on disk or upload to remote.


src/util/wacz.ts
```diff
@@ -0,0 +1,424 @@
...
+  }
+}
+
+// Merge CDX
+export function addDirFiles(fullDir: string): string[] {
+  const files = fs.readdirSync(fullDir);
+  return files.map((name) => path.join(fullDir, name));
+}
+
+export async function mergeCDXJ(
```
tw4l, line 311: One difference from py-wacz I can spot is that the resulting `index.cdx` no longer has a first line of `!meta 0 {"format": "cdxj-gzip-1.0", "filename": "index.cdx.gz"}`, should be simple enough to add

tests/basic_crawl.test.js
```diff
@@ -8,9 +8,9 @@ test("ensure basic crawl run with docker run passes", async () => {
     'docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url https://example.com/ --generateWACZ  --text --collection wr-net --combineWARC --rolloverSize 10000 --workers 2 --title "test title" --description "test description" --warcPrefix custom-prefix',
   );
 
-  child_process.execSync(
-    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler wacz validate --file collections/wr-net/wr-net.wacz",
-  );
+  // child_process.execSync(
```
tw4l, line 11: I wonder if it's worth using py-wacz just in some of the tests to validate that the resulting WACZs are valid, or implementing validation in the new `wacz` crawler module to do the same. They do pass validation in my manual testing but would be good to have that in the test suite to catch any potential regressions moving forward given how important this step of the crawling process is.

tw4l (commented):
A few initial comments/suggestions, continuing testing though so far it's looking pretty good!

src/util/wacz.ts
tw4l: Granted this is possibly unnecessary but seems like a good practice?

src/util/wacz.ts
```diff
@@ -0,0 +1,424 @@
...
+    await pipeline(this.generate(), fs.createWriteStream(filename));
+  }
+
+  async *iterDirForZip(files: string[]): AsyncGenerator<InputWithoutMeta> {
+    const encoder = new TextEncoder();
+    const end = new EndMarker();
+    // correctly handles DST
+    const hoursOffset = (24 - new Date(0).getHours()) % 24;
+    const timezoneOffset = hoursOffset * 60 * 60 * 1000;
+    //const timezoneOffset = new Date().getTimezoneOffset() * 60000;
```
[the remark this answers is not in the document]
ikreymer: This was adapted from archiveweb.page - where there no actual WARC files on disk, so had to use a fixed date, accounting for timezone, to ensure consistent WACZ files on each download of the WACZ in the extension. However, we do have actual WARC files with last modified dates, so can probably use real dates here.

src/util/wacz.ts
```diff
@@ -0,0 +1,424 @@
...
+  async generateToFile(filename: string) {
+    await pipeline(this.generate(), fs.createWriteStream(filename));
+  }
+
+  async *iterDirForZip(files: string[]): AsyncGenerator<InputWithoutMeta> {
+    const encoder = new TextEncoder();
+    const end = new EndMarker();
+    // correctly handles DST
+    const hoursOffset = (24 - new Date(0).getHours()) % 24;
+    const timezoneOffset = hoursOffset * 60 * 60 * 1000;
```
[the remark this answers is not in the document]
ikreymer: As mentioned above, we might not need to do this at all actually.

src/util/wacz.ts
ikreymer: ah yeah, forgot about this, should re-add for consistency.

src/util/wacz.ts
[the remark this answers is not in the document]
ikreymer: Edit: removed previous comment, confused myself - we do still need the timezone offset!

src/util/wacz.ts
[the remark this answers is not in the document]
ikreymer: Looking at this again, the timezone if always UTC in the Docker container, unless changed otherwise. But ZIP file contents are generally treated as local timezone, unless an 'extended timestamp field' is used (which I don't think the client-zip library supports) or we pass the offset to the docker container somehow. The result is that when running on non-UTC host, the files in the WACZ do have the wrong timestamp. However, this is already the case with py-wacz, so maybe something we address separately if we want to address it. Otherwise, the WARCs in the WACZ do have a UTC timestamp always.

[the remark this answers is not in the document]
tw4l: Interesting. I didn't know that particular quirk of the ZIP format. Seems like maybe we should aim for parity with py-wacz/current behavior for now and create a separate issue to track fixing this on non-UTC hosts.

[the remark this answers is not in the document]
ikreymer: yep, just removing the offset stuff altogether keeps that parity, did that for now.

tests/basic_crawl.test.js
ikreymer: yeah, we can run py-wacz outside of Docker and install just for CI, added that

ikreymer:
Updated to latest released warcio.js / wabac.js, should be good to merge when everything else is resolved.

src/util/wacz.ts
ikreymer: Added for index.idx in https://github.com/webrecorder/browsertrix-crawler/pull/673/commits/d2c172038bc2b80a28ac9653948a53d8bcc43f51

src/util/warcwriter.ts
```diff
@@ -60,8 +60,6 @@ export class WARCWriter implements IndexerOffsetLength {
   }) {
     this.archivesDir = archivesDir;
     this.tempCdxDir = tempCdxDir;
-    // for now, disabling CDX
-    this.tempCdxDir = undefined;
```
ikreymer, line 64: One thing to add regarding tempCdxDir - it's now required to be kept so that we don't have to reindex the WARCs again, if adding to an existing collection. Perhaps should rename to 'warcCdx' or some other name to indicate its not as temporary, eg. it'll be needed again if adding to collection or regenerating WACZ again, so we don't have to reindex.

ikreymer: We could regenerate the CDX from WARCs, but no reason to delete them as they're pretty small compared to the WARCs, so just keeping them around in the collection

tw4l: Good point! I'm in favor of renaming it, and might be good to document what this is and when you might need it in the crawler docs.

ikreymer: Renamed to `warcCdxDir` and using `warc-cdx` instead of `tmp-cdx` for the dir name, in https://github.com/webrecorder/browsertrix-crawler/pull/673/commits/fb427eae146a84e89d8451c037dd5f0ba9464b4f

tw4l: That makes sense. Only thing is I think if I didn't have the "dev on the team" perspective, I might be confused why there are indexes in both `warc-cdx` and `indexes` (assuming `--generateCDX` or `--generateWACZ` was used), and how they differ, so would be good to add something to the documentation about this before merging.

src/util/wacz.ts
```diff
@@ -0,0 +1,424 @@
...
+
+    let cdxLines: string[] = [];
+
+    let key = "";
+    let count = 0;
+
+    idxFile.write(
+      `!meta 0 ${JSON.stringify({
+        format: "cdxj-gzip-1.0",
+        filename: INDEX_IDX,
```
tw4l, line 331:
```suggestion
        filename: INDEX_CDX_GZ,
```

Small regression, this should point to the gzipped cdxj, not itself

src/util/warcwriter.ts
tw4l: Created new issue for that so as not to block this PR: https://github.com/webrecorder/browsertrix-crawler/issues/675

tw4l (approved):

PR #678, opened 2024-09-05
Additional direct fetch improvements

ikreymer:
- use existing headersTimeout in undici to limit time to headers fetch to 30 seconds, reject direct fetch if timeout is reached
- allow full page timeout for loading payload via direct fetch
- support setting global fetch() settings
- add markPageUsed() to only reuse pages when not doing direct fetch
- apply auth headers to direct fetch
- catch failed fetch and timeout errors
- support failOnFailedSeeds for direct fetch, ensure timeout is working

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -48,6 +48,9 @@ import {
   BEHAVIOR_LOG_FUNC,
   DEFAULT_SELECTORS,
   DISPLAY,
+  FETCH_HEADERS_TIMEOUT_SECS,
+  PAGE_OP_TIMEOUT_SECS,
+  SITEMAP_INITIAL_FETCH_TIMEOUT_SECS,
 } from "./util/constants.js";
 
 import { AdBlockRules, BlockRules } from "./util/blockrules.js";
@@ -80,10 +83,6 @@ const behaviors = fs.readFileSync(
   ),
   { encoding: "utf8" },
 );
-
-const FETCH_TIMEOUT_SECS = 30;
-const PAGE_OP_TIMEOUT_SECS = 5;
-const SITEMAP_INITIAL_FETCH_TIMEOUT_SECS = 30;
 
 const RUN_DETACHED = process.env.DETACHED_CHILD_PROC == "1";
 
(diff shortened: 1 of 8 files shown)


ikreymer:
To be merged after #677

src/crawler.ts
```diff
@@ -876,14 +871,24 @@ self.__bx_behaviors.selectMainBehavior();
 
     if (directFetchCapture) {
       try {
-        const { fetched, mime, ts } = await timedRun(
-          directFetchCapture({ url, headers: this.headers, cdp }),
+        const headers = auth ? { Authorization: auth } : this.headers;
```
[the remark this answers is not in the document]
ikreymer: good catch, this was a typo!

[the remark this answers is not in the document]
ikreymer: Was supposed to be `auth ? { Authorization: auth, ...this.headers } : this.headers;` now fixed!

ikreymer:
Removed the extra fetch timeout as suggested in #677 also, as this timeout is mostly unneeded and just pads total time most of the time, since its really direct fetch or regular fetch (and direct fetch will still timeout if no headers received after 30 seconds).

src/crawler.ts
```diff
@@ -900,15 +904,33 @@ self.__bx_behaviors.selectMainBehavior();
...
       }
     }
 
+    opts.markPageUsed();
+
+    if (auth) {
+      await page.setExtraHTTPHeaders({ Authorization: auth });
+      opts.isAuthSet = true;
+    } else if (opts.isAuthSet) {
+      await page.setExtraHTTPHeaders({});
```
tw4l, line 931: I know this is just moved not added, but what is setting empty extra headers doing for us? Do we maybe want to be doing `await page.setExtraHTTPHeaders(...this.headers)` here to set the user agent?

tw4l (commented):
Seems to be working well in testing with example crawl provided in the original issue, but one question and it seems like we have test failures that should be investigated

src/crawler.ts
ikreymer: This is just supposed to clear the Authorization header, which is added explicitly. The user agent is set separately on the browser via `page.setUserAgent()` in browser. These settings are per-page.

tw4l: Ah right, makes sense, thanks!

ikreymer:
> Seems to be working well in testing with example crawl provided in the original issue, but one question and it seems like we have test failures that should be investigated (will see if bumped timeout resolves that)

The test involved crawling webrecorder.net/community and youtube was occasionally taking a long time, excluded the page from test, since that's not what's being tested there..

tw4l (approved):

PR #679, opened 2024-09-05
WARC writer + incremental indexing fixes

ikreymer:
- ensure WARC rollover happens only after response/request + cdx or single record + cdx have been written
- ensure request payload is buffered for POST request indexing
- update to warcio 2.3.1 for POST request case-insensitive 'content-type' check
- recorder: remove unused 'tempdir', no longer used as warcio chooses a temp file on it's own

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "testTimeout": 90000
   },
   "resolutions": {
-    "wrap-ansi": "7.0.0"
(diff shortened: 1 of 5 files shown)


ikreymer:
To be merged after warcio.js 2.3.1 is released after https://github.com/webrecorder/warcio.js/pull/78

PR #689, opened 2024-09-18
support custom css selectors for extracting links

### LINKED ISSUE #217: Specifying selectors for extracting links.
I came across a site which uses an `<area>` tag with an `href` attribute to create links with a non-standard shape. I don't know if this is the correct way to approach this, but I was able to capture these links by implementing the following custom driver.
```js
module.exports = async ({data, page, crawler}) => {
  await crawler.loadPage(page, data, [
      {selector:"a[href]", extract:"href", isAttribute:false},
      {selector:"area[href]", extract:"href", isAttribute:false}
  ]);
};
```

However, I did not see anything in the documentation hinting at this and it required reading through the source code to even determine that the driver is what I should be looking into.

Furthermore, I've noticed that `defaultDriver.js` has changed significantly over time, so it is not clear to me whether this approach will remain valid in the long run. And to emphasize that point, it is worth mentioning that this driver works in 0.7.1 but breaks in 0.8.0-beta.1 (though I realize that fixing it just requires changing `module.exports =` to `export default`).

Would you consider implementing an easier way to configure the link extraction selectors? Or, if a custom driver is the recommended approach, is this documented somewhere?

ikreymer:
- specify selectors via --selectLinks property
- selectors are a comma separate list of the form [css selector]->[property to use] or [css selector]->@[attribute to use], default being 'a[href]->href'
- fixes #217


ikreymer:
Still need:
- tests
- validation of selectors?

src/util/argParser.ts
```diff
@@ -156,6 +160,14 @@ class ArgParser {
             "Allow Hashtag URLs, useful for single-page-application crawling or when different hashtags load dynamic content",
         },
 
+        selectLinks: {
+          describe:
+            "one or more selectors for extracting links, in the format [css selector]->[property to use],[css selector]->@[attribute to use]",
```
tw4l, line 165:
```suggestion
            "One or more selectors for extracting links, in the format [css selector]->[property to use],[css selector]->@[attribute to use]",
```
We're admittedly not super consistent about this, but most of the descriptions seem to use sentence casing

docs/docs/user-guide/cli-options.md
```diff
@@ -50,6 +50,11 @@ Options:
                                             e-page-application crawling or when
                                             different hashtags load dynamic cont
                                             ent
+      --selectLinks                         one or more selectors for extracting
+                                             links, in the format [css selector]
+                                            ->[property to use],[css selector]->
+                                            @[attribute to use]
+                                            [array] [default: ["a[href]->href"]]
```
tw4l, line 57: This is a good reference but I feel like this is both an important feature and one where the syntax isn't the most straightforward, so I'd suggest we add a brief section in the Commonly-Used Options page of the docs with an example or two?

src/util/argParser.ts
```diff
@@ -714,6 +725,23 @@ class ArgParser {
...
+        const parts = x.split("->");
+        const selector = parts[0];
+        const value = parts[1] || "";
+        const extract = parts.length > 1 ? value.replace("@", "") : "href";
+        const isAttribute = value.startsWith("@");
+        return { selector, extract, isAttribute };
+      });
+    } else {
+      selectLinks = DEFAULT_SELECTORS;
+    }
```
tw4l, line 741: If the syntax for a link selector passed in is broken, the failure is a bit silent - i.e. the crawler obviously won't find other pages, but it does finish with a 0 exit code and no error messages. I think that's not the most user-friendly behavior here. At minimum I'd suggest we log an error message, but I'm wondering if we should just fail the crawl early with a fatal message given how important link extraction is to almost all crawl scope types.

tw4l (commented):
Nice feature and thanks for the tests! Removing the default driver is a big step as well.

I think this is nearly there but we need to improve the user experience around it a little bit around failure cases and helping people a bit with the expected syntax.

src/util/argParser.ts
ikreymer: Added validation with https://www.npmjs.com/package/css-selector-parser, will fail instantly on invalid selectors.

tw4l (approved):
Beautiful, thank you!

PR #694, opened 2024-09-30
ensure extraHops also apply to maxDepth

### LINKED ISSUE #693: [Bug]: Crawl Configuration Inconsistency: Max Depth and Include Any Linked Page
### Browsertrix Version

v1.11.7-7a61568

I have a question about the settings Max Depth and Include Any Linked Page. 
I think there might me an error in the implementation of the rules. 

As I understand it, 
- the max depth apply to pages that are in scope. That means, pages would be cut, due to max depth. It limits the pages, that are in scope. "Limits how many hops away the crawler can visit while staying within the Start URL Scope."
- The one hop out adds all pages, that are linked by pages in scope. 
"If checked, the crawler will visit pages one link away outside of Start URL Scope."

Right now, the crawler from browsertix is not behaving as I would think.

### Reproduction instructions

**Example:**
I've created an example page for illustration: https://monaulrich.online/scopetest/start/

**Crawl Config "Scope Domain & Subdomain / Max Depth 1"**
Here the Crawler includes what I would expect. 
The start url and the 4 linked page, that have the same domain/subdomain are archived. 
These pages are in scope.
example_structure_in_scope

**Crawl Config "Scope Domain & Subdomain / Max Depth 1 / Include Any Linked Page: True"**
I would expect, that all the pages, that are in scope based on the settings Scope Domain & Subdomain and Max Depth: 1, are checked for any linked pages. But only the pages 1 depth away from the start URL are included. 

Browsertrix Output:
example_structure_scope_linked_pages_actual

What I would expect:
example_structure_scope_linked_pages_expected

I am not sure, if i misunderstand the settings, or if there is an inconsistency in the implementation.
Thanks in advance.

### Screenshots / Video

_No response_

### Environment

_No response_

### Additional details

_No response_

ikreymer:
- if extraHops is set, crawler should visit pages beyond maxDepth
- currently returning out of scope at depth limit even if extraHops is set
- adjust isInScope and isAtMaxDepth to account for extraHops
- fixes #693

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1921,7 +1921,7 @@ self.__bx_behaviors.selectMainBehavior();
 
     //data.filteredFrames = await page.frames().filter(frame => this.shouldIncludeFrame(frame, logDetails));
 
-    const { seedId } = data;
+    const { seedId, extraHops } = data;
 
     const seed = await this.crawlState.getSeedAt(
       this.seeds,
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Code changes look good and tested against reproduction example in the issue

PR #697, opened 2024-10-04
fix typo in QA exclude check, which resulted in all URLs being excluded

ikreymer:
- ensure exclusions now work as expected in replay mode
- add test for using --exclude with replay

### DIFF
--- a/tests/qa_compare.test.js
+++ b/tests/qa_compare.test.js
@@ -8,7 +8,7 @@ test("run initial crawl with text and screenshots to prepare for QA", async () =
   fs.rmSync("./test-crawls/qa-wr-net", { recursive: true, force: true });
 
   child_process.execSync(
-    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url https://webrecorder.net/ --url https://webrecorder.net/about --url https://browsertrix.com/ --scopeType page --collection qa-wr-net --text to-warc --screenshot view --generateWACZ",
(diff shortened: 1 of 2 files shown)


tests/qa_compare.test.js
```diff
@@ -20,7 +20,7 @@ test("run QA comparison, with write pages to redis", async () => {
   fs.rmSync("./test-crawls/qa-wr-net-replay", { recursive: true, force: true });
 
   const child = child_process.exec(
-    "docker run -p 36380:6379 -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler qa --qaSource /crawls/collections/qa-wr-net/qa-wr-net.wacz --collection qa-wr-net-replay --crawlId test --qaDebugImageDiff --writePagesToRedis --debugAccessRedis",
+    "docker run -p 36380:6379 -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler qa --qaSource /crawls/collections/qa-wr-net/qa-wr-net.wacz --collection qa-wr-net-replay --crawlId test --qaDebugImageDiff --writePagesToRedis --debugAccessRedis --exclude contact",
```
tw4l, line 23: Maybe could be more specific in the tests below to ensure this was excluded? As is it'll stop checking pages after the third, so the test wouldn't necessarily let us know if contact was included I think.

tw4l (approved):
Very good catch!

tests/qa_compare.test.js
ikreymer: done!

ikreymer: though check was more to see that we still get 3 pages, if exclude is flipped, we'd exclude anything that wasn't 'contact'

PR #701, opened 2024-10-10
link extraction promise cleanup:

ikreymer:
- catch frame.evaluate() directly and log errors there to avoid any possibility of exception being propagated before wrapping in timedRun()
- also add clearTimeout() to timedRun()
- possibly fixes openzim/zimit#376

### DIFF
--- a/src/util/timing.ts
+++ b/src/util/timing.ts
@@ -18,14 +18,16 @@ export function timedRun(
   // return Promise return value or log error if timeout is reached first
   const timeout = seconds * 1000;
 
+  let tm: NodeJS.Timeout;
+
   const rejectPromiseOnTimeout = (timeout: number) => {
     return new Promise((resolve, reject) => {
-      setTimeout(() => reject("timeout reached"), timeout);
+      tm = setTimeout(() => reject("timeout reached"), timeout);
     });
   };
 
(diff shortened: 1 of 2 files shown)


src/util/timing.ts
```diff
@@ -38,8 +40,8 @@ export function timedRun(
         //logger.error("Unknown exception", {...errJSON(err), ...logDetails}, context);
         throw err;
```
tw4l, line 41: It looks like we have a few uses of `timedRun` that aren't wrapped in a tightly scoped try/catch block (e.g. setting page title in the `crawlPage` method, running behaviors) but rather a higher-level one in the worker that will catch and log a "Worker Exception". I'm wondering if we might want to log the error closer to where it occurs or append some contextual information to the error message to help debug when something goes wrong?

tw4l (approved):
Overall looks good, one question about where we're catching and logging the errors being thrown that's not a showstopper but maybe worth thinking about a bit

src/util/timing.ts
ikreymer: Hm, i think that this should still catch the original error, but maybe worth another look.
This was removed to avoid redundancy error message - if the timedRun() throws, its not a timeout, but an actual exception that probably should interrupt the crawling of that page...

PR #707, opened 2024-10-22
Support loading custom behaviors from URLs and/or filepaths

### LINKED ISSUE #368: Load Custom Behaviour from URL

tw4l:
Fixes #368 

The `--customBehaviors` flag is now an array, making it repeatable. This should be backwards compatible with the CLI flag, but may require changes to YAML configs when custom behaviors are used.

Custom behaviors can be loaded from URLs, local filepaths, and paths to local directories, including any combination thereof.

New tests are added to ensure loading behaviors from URLs as well as a mixed combination of URL and filepath works as expected.

### DIFF
--- a/docs/docs/user-guide/cli-options.md
+++ b/docs/docs/user-guide/cli-options.md
@@ -248,7 +248,7 @@ Options:
                                                       [boolean] [default: false]
       --customBehaviors                     Custom behavior files to inject. Val
                                             ues can be URLs, paths to individual
-                                             behavior files, or pathsto a direct
+                                            behavior files, or paths to a direct
                                             ory of behavior files.
                                                            [array] [default: []]
       --debugAccessRedis                    if set, runs internal redis without
--- a/docs/docs/user-guide/cli-options.md
+++ b/docs/docs/user-guide/cli-options.md
@@ -246,9 +246,11 @@ Options:
                                             ailOnFailedSeed may result in crawl
                                             failing due to non-200 responses
                                                       [boolean] [default: false]
-      --customBehaviors                     injects a custom behavior file or se
-                                            t of behavior files in a directory
-                                                                        [string]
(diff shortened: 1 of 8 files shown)


src/util/argParser.ts
```diff
@@ -551,8 +551,10 @@ class ArgParser {
 
         customBehaviors: {
           describe:
-            "injects a custom behavior file or set of behavior files in a directory",
-          type: "string",
+            "Custom behavior files to inject. Values can be URLs, paths to individual behavior files, or paths" +
+            "to a directory of behavior files.",
```
tw4l, line 555:
```suggestion
            " to a directory of behavior files",
```

docs/docs/user-guide/cli-options.md
```diff
@@ -246,9 +246,11 @@ Options:
                                             ailOnFailedSeed may result in crawl
                                             failing due to non-200 responses
                                                       [boolean] [default: false]
-      --customBehaviors                     injects a custom behavior file or se
-                                            t of behavior files in a directory
-                                                                        [string]
+      --customBehaviors                     Custom behavior files to inject. Val
+                                            ues can be URLs, paths to individual
+                                             behavior files, or pathsto a direct
+                                            ory of behavior files.
```
tw4l, line 252:
```suggestion
                                            behavior files, or paths to a direct
                                            ory of behavior files.
```

ikreymer (approved):

Issue #706, opened 2024-10-22
Browser disconnected (crashed?)


rgaudin:
This week, with Browsertrix-Crawler 1.3.3 (with warcio.js 2.3.1), I am getting several cases of the following:

- a website is being crawled
- there are videos
- behaviors are ran
- `Browser disconnected (crashed?), interrupting crawl` log
- Then `Failed to load response body`
- Then `Large payload written to WARC, but not returned to browser (would require rereading into memory)`
- Then `Rollover size exceeded, creating new WAR`
- limit not hit is logged: `{"crawled":27,"total":690,"pending":0,"failed":0,"limit":{"max":0,"hit":false}`
- crawler exits as interrupted `Exiting, Crawl status: interrupted`

I don't know of those connect to each other but This happened on multiple different websites and it happens consistently.

You can try with `https://fsfe.org/`

```
{"timestamp":"2024-10-15T08:43:10.788Z","logLevel":"warn","context":"recorder","message":"continueResponse failed","details":{"url":"https://download.fsfe.org/videos/peertube/xs29yhLxSP1uKLYkSeoKKp_720p.mp4"}}
{"timestamp":"2024-10-15T08:43:10.806Z","logLevel":"warn","context":"recorder","message":"continueResponse failed","details":{"url":"https://download.fsfe.org/videos/peertube/xs29yhLxSP1uKLYkSeoKKp_720p.mp4"}}
{"timestamp":"2024-10-15T08:43:10.823Z","logLevel":"warn","context":"recorder","message":"continueResponse failed","details":{"url":"https://download.fsfe.org/videos/peertube/xs29yhLxSP1uKLYkSeoKKp_720p.mp4"}}
{"timestamp":"2024-10-15T08:43:10.844Z","logLevel":"warn","context":"recorder","message":"continueResponse failed","details":{"url":"https://download.fsfe.org/videos/peertube/xs29yhLxSP1uKLYkSeoKKp_720p.mp4"}}
{"timestamp":"2024-10-15T08:43:10.869Z","logLevel":"warn","context":"recorder","message":"continueResponse failed","details":{"url":"https://download.fsfe.org/videos/peertube/xs29yhLxSP1uKLYkSeoKKp_720p.mp4"}}
[...]

PR #709, opened 2024-10-27
various edge-case loading optimizations:

ikreymer:
- if too many range requests for same URL are being made, try skipping/failing right away to reduce load
- assume main browser context is used not just for service workers, always enable
- check false positive 'net-aborted' error that may actually be ok for media, as well as documents
- improve logging
- possible fix for issues in #706
- interrupt any pending requests (that may be loading via browser context) after page timeout, log dropped requests

### DIFF
--- a/src/util/browser.ts
+++ b/src/util/browser.ts
@@ -6,7 +6,7 @@ import { Readable } from "node:stream";
 import os from "os";
 import path from "path";
 
-import { LogContext, logger } from "./logger.js";
+import { formatErr, LogContext, logger } from "./logger.js";
 import { initStorage } from "./storage.js";
 
 import { DISPLAY, type ServiceWorkerOpt } from "./constants.js";
@@ -126,7 +126,7 @@ export class Browser {
         ? undefined
         : (target) => this.targetFilter(target),
     };
-    await this._init(launchOpts, ondisconnect, recording);
+    await this._init(launchOpts, ondisconnect);
   }
 
   targetFilter(target: Target) {
@@ -392,7 +392,7 @@ export class Browser {
     launchOpts: PuppeteerLaunchOptions,
     // eslint-disable-next-line @typescript-eslint/ban-types
     ondisconnect: Function | null = null,
-    recording: boolean,
+    //_recording: boolean,
   ) {
     this.browser = await puppeteer.launch(launchOpts);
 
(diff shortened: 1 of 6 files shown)


tw4l:
I'm wondering if we should also move the `"Large payload written to WARC, but not returned to browser (would require rereading into memory)"` log messages to debug since users can interpret it as an error rather than expected behavior

src/util/browser.ts
```diff
@@ -392,17 +392,15 @@ export class Browser {
     launchOpts: PuppeteerLaunchOptions,
     // eslint-disable-next-line @typescript-eslint/ban-types
     ondisconnect: Function | null = null,
-    recording: boolean,
+    //_recording: boolean,
```
tw4l, line 395: Maybe just remove instead of commenting out?

```suggestion
```

src/util/recorder.ts
```diff
@@ -426,15 +429,13 @@ export class Recorder {
       case "net::ERR_ABORTED":
         // check if this is a false positive -- a valid download that's already been fetched
         // the abort is just for page, but download will succeed
-        if (type === "Document" && reqresp.isValidBinary()) {
+        if (
+          type === "Document" ||
+          (type === "Media" && reqresp.isValidBinary())
+        ) {
```
tw4l, line 435: Just double-checking, did you mean to remove the check for `reqresp.isValidBinary() for downloads of type `Document` or should we shift the parentheses around a bit?

tw4l (approved):
Looks good, a few small comments. I tested against a few of the examples in https://github.com/webrecorder/browsertrix-crawler/issues/706#issuecomment-2441017392 and am not seeing any browser crashes, though these were admittedly smaller-scoped crawls.

ikreymer:
This is ready for re-review - though probably want to merge test fixes from #710 first

ikreymer:
Main changes are actually streaming 206 responses (!) which were before always loaded into memory, also failing duplicate 206 responses that are not from 0- to avoid additional load.

tw4l (approved):
Nice! Great catch on the 206 streaming issue. Just left one small suggestion

src/util/recorder.ts
```diff
@@ -1031,6 +1075,40 @@ export class Recorder {
...
+    ) {
+      return true;
+    }
+
+    // if contentLength is unknown, also stream if its an essential resource and not 3xx / 4xx / 5xx
+    // status code, as these codes may have no content-length, and are likely small
+    if (
+      contentLength < 0 &&
+      !this.isEssentialResource(resourceType, mimeType) &&
+      reponseStatusCode >= 200 && responseStatusCode < 300
```
tw4l, line 1104:
```suggestion
      responseStatusCode >= 200 && responseStatusCode < 300
```

my bad!

PR #713, opened 2024-11-05
fix cookie not being passed to replay regression: for now, add x-waba…

ikreymer:
…c-preset-cookie header for quick fix (same as AWP)

longer fix will involve fixing CDXJ to include the header

Fix replay of X captures and other pages requiring cookies

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -1751,6 +1751,13 @@ function createResponse(
 
   const httpHeaders = reqresp.getResponseHeadersDict(reqresp.payload.length);
 
+  const cookie =
+    reqresp.requestHeaders &&
+    (reqresp.requestHeaders["cookie"] || reqresp.requestHeaders["Cookie"]);
+  if (cookie) {
+    httpHeaders["x-wabac-preset-cookie"] = cookie;
+  }
+
   const warcHeaders: Record<string, string> = {
     "WARC-Page-ID": pageid,
   };


src/util/recorder.ts
```diff
@@ -1751,6 +1751,13 @@ function createResponse(
 
   const httpHeaders = reqresp.getResponseHeadersDict(reqresp.payload.length);
 
+  const cookie =
+    reqresp.requestHeaders &&
+    (reqresp.requestHeaders["cookie"] || reqresp.requestHeaders["Cookie"]);
+  if (cookie) {
+    httpHeaders["x-wabac-preset-cookie"] = cookie;
+  }
+
```
tw4l, line 1760: I think unlike the implementation in AWP, `cookie` here is just a boolean because we're testing whether reqresp.requestHeaders["cookie"] or reqresp.requestHeaders["Cookie"] exist but never actually setting `cookie` as their value.

Instead we probably want to do something like:

```suggestion
  const hasCookie =
    reqresp.requestHeaders &&
    (reqresp.requestHeaders["cookie"] || reqresp.requestHeaders["Cookie"]);
  if (hasCookie) {
    const cookie = reqresp.requestHeaders["cookie"] || reqresp.requestHeaders["Cookie"];
    httpHeaders["x-wabac-preset-cookie"] = cookie;
  }

```

tw4l: Oh but it does seem to work when testing, so maybe I'm mistaken! Could be I just don't understand properly what this is actually doing

tw4l: TIL something about Javascript 😆

tw4l (approved):

PR #714, opened 2024-11-07
fix indexing of cookie header:

ikreymer:
- add fields option for adding req.http:cookie and referrer entries to the cdxj
- update to warcio 2.4.0 to support this functionality
 
Pending release of warcio.js 2.4.0


tw4l (commented):
Nice! Once we release warcio 2.4.0 and update the reference to a non-beta release here, I think this is good to merge.

Could also add a test here to ensure that the cookie headers end up in the index

tw4l (approved):

PR #715, opened 2024-11-07
Add WARC-Protocol header

### LINKED ISSUE #641: Add support for WARC-Protocol and WARC-Cipher-Suite headers
It should be fairly easy to add support for WARC-Protocol (https://github.com/iipc/warc-specifications/issues/42)
and likely for the WARC-Cipher-Suite (https://github.com/iipc/warc-specifications/issues/86) headers.

ikreymer:
- add WARC-Protocol header combining http + tls protocols as repeated headers
- ~~add WARC-Cipher-Suite header, mapping Chrome NetworkSecurityDetails to known cipher suites~~
- fixes #641

A few caveats:
For now, just adding WARC-Protocol here as WARC-Cipher-Suite needs more testing.
~~- The WARC-Cipher-Suite data is not directly available from the browser, so must be inferred based on the available info.~~

### DIFF
--- a/tests/pageinfo-records.test.js
+++ b/tests/pageinfo-records.test.js
@@ -24,10 +24,9 @@ test("run warc and ensure pageinfo records contain the correct resources", async
   let foundInvalid = false;
 
   for await (const record of parser) {
-    if (record.warcType === "response" && 
+    if (record.warcType === "response" &&
       (record.warcTargetURI === "https://oldwebrecorder.net/" || record.warcTargetURI === "https://old.webrecorder.net/about")) {
       expect(record.warcHeaders.headers.get("WARC-Protocol")).toBe("h2, tls/1.3");
-      expect(record.warcHeaders.headers.get("WARC-Cipher-Suite")).toBe("TLS_AES_128_GCM_SHA256");
     }
 
     if (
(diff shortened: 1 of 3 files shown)


tw4l:
Re: the WARC-Protocol header being comma separated rather than repeated, it looks like the IIPC membership is trying to get consensus on whether this is acceptable this week. See ongoing discussion here: https://github.com/iipc/warc-specifications/issues/42

ikreymer:
> One suggestion re: import cleanup.
> 
> Community consensus in [iipc/warc-specifications#42](https://github.com/iipc/warc-specifications/issues/42) seems to be for repeated headers rather than a single header with a comma-separated list so we should probably modify this PR to go with that approach.

Updated to now generate multiple WARC-Protocol headers, per consensus there.

ikreymer:
Though, may also want to get clarification on WARC-Cipher-Suite since its not an exact one-to-one mapping there..

tests/pageinfo-records.test.js
```diff
@@ -24,6 +24,11 @@ test("run warc and ensure pageinfo records contain the correct resources", async
   let foundInvalid = false;
 
   for await (const record of parser) {
+    if (record.warcType === "response") {
+      expect(record.warcHeaders.headers.get("WARC-Protocol")).toBe("h2, tls/1.3");
```
tw4l, line 28: Seeing a test failure here. Getter might not be working as expected due to repeated headers now?

tw4l (approved):
Tested independently with the crawler and this is working as expected. Nice work.

PR #717, opened 2024-11-08
Support loading custom behaviors from git repo

### LINKED ISSUE #712: Support loading behaviors from a Git repo with branch + path
As a follow-up to #707, perhaps it'd be helpful to be able to specify a git repo, which the crawler would then clone locally.
Looks like we already have `git` installed in the image. The main addition would be how to specify the repo, for max flexibility.

Perhaps it could be: `--customBehaviors git+https://git.example.com/repo.git@branch#root/path` where the `@branch` and `/root/path` would be option.

Eg. given `git+https://git.example.com.git@branch#!/root/path`, the crawler would do:

```
git clone --branch branch https://git.example.com.git /tmp/behaviors-repo
cd /tmp/behaviors-repo
cd root/path
```
and then grab all of the behaviors from the current directory.

tw4l:
Fixes #712 

Also expands the existing documentation about behaviors and adds a test.

For the branch and relative path information I elected to go with query strings added to the URL rather than implementing a custom syntax. I think it works well but let me know what you think

### DIFF
--- a/src/util/file_reader.ts
+++ b/src/util/file_reader.ts
@@ -2,8 +2,12 @@ import fsp from "fs/promises";
 import path from "path";
 import crypto from "crypto";
 import { fetch } from "undici";
+import util from "util";
+import { exec as execCallback } from "child_process";
 
 import { logger } from "./logger.js";
+
+const exec = util.promisify(execCallback);
 
 const MAX_DEPTH = 2;
 
@@ -23,7 +27,10 @@ export async function collectCustomBehaviors(
   const collectedSources: FileSources = [];
 
   for (const fileSource of sources) {
-    if (fileSource.startsWith("http")) {
+    if (fileSource.startsWith("git+")) {
+      const newSources = await collectGitBehaviors(fileSource);
+      collectedSources.push(...newSources);
+    } else if (fileSource.startsWith("http")) {
       const newSources = await collectOnlineBehavior(fileSource);
       collectedSources.push(...newSources);
     } else {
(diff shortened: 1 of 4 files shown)


tests/custom-behavior.test.js
```diff
@@ -71,6 +71,41 @@ test("test mixed custom behavior sources", async () => {
   ).toBe(true);
 });
 
+test("test custom behaviors from git repo", async () => {
+  const res = child_process.execSync(
+    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url https://specs.webrecorder.net/ --url https://example.org/ --url https://old.webrecorder.net/ --customBehaviors \"git+https://github.com/webrecorder/browsertrix-crawler.git?branch=custom-behaviors-testing&path=tests/custom-behaviors\" --scopeType page",
```
ikreymer, line 76: I guess we should change this to main for merge, so it uses this branch itself?

src/util/file_reader.ts
```diff
@@ -35,6 +42,45 @@ export async function collectCustomBehaviors(
...
 
+async function collectGitBehaviors(gitUrl: string): Promise<FileSources> {
+  const url = gitUrl.split("git+").pop() || "";
+  const params = new URL(url).searchParams;
+  const branch = params.get("branch") || "";
+  const relPath = params.get("path") || "";
+  const urlStripped = url.split("?")[0];
+
+  const timestamp = Date.now();
+  const tmpDir = `/tmp/behaviors-repo-${timestamp}`;
```
ikreymer, line 53: I suppose very little change of conflict here, but could also use a random string, maybe for consistency

ikreymer, line 53:
```suggestion
  const tmpDir = `/tmp/behaviors-repo-${crypto.randomBytes(4).toString("hex")}`;
```

Maybe just generate a dir with random values, similar to other place in this file.

tests/custom-behavior.test.js
tw4l: We can switch it to main, but I also created the referenced `custom-behaviors-testing` branch so as long as we don't delete that it should continue to be okay.

tw4l, line 76:
```suggestion
    "docker run -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url https://specs.webrecorder.net/ --url https://example.org/ --url https://old.webrecorder.net/ --customBehaviors \"git+https://github.com/webrecorder/browsertrix-crawler.git?branch=main&path=tests/custom-behaviors\" --scopeType page",
```

tw4l: Changed to use main to make it less brittle

ikreymer (approved):

PR #721, opened 2024-11-13
Ensure partial responses are not written

ikreymer:
various fixes for streaming, especially related to range requests
- follow up to #709
- fix: prefer streaming current response via takeStream, not only when size is unknown
- don't serialize async responses prematurely
- don't serialize 206 responses if there is size mismatch


ikreymer:
Merging for now so can do a 1.4.0-beta.0 release

PR #724, opened 2024-11-21
support removing range from query (via wabac.js 2.20.6):

ikreymer:
- fix for archiving facebook video, to match webrecorder/archiveweb.page#272
- permissions: auto enable permissions to avoid possibly modal (for both profiles and crawling)
- deps: update to latest wabac.js + warcio.js

### DIFF
--- a/src/util/browser.ts
+++ b/src/util/browser.ts
@@ -407,6 +407,21 @@ export class Browser {
     this.browser.on("disconnected", () => {
       this.browser = null;
     });
+
+    // common permissions
+    const permissions = [
+      "notifications",
+      "geolocation",
+      "camera",
+      "microphone",
+    ];
+
+    for (const name of permissions) {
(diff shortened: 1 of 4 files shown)


src/util/browser.ts
```diff
@@ -407,6 +407,21 @@ export class Browser {
...
+    // common permissions
+    const permissions = [
+      "notifications",
+      "geolocation",
+      "camera",
+      "microphone",
+    ];
+
+    for (const name of permissions) {
+      await this.firstCDP.send("Browser.setPermission", {
```
tw4l, line 420: nice addition!

tw4l (approved):

src/util/browser.ts
ikreymer: yes, i think this may help avoid some modal permission issues that arise on some sites.

Issue #665, opened 2024-08-06
Crawl button with javascript navigation


hamzamac:
Hi, 
we are try to crawl a site that use <button>s with Javascript based navigation instead of links of <a> tags. The JavaScripts code controlling the behavior of the buttons is hosted on a different domain (CDN) from that of the target site. 
When replaying the WACZ, everywhere we have buttons instead of links result into en error, it can not reach the link to JavaScript file.

How can we crawl such a website with Browsertrix-crawler?

tw4l:
Hi @hamzamac, would you be able to share the URL of the site you're trying to capture so I can take a look?

hamzamac:
Hi @tw4l, thank you for responding. The site is actually a SharePoint site with MFA. We manages to crawl it by creating a profile. but the links to folders appears to be spans.
![image](https://github.com/user-attachments/assets/3808e3d0-4cfb-41c9-88b0-4a21d5bfd943)

when when clicking the button on the replayweb.page it shows this error below
![image](https://github.com/user-attachments/assets/30ff0c5e-6928-43d0-97f1-a0efac0b824a)
 (the URL is pointing to is a public CDN URL which is accessible)
Do we need to include all the URI for JavaScripts in the seeds?

tw4l:
Hm, you shouldn't need to include the URIs for scripts - if the script is on the page, the crawler will discover it. This looks to me like it's more likely to be a bug in our replay engine than a missing script. It's hard to tell further without being able to reproduce it ourselves - would you be able to share a copy of the WACZ by email?

hamzamac:
Hi @tw4l, sure I will send the WACZ to the email on your profile.

hamzamac:
Hi @tw4l , can you please confirm if you have received the WACZ file? thanks.

Issue #216, opened 2023-02-02
Automatically Clicking on Elements


ttaomae:
I am trying to crawl some sites which require clicking on certain elements to load some content.

In one case, there is an `<img>` element with an `onclick` handler to open a page in a new window.
In another case, there is a `<form>` which performs a POST and includes an `onsubmit` handler.

#31 appears to describe a similar issue, but it looks like that was never resolved.

Is it possible to configure Browertrix to automatically click on certain elements? My understanding is that this is something that could theoretically be done with Browsertrix Behaviors, but I don't see a way to provide custom behaviors, other than modifying Browsertrix Behaviors and repackaging everything.

Rather than full custom behaviors, I think something like an "autoclick" behavior which allows the user to specify selectors to identify which elements to click would solve both the cases I described as well as the one described in #31.

Are either of these (custom behaviors or a customizable "autoclick" behavior) something that would be considered?

ikreymer:
Another option besides the behaviors, is that you have is to create a custom Puppeteer driver script, and pass that in via the `--driver`.
You can then use all of the API that is part of [Puppeteer](https://pptr.dev) to manipulate the page.
(Soon we'll be moving to playwright, but it has a comparable API).
Here is an example of more complex driver script:
https://github.com/webrecorder/dashboard-custom-drivers/blob/main/driver/az.js

As far as adding onclick handler support, it is tricky, because it is unpredictable what they may do.
It could open a new window, or navigate away to another page (less ideal, as now we're lost the original page).
There might be a clever way to do this in a generic way, but as first step, would be good to experiment with this via a custom driver script.

The driver script allows more control over the entire crawl process, while the behaviors are more for anything within a single site. We are still evaluating how to improve these, so any feedback definitely welcome if you have a chance to try this!

ttaomae:
Thank you for the pointer to the dashboard-custom-drivers repository. It was very helpful and I was able to quickly set up a driver that would click on the appropriate elements. Though as I mentioned, in my case, it opens a new window, which turned out to be more difficult to handle, but I did get it to work (see below).

> The driver script allows more control over the entire crawl process, while the behaviors are more for anything within a single site. We are still evaluating how to improve these, so any feedback definitely welcome if you have a chance to try this!

I guess I did not really understand what drivers were or how exactly they should be used, but it seems to provide the functionality I need. I no longer think custom behaviors or an "autoclick" behavior are necessary, so I'm going to close this issue.

[...]

Issue #31, opened 2021-02-28
Wordpress lightbox plugin content skipped


afward:
When trying to crawl a website using the [Wonder Lightbox](https://www.wonderplugin.com/wordpress-lightbox/) Wordpress plugin, content using the lightbox (marked with class wplightbox) are unfortunately skipped. 

Here are examples: https://www.wonderplugin.com/wordpress-lightbox/#examples.

![screenshot-2](https://user-images.githubusercontent.com/65351553/109418115-89592500-79be-11eb-97ad-78be0e549490.png)
![screenshot-1](https://user-images.githubusercontent.com/65351553/109418119-8a8a5200-79be-11eb-9dc3-9cd644c86bfc.png)

ikreymer:
Thanks! Yes, currently this would not be captured, but it seems like this could be addressed by adding a behavior that selects all `<a>` tags that have `class="wplightbox"` and clicking those. But, that's very specific to this particular plugin.

A more generic approach might be finding all anchor tags that have an onclick handler, although that's a bit more tricky.. Still, I wonder if that's a better approach for a generic behavior...

afward:
Similarly, it would be useful if custom behaviors could be setup for specific websites. Is this currently possible?

PR #729, opened 2024-12-03
Autoclick Support

### LINKED ISSUE #728: [Feature] Support for clicking on links / other elements.
This is an issue to track development of a new feature, also being tracked via webrecorder/browsertrix-behaviors#81, which would enable clicking on links (and possibly other elements as well).

This should have a potentially to fix a whole class of crawling issues, including #216, #665, #31 and more, where clicking on specific elements is needed.

The new `autoclick` behavior can be enabled optionally, and would use the autoclick behavior, which would default to `a` selector, but could be used with a custom selector when needed to click on other elements. The behavior would use onbeforeunload to prevent nav away from the page, and also track duplicates, (see behavior for more info).

The trade-off is slightly longer crawling time, as there'll be some false positives where clicking is not needed when this behavior is enabled.

ikreymer:
Adds support for autoclick behavior:
- Adds new `autoclick` behavior option to `--behaviors`, but not enabling by default
- Adds support for new exposed function `__bx_addSet` which allows autoclick behavior to persist state about links that have already been clicked to avoid duplicates, only used if link has an href
- Adds a new pageFinished flag on the worker state.
- Adds a on('dialog') handler to reject onbeforeunload page navigations, when in behavior (page not finished), but accept when page is finished - to allow navigation away only when behaviors are done
- Update to browsertrix-behaviors 0.7.0, which supports autoclick
- Add --clickSelector option to customize elements that will be clicked, defaulting to `a`.
- Add --linkSelector as alias for --selectLinks for consistency
- Unknown options for --behaviors printed as warnings, instead of hard exit, for forward compatibility for new behavior types in the future

Fixes #728, also #216, #665, #31


ikreymer:
Some examples of sites where the autoclick behavior is helpful:
- https://www.bs.ch/themen -- all links involve a distinct ajax call, in page clicking otherwise not working.
- https://globalchange.gov/our-work/highlights - clicking on page links results in ajax call, pages nav not working w/o
- https://www.wonderplugin.com/wordpress-lightbox/ - clicking on lightbox links results in custom behavior that is not crawled otherwise.

tw4l:
Thanks for the sites to test out! I can confirm that it's working in the ways you describe for each, but I'm also noticing a whole lot of these messages in the logs for all of the autoclick-enabled crawls:

```
{"timestamp":"2024-12-04T21:08:45.377Z","logLevel":"warn","context":"behavior","message":"Behavior run partially failed","details":{"reason":{"type":"exception","message":"Protocol error (Runtime.evaluate): Execution context was destroyed.","stack":"ProtocolError: Protocol error (Runtime.evaluate): Execution context was destroyed.\n    at <instance_members_initializer> (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:90:14)\n    at new Callback (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:94:16)\n    at CallbackRegistry.create (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:20:26)\n    at Connection._rawSend (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Connection.js:86:26)\n    at CdpCDPSession.send (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/CDPSession.js:63:33)\n    at Browser.evaluateWithCLI (file:///app/dist/util/browser.js:208:56)\n    at file:///app/dist/crawler.js:747:89\n    at Array.map (<anonymous>)\n    at Crawler.runBehaviors (file:///app/dist/crawler.js:747:61)\n    at Crawler.doPostLoadActions (file:///app/dist/crawler.js:669:49)"},"page":"https://www.bs.ch/themen","workerid":0}}
```

I haven't caught exactly why yet, but worth looking into.

For selectors I like the idea of a `--clickLinks` arg!

ikreymer:
> https://www.bs.ch/themen



> Thanks for the sites to test out! I can confirm that it's working in the ways you describe for each, but I'm also noticing a whole lot of these messages in the logs for all of the autoclick-enabled crawls:
> 
> ```
> {"timestamp":"2024-12-04T21:08:45.377Z","logLevel":"warn","context":"behavior","message":"Behavior run partially failed","details":{"reason":{"type":"exception","message":"Protocol error (Runtime.evaluate): Execution context was destroyed.","stack":"ProtocolError: Protocol error (Runtime.evaluate): Execution context was destroyed.\n    at <instance_members_initializer> (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:90:14)\n    at new Callback (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:94:16)\n    at CallbackRegistry.create (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:20:26)\n    at Connection._rawSend (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Connection.js:86:26)\n    at CdpCDPSession.send (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/CDPSession.js:63:33)\n    at Browser.evaluateWithCLI (file:///app/dist/util/browser.js:208:56)\n    at file:///app/dist/crawler.js:747:89\n    at Array.map (<anonymous>)\n    at Crawler.runBehaviors (file:///app/dist/crawler.js:747:61)\n    at Crawler.doPostLoadActions (file:///app/dist/crawler.js:669:49)"},"page":"https://www.bs.ch/themen","workerid":0}}
> ```
> 
> I haven't caught exactly why yet, but worth looking into.

Good catch! Turns out blocking the page unload was not actually working, and actually workeropts were being unnecessarily duplicated, see: https://github.com/webrecorder/browsertrix-crawler/pull/729/commits/8a4e4db43bdc0dd511691bf88be76e7963edffe2 - should be fixed now!

tw4l:
Thanks for that change! :) Not sure whether to actually approve this PR since we won't be merging this as-is with everything hardcoded and still need to add the selector flag, but based on testing I think this behavior is working as intended and good to add!

ikreymer:
Ready for final review, not enabling autoclick by default for now.

tw4l (approved):

PR #730, opened 2024-12-05
Add WARC resource containing DOM tree after load

magbb:
I use browsertrix for handling JavaScript heavy pages and extracting text from these. While the text extraction in Browsertrix definitely works, I need a fast and simple way to do boilerplate removal (like Trafilatura or jusText) on the DOM tree after rendering (i.e. I need the HTML tags).

Perhaps there is another way to do this already, but I implemented an option "dom" (similar to "text" and heavily inspired by that) which creates a WARC resource of DOM.getOuterHTML. In that way, I can simply run Trafilatura or jusText on these records while traversing the archive without having to replay the whole archive in a browser.

Does that make sense?

Thanks for creating browsertrix. It makes crawling so much easier!

### DIFF
--- a/src/.directory
+++ b/src/.directory
@@ -1,0 +1,6 @@
+[Dolphin]
+SortOrder=1
+SortRole=modificationtime
+Timestamp=2024,12,5,13,24,56.234
+Version=4
+ViewMode=1
--- a/src/.directory
+++ b/src/.directory
@@ -1,6 +1,0 @@
-[Dolphin]
-SortOrder=1
-SortRole=modificationtime
-Timestamp=2024,12,5,13,24,56.234
-Version=4
-ViewMode=1
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -20,6 +20,7 @@ import { WACZ, WACZInitOpts, mergeCDXJ } from "./util/wacz.js";
 
 import { HealthChecker } from "./util/healthcheck.js";
 import { TextExtractViaSnapshot } from "./util/textextract.js";
+import { DomExtractViaSnapshot } from "./util/domextract.js";
 import {
   initStorage,
   getFileSize,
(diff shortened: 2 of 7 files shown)


ikreymer:
Hi, apologies for late response - I've been meaning to comment, but didn't have a chance. Actually started a similar approach for DOM snapshots on this branch:
https://github.com/webrecorder/browsertrix-crawler/tree/dom-snapshot
I think the approach I was taking was to use the existing `DOMSnapshot.captureSnapshot` that we use for text, and also use it to serialize to html...
https://github.com/webrecorder/browsertrix-crawler/compare/main...dom-snapshot
The reason for this approach was to have more flexibility: for example, probably want to string out <script> tags and potentially be able to capture shadow DOM, which https://chromedevtools.github.io/devtools-protocol/tot/DOMSnapshot/#method-captureSnapshot should include.
Otherwise, the approach is about the same, I believe!
Do you think the approach I'd started would make sense for your use case?
Perhaps its worth comparing the results a bit more...

magbb:
Hi Ilya! Thanks for your feedback! Yes, that approach makes sense for my case. Would be great to have this functionality in Browsertrix!

wvengen:
I've been interested in this feature several times. My needs for this were:
- (usually) extract data from the archives: the DOM is what the user sees (in contrast to e.g. data from JSON responses)
- (sometimes) pages using Javascript heavily, making ReplayWeb.page really show (just showing the DOM could be faster)
- (rarely) pages using complex Javascript to probably block bots, loading JS files based on changing things (time), breaking replay in unexpected ways

I'd tend to think that how a stored DOM is replayed, may best be handled by ReplayWeb.page, instead of changing it at crawl time. That is more 'pure' in representing the state at crawl. If storing it changes over time, the exact DOM details would become dependent on crawler version. And if handling is improved, having that logic in ReplayWeb.page would make older archives benefit of these improvements as well.
In any case, using `DOMSnapshot.captureSnapshot` sounds like a good idea, e.g. for shadow DOM later as mentioned.

PR #736, opened 2024-12-31
separate fetch api for autofetch bbehavior + additional improvements on partial responses:

### LINKED ISSUE #735: Autofetch behavior results in empty 200 responses
Not sure how `autofetch` behavior is intended to work, but on this page, it retreives empty 200 responses with no file for the two downloadable docs at the bottom of the page:
https://hdsr.mitpress.mit.edu/pub/9c034nf7

ikreymer:
Chromium now interrupts fetch() if abort() is called or page is navigated, so autofetch behavior using native fetch() is less than ideal. This PR adds support for __bx_fetch() command for autofetch behavior (supported in browsertrix-behaviors 0.6.6) to fetch separately from browser's reguar fetch()
- __bx_fetch() starts a fetch, but does not return content to browser, doesn't need abort(), unaffected by page navigation, but will still try to use browser network stack when possible, making it more efficient for background fetching.
- if network stack fetch fails, fallback to regular node fetch() in the crawler.
Additional improvements for interrupted fetch:
- don't store truncated media responses, even for 200
- avoid doing duplicate async fetching if response already handled (eg. fetch handled in multiple contexts)
- fixes #735, where fetch was interrupted, resulted in an empty response

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,7 +61,7 @@ jobs:
         run: yarn add -D http-server
 
       - name: install py-wacz as root for tests
-        run: sudo pip install wacz
+        run: sudo pip install wacz --ignore-installed
 
       - name: run all tests as root
         run: sudo DOCKER_HOST_NAME=172.17.0.1 CI=true yarn test -validate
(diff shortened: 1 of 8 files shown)


tw4l (approved):
Looks good and tested that the fix is working as expected against sample URL in the issue. Nice work!

PR #742, opened 2025-01-18
set failed URL retry to 5 by default

ikreymer:
Just in case, for larger crawls where sometimes multiple retries may be needed.

Todo:
- make configurable?
- ensure doesn't add to done count


ikreymer:
Closing in favor of #743

PR #743, opened 2025-01-18
Retry support and additional fixes

ikreymer:
- retries: for failed pages, set retry to 5 in cases multiple retries may be needed.
- redirect: if page url is /path/ -> /path, don't add as extra seed
- proxy: don't use global dispatcher, pass dispatcher explicitly when using proxy, as proxy may interfere with local network requests
- final exit flag: if crawl is done and also interrupted, ensure WACZ is still written/uploaded by setting final exit to true
- hashtag only change force reload: if loading page with same URL but different hashtag, eg. `https://example.com/#B` after `https://example.com/#A`, do a full reload

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -700,6 +700,9 @@ class ArgParser {
 
     // background behaviors to apply
     const behaviorOpts: { [key: string]: string | boolean } = {};
+    if (argv.blockAds) {
+      argv.behaviors.push("autoclick");
+    }
     if (argv.behaviors.length > 0) {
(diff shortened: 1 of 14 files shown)


src/util/argParser.ts
```diff
@@ -700,6 +700,9 @@ class ArgParser {
 
     // background behaviors to apply
     const behaviorOpts: { [key: string]: string | boolean } = {};
+    if (argv.blockAds) {
+      argv.behaviors.push("autoclick");
+    }
```
tw4l, line 705: Just leaving a comment so we remember to undo this

PR #748, opened 2025-01-29
if uploading wacz files, compute waczfile name on load to be able to …

### LINKED ISSUE #747: Set WACZ URL ahead of time so it can be known when pages are written.
For Browsertrix integration and `--writePagesToRedis`, it would be good to have the WACZ URL computed ahead of time so it can be included in Redis, so that we can track with pages are in which WACZ files.

Mostly need to just compute the filename at crawl start time.

ikreymer:
…store filename along with page data:

- set filename on crawler load, if not already set, otherwise use existing
- store filename per crawler instance in <crawlid>:nextWacz
- add 'filename' field to page when writing pages to redis
- clear wacz filename when wacz is uploaded to set a new one
- fixes #747

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -244,7 +244,7 @@ export class Crawler {
     // pages file
     this.pagesFH = null;
 
-    this.crawlId = process.env.CRAWL_ID || os.hostname();
+    this.crawlId = this.params.crawlId;
 
     this.startTime = Date.now();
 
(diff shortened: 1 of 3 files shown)


ikreymer:
Tested this with Browsertrix, but really should add a self-contained test using minio probably.

tw4l (approved):
Would be good to add the test, otherwise LGTM

PR #757, opened 2025-02-06
Retry same queue

### LINKED ISSUE #132: [Proposal] Basic retry strategy for crawling errors
As far as I've understood, when a page fetch fails, that page will be skipped and won't be included in the final result, whatever the error was (even if it's just a temporary error like a 503).
Since this is very suboptimal, I'd propose a basic retry strategy that works like the following:
- when the crawling of a page fails, re-enqueue it so that it gets retried after some time
- there is a limited number of enqueue-and-retry attempts for each failed page, settable with an option (maybe `--max-retries`?) that has a reasonable default value (e.g. 2)

Furthermore, in case of network connection errors the crawler should ideally remain on the current page and retry until the connection comes back.
What do you think?

ikreymer:
- follow up to #743
- page retries are simply added back to the same queue with `retry` param incremented and a higher scope, after extraHops, to ensure retries are added at the end.
- score calculation is: `score = depth + (extraHops * MAX_DEPTH) + (retry * MAX_DEPTH * 2)`, this ensures that retries have lower priority than extraHops, and additional retries even lower priority (higher score).
- warning is logged when a retry happens, error only when all retries are exhausted.
- back to one failure list, urls added there only when all retries are exhausted.
- rename --numRetries -> --maxRetries / --retries for clarity
- state load: allow retrying previously failed URLs if --maxRetries is higher then on previous run.
- ensure working with --failOnFailedStatus, if provided, invalid status codes (>= 400) are retried along with page load failures
- fixes #132


src/util/state.ts
```diff
@@ -288,42 +284,30 @@ end
...
     redis.call('zadd', KEYS[2], score, json);
-    return data['retry'];
+    return retry;
   else
     redis.call('lpush', KEYS[3], json);
-    return 0;
   end
 end
-return -1;
+return -1
```
ikreymer, line 309:
```suggestion
return -1;
```

src/crawler.ts
```diff
@@ -1982,12 +1980,32 @@ self.__bx_behaviors.selectMainBehavior();
...
             data.pageSkipped = true;
             logger.warn("Page Load Blocked, skipping", { msg, loadState });
+          } else if (retry < this.params.maxPageRetries) {
+            logger.warn(
+              "Page Load Failed, will retry",
+              {
+                msg,
+                url,
+                retry,
+                maxRetries: this.params.numRetries,
```
tw4l, line 1990:
```suggestion
                maxRetries: this.params.maxPageRetries,
```

src/crawler.ts
```diff
@@ -1982,12 +1980,32 @@ self.__bx_behaviors.selectMainBehavior();
...
-              loadState,
-              ...logDetails,
-            });
+            logger.error(
+              "Page Load Failed, no retries left",
+              {
+                msg,
+                url,
+                retry,
+                maxRetries: this.params.numRetries,
```
tw4l, line 2003:
```suggestion
                maxRetries: this.params.maxPageRetries,
```

PR #761, opened 2025-02-09
Fix exit code and stats

ikreymer:
Follow up to #753, ensure that:
- Use new 'FailCrawl' exit code explicitly when crawl should fail due to fatal(), not just restart, taking into account
- FailCrawl status will resolve to actual status set via setDefaultFatalExitCode(), which may be 0 for restartOnError on FatalError
- Browser crashes do not result in a failed crawl!
- Add additional exit codes: UploadError, RedisMissing, remove GenericError
- stats log: include failed stat in 'total', also include 'totalSeen'

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "browsertrix-crawler",
-  "version": "1.5.1",
+  "version": "1.5.2",
   "main": "browsertrix-crawler",
   "type": "module",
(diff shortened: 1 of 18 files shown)


ikreymer:
Adding a simpler fix via #762 first to address urgent issues, this adds further cleanup.

ikreymer:
Issue was something else, closing this for now.

PR #764, opened 2025-02-10
Add more exit codes to detect interruption reason

### LINKED ISSUE #584: Better indicate the interruption reason

benoit74:
Fix #584

Following recent changes in handling of exit code, and as suggested, I propose to refine exit codes a bit further, especially to be better informed of situations where we would like to indicate something to the end user in Kiwix usage.

I replaced `interrupted` and `browserCrashed` properties of the crawler with a single `interrupt_reason` property, whose value is an enum indicating (when possible) the interruption reason.

I slightly changed the message on cancellation since there was now a repetition of "gracefully finishing current pages".

These changes made me realize that there was maybe a flaw in the `serializeAndExit` function since according to my code analysis the `interrupted` property is never set when this function is called, unless I miss something.

Tested locally and works as expected, some cases (reasonably easily feasible ones) covered by automated tests.

There is finally a small fix of documentation.


benoit74:
@tw4l @ikreymer this is ready for review

src/util/constants.ts
```diff
@@ -63,8 +63,21 @@ export enum ExitCodes {
...
   OutOfSpace = 3,
   BrowserCrashed = 10,
-  InterruptedGraceful = 11,
-  InterruptedImmediate = 13,
+  Cancelled = 11,
+  FailedLimit = 12,
+  Interrupted = 13,
+  SizeLimit = 14,
+  TimeLimit = 15,
+  DiskUtilization = 16,
```
tw4l, line 71: I think this may be a duplicate of exit code 3 (OutOfSpace)

ikreymer: I think they're slightly different -- one is projected DiskUtilization based on the --diskUtilization, the other is crawler has actually run out of space already.

tw4l: Ah, good point

tw4l:
Nice to see this improvement! It would be great to document these exit codes in the crawler documentation so that folks don't have to dig into the code to understand what the codes mean.

tw4l:
> Definitely, will tackle this as well. Probably in a distinct PR to move this code change asap to release, should documentation take more discussions than expected ... Where should this be placed? Should we create a new `Exit codes` page in the [User Guide](https://crawler.docs.browsertrix.com/user-guide/), right after the `Outputs` page? Or as a section of this [Outputs](https://crawler.docs.browsertrix.com/user-guide/outputs/) page?

Thank you! Much appreciated :) I think a new Exit Codes section would be the simplest/most easily discovered way to document this.

tw4l (approved):
Looks good, thanks for the contribution!

ikreymer:
Yep, looks good! I merged some changes that readded browser.crashed, which is also used by healthchecker.
Tested with crawler that was getting stuck per #763, and seems to be good now.

Can add docs in a follow-up, this should be good to go!

src/main.ts
```diff
@@ -29,9 +29,9 @@ async function handleTerminate(signame: string) {
   try {
     await crawler.checkCanceled();
 
-    if (!crawler.interrupted) {
-      logger.info("SIGNAL: gracefully finishing current pages...");
-      crawler.gracefulFinishOnInterrupt();
+    if (!crawler.interruptReason) {
+      logger.info("SIGNAL: canceled received...");
```
ikreymer, line 33:
```suggestion
      logger.info("SIGNAL: interrupt request received, finishing current pages before exit...");
```

src/main.ts
```diff
@@ -29,9 +29,9 @@ async function handleTerminate(signame: string) {
   try {
     await crawler.checkCanceled();
 
-    if (!crawler.interrupted) {
-      logger.info("SIGNAL: gracefully finishing current pages...");
-      crawler.gracefulFinishOnInterrupt();
+    if (!crawler.interruptReason) {
+      logger.info("SIGNAL: interrupt request received, finishing current pages before exit...");
+      crawler.gracefulFinishOnInterrupt(InterruptReason.Cancelled);
```
ikreymer, line 34: Actually, this isn't quite correct - this isn't necessarily cancelled.
When we run in k8s, interrupt does not mean to cancel the crawl, it means to stop and possibly restart the crawler..
The signal could come from user or from k8s system (such as memory pressure, etc...)

ikreymer:
Ah but there is some confusion but cancellation / interruption.
The way it works is: first interrupt signal results in waiting for crawler to finish pages, second interrupt signal results in immediate termination.

However, this does not mean the crawl is cancelled, esp. in k8s, but that the crawler needs to be restarted (for whatever reason). It could mean that its being moved to a different node, etc.. This is very important that the deletion of pod does not actually cancel the crawl automatically. The 'cancel' state is only entered through a special message sent via redis.
The exit codes should differentiate between the graceful and immediate interrupt (though on k8s, it may only second one anyway).

This is slightly more confusing because need to support different environments, both k8s and plain docker/podman

ikreymer (approved):

PR #765, opened 2025-02-11
Add documentation on exit codes

benoit74:
@tw4l @ikreymer ready for your review

### DIFF
--- a/docs/docs/user-guide/exit-codes.md
+++ b/docs/docs/user-guide/exit-codes.md
@@ -5,13 +5,13 @@ The crawler uses following exit codes to indicate crawl result.
 | Code | Name | Description |
 |--|--|--|
 | 0 | Success | Crawl completed normally |
-| 1 | GenericError | SIGINT or SIGTERM signal forced crawler to stop |
+| 1 | GenericError | Unspecified error, check logs for more details |
 | 3 | OutOfSpace | Disk is already full |
-| 9 | Failed | Something bad happened during the crawl, might be worth to retry it |
+| 9 | Failed | Crawl failed unexpectedly, might be worth retrying |
 | 10 | BrowserCrashed | Browser used to fetch pages has crashed |
-| 11 | SignalInterrupted | Cancellation has been requested |
-| 12 | FailedLimit | Limit on amount of failed pages, configured with `--limit`, has been reached |
-| 13 | SignalInterruptedForce | Cancellation has been forced |
+| 11 | SignalInterrupted | Crawl stopped gracefully in response to SIGINT signal |
+| 12 | FailedLimit | Limit on amount of failed pages, configured with `--failOnFailedLimit`, has been reached |
+| 13 | SignalInterruptedForce | Crawl stopped forcefully in response to SIGTERM or repeated SIGINT signal |
 | 14 | SizeLimit | Limit on maximum WARC size, configured with `--sizeLimit`, has been reached |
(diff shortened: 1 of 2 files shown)


docs/docs/user-guide/exit-codes.md
```diff
@@ -0,0 +1,19 @@
+# Exit codes
+
+The crawler uses following exit codes to indicate crawl result.
+
+| Code | Name | Description |
+|--|--|--|
+| 0 | Success | Crawl completed normally |
+| 1 | GenericError | SIGINT or SIGTERM signal forced crawler to stop |
```
tw4l, line 8:
```suggestion
| 1 | GenericError | Unspecified error, check logs for more details |
```

In `main.ts`, though it hapens in response to a SIGINT/SIGTERM, we only return this exit code instead of `SignalInterrupted` or `SingalInterruptedForce` if the crawler isn't running or doesn't have any state, indicating some sort of underlying problem.

There's also a case in `create-login-profile.ts` where exit code 1 gets returned for an entirely unrelated reason (login form can't be found for automated login), though it looks like we missed the `ExitCodes` enum there.

So for now, I think keeping this message more generic makes sense.

Probably the uses of this exit code in `create-login-profile.ts` would probably be better switched over to `SignalInterrupted` or `SignalInterruptedForce`, where this is returned directly as result of getting a SIGINT or SIGTERM signal, but I can do that separately in a follow-up.

docs/docs/user-guide/exit-codes.md
```diff
@@ -0,0 +1,19 @@
...
+| 3 | OutOfSpace | Disk is already full |
+| 9 | Failed | Something bad happened during the crawl, might be worth to retry it |
+| 10 | BrowserCrashed | Browser used to fetch pages has crashed |
+| 11 | SignalInterrupted | Cancellation has been requested |
```
tw4l, line 12:
```suggestion
| 11 | SignalInterrupted | Crawl stopped gracefully in response to SIGINT signal |
```

docs/docs/user-guide/exit-codes.md
```diff
@@ -0,0 +1,19 @@
...
+| 12 | FailedLimit | Limit on amount of failed pages, configured with `--limit`, has been reached |
+| 13 | SignalInterruptedForce | Cancellation has been forced |
```
tw4l, line 14:
```suggestion
| 13 | SignalInterruptedForce | Crawl stopped forcefully in response to SIGTERM or repeated SIGINT signal |
```

docs/docs/user-guide/exit-codes.md
```diff
@@ -0,0 +1,19 @@
...
+
+| Code | Name | Description |
+|--|--|--|
+| 0 | Success | Crawl completed normally |
+| 1 | GenericError | SIGINT or SIGTERM signal forced crawler to stop |
+| 3 | OutOfSpace | Disk is already full |
+| 9 | Failed | Something bad happened during the crawl, might be worth to retry it |
+| 10 | BrowserCrashed | Browser used to fetch pages has crashed |
+| 11 | SignalInterrupted | Cancellation has been requested |
+| 12 | FailedLimit | Limit on amount of failed pages, configured with `--limit`, has been reached |
```
tw4l, line 13:
```suggestion

| 12 | FailedLimit | Limit on amount of failed pages, configured with `--failOnFailedLimit`, has been reached |
```

tw4l (approved):
Thank you!

PR #767, opened 2025-02-11
Make sure all exit calls use ExitCodes enum

tw4l:
Quick follow-up to #584 to make sure enum is used everywhere.

I debated modifying `create-login-profile.ts`'s handleTerminate method to return `ExitCodes.SignalInterrupted` or `ExitCodes.SignalInterruptedForce` depending on the signal received, but I'm not sure that's actually helpful in the case of profiles and the behavior doesn't match exactly what's documented for the crawler, so left that as-is as a `ExitCodes.GenericError`.

### DIFF
--- a/src/create-login-profile.ts
+++ b/src/create-login-profile.ts
@@ -147,7 +147,7 @@ function getDefaultWindowSize() {
 
 function handleTerminate(signame: string) {
   logger.info(`Got signal ${signame}, exiting`);
-  process.exit(ExitCodes.GenericError);
+  process.exit(ExitCodes.SignalInterrupted);
 }
 
 async function main() {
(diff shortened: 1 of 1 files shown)


ikreymer:
> I debated modifying `create-login-profile.ts`'s handleTerminate method to return `ExitCodes.SignalInterrupted` or `ExitCodes.SignalInterruptedForce` depending on the signal received, but I'm not sure that's actually helpful in the case of profiles and the behavior doesn't match exactly what's documented for the crawler, so left that as-is as a `ExitCodes.GenericError`.

Yeah, I could go either way on that, since its not interrupting a crawl, it doesn't make too much of a difference, could set it to that to indicate interrupt vs some other random error..

tw4l:
> Edit: I guess ExitCodes.SignalInterrupted might make most sense, since it quickly exists on first signal.

Makes sense to me, went ahead and made that change.

PR #771, opened 2025-02-14
Add clarification in usage about hostname used

benoit74:
@tw4l @ikreymer this is ready for review, mostly minor change but we've had the question so better to clarify for everyone

### DIFF
--- a/docs/docs/user-guide/cli-options.md
+++ b/docs/docs/user-guide/cli-options.md
@@ -18,7 +18,7 @@ Options:
       --crawlId, --id                       A user provided ID for this crawl or
                                              crawl configuration (can also be se
                                             t via CRAWL_ID env var, defaults to
-                                            hostname)                   [string]
+                                            Docker container hostname)  [string]
       --waitUntil                           Puppeteer page.goto() condition to w
                                             ait for before continuing, can be mu
                                             ltiple separated by ','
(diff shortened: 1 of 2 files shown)


ikreymer:
This is actually the container name, not the machine hostname! We don't have a way to access the machine hostname, just using os.hostname() which is the hostname of the container.

benoit74:
Indeed, I clarified it a bit further

docs/docs/user-guide/cli-options.md
```diff
@@ -18,7 +18,7 @@ Options:
       --crawlId, --id                       A user provided ID for this crawl or
                                              crawl configuration (can also be se
                                             t via CRAWL_ID env var, defaults to
-                                            hostname)                   [string]
+                                            Docker container hostname)  [string]
```
ikreymer, line 21:
```suggestion
                                           container hostname)  [string]
```

src/util/argParser.ts
```diff
@@ -85,7 +85,7 @@ class ArgParser {
         crawlId: {
           alias: "id",
           describe:
-            "A user provided ID for this crawl or crawl configuration (can also be set via CRAWL_ID env var, defaults to hostname)",
+            "A user provided ID for this crawl or crawl configuration (can also be set via CRAWL_ID env var, defaults to Docker container hostname)",
```
ikreymer, line 88:
```suggestion
            "A user provided ID for this crawl or crawl configuration (can also be set via CRAWL_ID env var, defaults to the container hostname)",
```

PR #775, opened 2025-02-21
instead of puppeteer.exposeFunction(), use cdp function bindings directly to avoid issues custom toJSON overrides:

### LINKED ISSUE #770: "Invalid Page - not a valid URL" on particular website
Hello, I'm getting 

```
│ {"timestamp":"2025-02-13T10:32:17.911Z","logLevel":"warn","context":"general","message":"Invalid Page - not a valid URL","details":{"url":"[","page":"https://www.exlayer.jp/","workerid":0}}
```

when trying to crawl `https://www.exlayer.jp`. Other sites work fine. Any ideas?
The crawler somehow thinks there is an  URL `[` on the page.

ikreymer:
- add Runtime.addBinding for each exposed function, handle in one place with Runtime.bindingCalled
- convert binding names to BxFunctionBindings enum
- update to browsertrix-behaviors 0.7.1 to avoid waiting for return value
- fixes #770

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "dependencies": {
     "@novnc/novnc": "1.4.0",
     "@webrecorder/wabac": "^2.20.8",
-    "browsertrix-behaviors": "^0.7.0",
+    "browsertrix-behaviors": "^0.7.1",
     "client-zip": "^2.4.5",
     "css-selector-parser": "^3.0.5",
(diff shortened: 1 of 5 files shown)


ikreymer:
Closing in favor of #803 - was able to keep using puppeteer.exposeFunction() as it provides support for returning values, implemented in behaviors 0.8.0. Enums from this PR also added to #803.

Issue #766, opened 2025-02-11
Unable to load new page, browser needs restart


zlodejpapiru:
Hi, since the 1.5 version crawler keep crashing. Save state is saved, warcs not generated. Here is excerpt form log (1.5.3) version + my config. 

`workers: 2
limit: 200000
generateWACZ: true
behaviors:
 - autoscroll
 - siteSpecific
 - autoplay
 - autofetch
#behaviorTimeout: 0
#timeout: 36000
saveState: always
warcinfo: 
  operator: xxxx
generateCdx: true
#pageLoadTimeout: 90
scopeType: page-spa
extraHops: 50000
#waitUntil: true
#sitemap: true ` 

`{"timestamp":"2025-02-11T10:44:37.364Z","logLevel":"warn","context":"recorder","message":"Skipping URL from unknown frame","details":{"url":"https://twitter.com/zenisek_m","frameId":"EAD6CFF294DEABBEC9EA03D736541A43"}}
{"timestamp":"2025-02-11T10:44:37.594Z","logLevel":"warn","context":"recorder","message":"Skipping URL from unknown frame","details":{"url":"https://x.com/zenisek_m","frameId":"EAD6CFF294DEABBEC9EA03D736541A43"}}
{"timestamp":"2025-02-11T10:44:37.899Z","logLevel":"warn","context":"recorder","message":"Skipping URL from unknown frame","details":{"url":"https://abs.twimg.com/responsive-web/client-web/vendor.c4b9145a.js","frameId":"EAD6CFF294DEABBEC9EA03D736541A43"}}
{"timestamp":"2025-02-11T10:44:37.903Z","logLevel":"warn","context":"recorder","message":"Skipping URL from unknown frame","details":{"url":"https://abs.twimg.com/responsive-web/client-web/i18n/en.08723eda.js","frameId":"EAD6CFF294DEABBEC9EA03D736541A43"}}
{"timestamp":"2025-02-11T10:44:37.905Z","logLevel":"warn","context":"recorder","message":"Skipping URL from unknown frame","details":{"url":"https://abs.twimg.com/responsive-web/client-web/main.2b2a6b7a.js","frameId":"EAD6CFF294DEABBEC9EA03D736541A43"}}
[...]

PR #779, opened 2025-02-26
further fix to stuck on getting new window:

ikreymer:
- set retries back to 3, was set high by mistake
- if will restart, throw exception to restart crawler
- otherwise, attempt to kill browser process that is stalled (appears to work in testing)
- follow-up to #766

### DIFF
--- a/src/util/worker.ts
+++ b/src/util/worker.ts
@@ -231,7 +231,7 @@ export class PageWorker {
           break;
         }
 
-        if (retry >= 3) {
+        if (retry >= 2) {
           this.crawler.markBrowserCrashed();
           if (this.crawler.params.restartsOnError) {
             throw new Error("Unable to load new page, browser needs restart");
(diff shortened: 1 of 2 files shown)


src/util/worker.ts
```diff
@@ -231,9 +231,13 @@ export class PageWorker {
           break;
         }
 
-        if (retry >= 1000000000) {
+        if (retry >= 3) {
```
tw4l, line 234: I wonder if we even want to set this lower, like 2 by default? Can do that in separate issue/PR though if we want.

tw4l (approved):
Thank you!

src/util/worker.ts
ikreymer: Yeah, maybe it should be 2 really, since this generally shouldn't be timing out..

ikreymer, line 234:
```suggestion
        if (retry >= 2) {
```

PR #783, opened 2025-02-28
Fix using cached WACZ filename if already set ahead of time.

### LINKED ISSUE #747: Set WACZ URL ahead of time so it can be known when pages are written.

ikreymer:
- if <uid>:nextWacz filename already exists, actually get it and use that!
- don't merge cdx if not generating wacz yet, use same condition for both bump version to 1.5.8
- fix follow-up to #748, fix #747

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "browsertrix-crawler",
-  "version": "1.5.7",
+  "version": "1.5.8",
   "main": "browsertrix-crawler",
   "type": "module",
   "repository": "https://github.com/webrecorder/browsertrix-crawler",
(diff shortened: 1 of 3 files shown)


ikreymer:
Should do some manual testing, esp. for crawls where pod fails / is delete mid crawl to ensure filename on pages matches the actual filenames now..

ikreymer:
Did some testing on dev, appears to be working as expected - same filename is kept when a pod is deleted / restarted.

PR #799, opened 2025-03-26
Fail crawl with fatal message if custom behavior isn't loaded

### LINKED ISSUE #797: Fail crawl if custom behavior cannot be fetched
Currently, the crawler will log a warning and continue without the custom behavior if a custom behavior cannot be loaded from the specified URL or Git repository. This can result in unexpected crawl results. Instead, the crawler should log an error and exit if a custom behavior cannot be loaded.

We should do something similar for custom behaviors specified via a local filepath, namely fail if if no custom behaviors are collected or the specified filepath doesn't exist.

tw4l:
Fixes #797 

The crawler will now exit with a fatal log message and exit code 17 if:

- A Git repository specified with `--customBehavior` cannot be cloned successfully (new)
- A custom behavior file at a URL specified with `--customBehavior` is not fetched successfully (new)
- No custom behaviors are collected at a local filepath specified with `--customBehavior`, or if an error is thrown while attempting to collect files from a nonexistent path (new)
- Any custom behaviors collected fail `Browser.checkScript` validation (existing behavior)

Tests have also been added accordingly.

### DIFF
--- a/src/util/file_reader.ts
+++ b/src/util/file_reader.ts
@@ -153,6 +153,14 @@ async function collectLocalPathBehaviors(
     }
   }
 
+  if (!behaviors && depth === 0) {
+    logger.fatal(
+      "No custom behaviors found at specified path",
+      { path: resolvedPath },
+      "behavior",
+    );
+  }
+
   return behaviors;
 }
 
--- a/src/util/file_reader.ts
+++ b/src/util/file_reader.ts
@@ -120,37 +120,48 @@ async function collectLocalPathBehaviors(
     return [];
   }
 
-  const stat = await fsp.stat(resolvedPath);
-
-  if (stat.isFile() && ALLOWED_EXTS.includes(path.extname(resolvedPath))) {
-    const contents = await fsp.readFile(resolvedPath);
-    return [
-      {
-        path: resolvedPath,
-        contents: `/* src: ${resolvedPath} */\n\n${contents}`,
-      },
-    ];
-  }
-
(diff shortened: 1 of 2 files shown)


src/util/file_reader.ts
```diff
@@ -120,37 +120,56 @@ async function collectLocalPathBehaviors(
...
   }
 
-  if (isDir) {
-    const files = await fsp.readdir(resolvedPath);
-    for (const file of files) {
-      const filePath = path.join(resolvedPath, file);
-      const newBehaviors = await collectLocalPathBehaviors(filePath, depth + 1);
-      behaviors.push(...newBehaviors);
-    }
+  if (!behaviors && depth === 0) {
```
ikreymer, line 167: Hm, does this mean that specifying an empty directory will lead to an error? I think that could be allowed, as long as directory exists?

tw4l: (Pointing out Git repo here because we first clone the git repo and then collect behaviors from the cloned repo via this codepath)

ikreymer: Yeah, just thinking if maybe should allow having a placeholder directory that things will be added to later.. I guess could always add a sample behavior or something like that..

tw4l: Yeah if we go that way I'd suggest we do add a sample behavior. Otherwise we might be setting people up for silent errors?

tw4l: I'll leave it up to your judgment though

PR #805, opened 2025-03-31
Add option to push behavior + behavior script logs to Redis

### LINKED ISSUE #804: Push behavior script logs to Redis
Similar to how we currently push error and fatal log messages to Redis if the `--logErrorsToRedis` flag is provided, we should provide a means for the crawler to push behavior script logs to Redis so that they can be picked up and displayed in the Browsertrix UI.

We may want to limit which log lines are added to Redis, e.g. by only pushing site-specific behavior messages and/or excluding debug messages. Details TBD.

It may be useful to also push logs from the `behavior` context so that Browsertrix logs include which custom behaviors have been loaded and relevant messages about their execution.

tw4l:
Fixes #804 

Noisy logs from built-in behaviors like autoscroll are now logged to debug in https://github.com/webrecorder/browsertrix-behaviors/pull/92 and so won't be pushed to Redis for newer versions of the crawler.

Also updates browsertrix-behaviors to 0.8.3 and makes some changes to log format in tests accordingly.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -407,7 +407,7 @@ export class Crawler {
       logger.setLogBehaviorsToRedis(true);
     }
 
-    if (this.param.logErrorsToRedis || this.params.logBehaviorsToRedis) {
+    if (this.params.logErrorsToRedis || this.params.logBehaviorsToRedis) {
       logger.setCrawlState(this.crawlState);
     }
 
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1502,7 +1502,6 @@ self.__bx_behaviors.selectMainBehavior();
   async setStatusAndExit(exitCode: ExitCodes, status: string) {
     logger.info(`Exiting, Crawl status: ${status}`);
 
-
     await this.closeLog();
 
     if (this.crawlState && status) {
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -32,7 +32,7 @@ import {
 import { ScreenCaster, WSTransport } from "./util/screencaster.js";
 import { Screenshots } from "./util/screenshots.js";
 import { initRedis } from "./util/redis.js";
-import { logger, formatErr, LogDetails } from "./util/logger.js";
+import { logger, formatErr, LogDetails, LogContext } from "./util/logger.js";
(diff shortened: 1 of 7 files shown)


src/crawler.ts
```diff
@@ -1470,6 +1501,10 @@ self.__bx_behaviors.selectMainBehavior();
   async setStatusAndExit(exitCode: ExitCodes, status: string) {
     logger.info(`Exiting, Crawl status: ${status}`);
 
+    // TODO: Remove before merging, just for testing
+    const behaviorsLoggedToRedis = await this.crawlState.getBehaviorList();
+    logger.info("Behavior lines logged to Redis", { behaviorsLoggedToRedis });
```
tw4l, line 1506:
```suggestion
```

src/util/state.ts
```diff
@@ -875,6 +878,11 @@ return inx;
     return await this.redis.lrange(this.ekey, 0, -1);
   }
 
+  // TODO: Remove before merging, just for testing
+  async getBehaviorList() {
+    return await this.redis.lrange(this.bkey, 0, -1);
+  }
+
```
tw4l, line 885:
```suggestion
```

tw4l:
@ikreymer This should be ready to merge now if you want to take a last look

yarn.lock
```diff
@@ -1460,10 +1460,10 @@ browserslist@^4.24.0:
     node-releases "^2.0.18"
     update-browserslist-db "^1.1.1"
 
-browsertrix-behaviors@^0.8.1:
-  version "0.8.1"
-  resolved "https://registry.yarnpkg.com/browsertrix-behaviors/-/browsertrix-behaviors-0.8.1.tgz#21d2726d32ca23f94efd2c3e9a0b8b98c42482bb"
-  integrity sha512-bG5KA/9fF60slakycvbJZ4COo/nD9GKflNEhEk3yQQTy3C0ikRVOpueG9XfvAgq2r4Ji1NMY9pesJRpQqtHKXg==
+browsertrix-behaviors@^0.8.2:
```
ikreymer, line 1463: This is out of date, should be pinned to 0.8.3, can fix.

Issue #784, opened 2025-03-03
Running `crawl` command for the second time doesn't crawl site


dltj:
Atop the Browsertrix docker image, I'm adding another layer with some Python code with the intention of running the resulting container as an AWS Lambda. My use case is fairly specific...I want to get a WARC (and screenshots) of only one web page per `crawl` invocation. The Lambda can receive an SQS message with several single-page URLs to capture, so I'm having a subprocess call `crawl` for each URL. The first crawl works fine. What I'm getting instead for the second crawl is no WARC file and no errors/warnings in the output. 

Setting aside my Python Lambda handler code for the moment, I'm seeing the problem even when I run back-to-back `crawl` commands in the container's shell:

```bash
❯ docker run --rm -it --entrypoint /bin/bash browsertrix_lambda
root@76660a07cde6:/var/task# crawl  --url "https://dltj.org"  --collection "202503021403-dltj-homepage"  --limit 1   --timeout 30  --diskUtilization 0   --screenshot   view,fullPage
{"timestamp":"2025-03-03T03:59:12.859Z","logLevel":"info","context":"general","message":"Browsertrix-Crawler 1.5.8 (with warcio.js 2.4.3)","details":{}}
{"timestamp":"2025-03-03T03:59:12.861Z","logLevel":"info","context":"general","message":"Seeds","details":[{"url":"https://dltj.org/","scopeType":"prefix","include":["/^https?:\\/\\/dltj\\.org\\//"],"exclude":[],"allowHash":false,"depth":-1,"sitemap":null,"auth":null,"_authEncoded":null,"maxExtraHops":0,"maxDepth":1000000}]}
{"timestamp":"2025-03-03T03:59:12.861Z","logLevel":"info","context":"general","message":"Link Selectors","details":[{"selector":"a[href]","extract":"href","isAttribute":false}]}
{"timestamp":"2025-03-03T03:59:12.861Z","logLevel":"info","context":"general","message":"Behavior Options","details":{"message":"{\"autoplay\":true,\"autofetch\":true,\"autoscroll\":true,\"siteSpecific\":true,\"log\":\"__bx_log\",\"startEarly\":true,\"clickSelector\":\"a\"}"}}
{"timestamp":"2025-03-03T03:59:13.534Z","logLevel":"info","context":"worker","message":"Creating 1 workers","details":{}}
[...]

PR #806, opened 2025-04-01
Better default crawlId

ikreymer:
- set crawl id from collection, not other way around, to ensure unique redis keyspace for different collections
- by default, set crawl id to unique value based on host and collection, eg. '@hostname-@id'
- don't include '@id' in collection interpolation, can only used hostname or timestamp
- fixes issue mentioned / workaround provided in #784 
- ci: add docker login + cacheing to work around rate limits
- tests: fix sitemap tests

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -204,6 +204,8 @@ export class Crawler {
     this.params = args as CrawlerArgs;
     this.origConfig = this.params.origConfig;
 
+    this.crawlId = this.params.crawlId;
+
     // root collections dir
     this.collDir = path.join(
       this.params.cwd,
@@ -213,7 +215,7 @@ export class Crawler {
     this.logDir = path.join(this.collDir, "logs");
     this.logFilename = path.join(
       this.logDir,
-      `crawl-${new Date().toISOString().replace(/[^\d]/g, "")}.log`,
+      `${interpolateFilename("$ts", "")}.log`,
     );
 
     const debugLogging = this.params.logging.includes("debug");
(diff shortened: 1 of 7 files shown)


src/crawler.ts
```diff
@@ -2739,13 +2739,11 @@ self.__bx_behaviors.selectMainBehavior();
...
     this.lastSaveTime = now.getTime();
 
-    const ts = now.toISOString().slice(0, 19).replace(/[T:-]/g, "");
-
     const crawlDir = path.join(this.collDir, "crawls");
 
     await fsp.mkdir(crawlDir, { recursive: true });
 
-    const filenameOnly = `crawl-${ts}-${this.crawlId}.yaml`;
+    const filenameOnly = `crawl-${this.crawlId}.yaml`;
```
tw4l, line 2746: This ends up writing files named like `crawl-7fd3b6baf75f-crawl-20250401141145058.yaml` by default. Maybe we drop the initial `crawl-` prefix?

```suggestion
    const filenameOnly = `${this.crawlId}.yaml`;
```

tw4l (approved):
One very minor suggestion, looks good

src/crawler.ts
ikreymer: Changed this to be `$ts-$id.yaml` to match previous pattern.

ikreymer: We actually do want a unique ts here as well, as its a new save state

tw4l: ah true!

PR #807, opened 2025-04-01
Tweaks for custom behavior loading

tw4l:
Follow-up to #712 

Fixes a few things I noticed while testing out https://github.com/webrecorder/browsertrix/pull/2520

- Ignore `.git` directory of git repositories when recursively walking cloned git repo to collect custom behaviors
- Increase MAX_DEPTH for collecting behaviors to 5 (previous limit of 2 was overly restrictive for Git repositories)
- Log name of custom behavior scripts added as info messages in `behavior` context

### DIFF
--- a/src/util/file_reader.ts
+++ b/src/util/file_reader.ts
@@ -127,6 +127,11 @@ async function collectLocalPathBehaviors(
     const stat = await fsp.stat(resolvedPath);
 
     if (stat.isFile() && ALLOWED_EXTS.includes(path.extname(resolvedPath))) {
+      logger.info(
+        "Custom behavior script added",
+        { filename },
+        "behavior"
+      );
       const contents = await fsp.readFile(resolvedPath);
       return [
         {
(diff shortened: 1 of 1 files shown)


src/util/file_reader.ts
```diff
@@ -126,6 +127,7 @@ async function collectLocalPathBehaviors(
     const stat = await fsp.stat(resolvedPath);
 
     if (stat.isFile() && ALLOWED_EXTS.includes(path.extname(resolvedPath))) {
+      logger.info("Custom behavior script added", { filename }, "behavior");
```
tw4l, line 130: This may actually be confusing/unhelpful for behaviors downloaded via URL, since we give them an arbitrary filename :/

ikreymer: Maybe we add something about where it was downloaded from instead? If its possible to pass in here?
Like displaySource or something?

ikreymer (approved):

src/util/file_reader.ts
tw4l: Good call! Changed it to pass in the original filename from the URL for logging purposes

ikreymer:
Updated to just display the full URL as source, no reason to just show pathname, will help avoid confusion later on.

Issue #791, opened 2025-03-18
State file written and then immediately deleted


Mr0grog:
In a recent crawl, the browser got stuck (it timed out getting a new window/new page) and, while exiting, the crawler wrote a state file and then immediately deleted it (leaving with no state file to resume from).

Specifically, one of two workers hit a fatal error, and then the crawler:
1. Writes a new state file.
2. Deletes an old state file. (So far this makes sense, I have `saveState: always; saveStateHistory: 1`.)
3. Overwrites the state file from (1).
4. Deletes that state file!

I *suspect* the two workers are stepping on each others’ toes, so for now I’ve set `saveStateHistory` to be the same as the number of workers.

I have only skimmed the internals, but if someone has any ideas/pointers on the cause, I’m to try and help fix.

---

Full details:
- Docker image `webrecorder/browsertrix-crawler:1.5.8`
- Config file:

    ```yaml
    workers: 2
    saveState: always
    saveStateHistory: 1
    scopeType: page
    rolloverSize: 8000000000
    ```

I did **not** have verbose logging turned on. The fatal error logs:

```
{"timestamp":"2025-03-18T06:14:41.303Z","logLevel":"warn","context":"worker","message":"Page Worker Timeout","details":{"seconds":190,"page":"https://maps.nrel.gov/slope","workerid":0}}
{"timestamp":"2025-03-18T06:14:51.333Z","logLevel":"error","context":"worker","message":"Page Close Timed Out","details":{"seconds":10,"page":"https://maps.nrel.gov/slope","workerid":0}}
{"timestamp":"2025-03-18T06:15:11.354Z","logLevel":"warn","context":"worker","message":"New Window Timed Out","details":{"seconds":20,"workerid":0}}
[...]

PR #808, opened 2025-04-01
stuck page handling: when attempting to restart browser, add more retries

ikreymer:
fixes issue mentioned in: https://github.com/webrecorder/browsertrix-crawler/issues/791#issuecomment-2734342186

### DIFF
--- a/src/util/worker.ts
+++ b/src/util/worker.ts
@@ -237,6 +237,9 @@ export class PageWorker {
             throw new Error("Unable to load new page, browser needs restart");
           } else {
             this.crawler.browser.killBrowser();
+            // retry again after killing browser
+            retry = 0;
+            await sleep(2.0);
           }
         }
 
--- a/tests/sitemap-parse.test.js
+++ b/tests/sitemap-parse.test.js
@@ -31,7 +31,6 @@ async function waitContainer(containerId) {
 
 async function runCrawl(numExpected, url, sitemap="", limit=0, numExpectedLessThan=0, extra="") {
   const command = `docker run -d -p 36381:6379 -e CRAWL_ID=test webrecorder/browsertrix-crawler crawl --url ${url} --sitemap ${sitemap} --limit ${limit} --context sitemap --logging debug --debugAccessRedis ${extra}`;
-  console.log(command);
   const containerId = child_process.execSync(command, {encoding: "utf-8"});
 
   await sleep(3000);


tw4l (approved):

Mr0grog:
Sorry for commenting on a closed PR there, but this change was in response to something I posted, and I am confused at how this doesn’t just add to the problem.

The thing I was trying to call out in https://github.com/webrecorder/browsertrix-crawler/issues/791#issuecomment-2734342186 was that in [this `if (retry >= 2)` block in `PageWorker.initPage()`](https://github.com/webrecorder/browsertrix-crawler/blob/6898bcf7ae8a41b72b735917167c21fc7e00c409/src/util/worker.ts#L234-L244), the browser is marked as crashed (line 235) so there will be no retries. But then it:
1. steps out of that `if` block,
2. sleeps for half a second *(not necessary since we are effectively done)*,
3. logs that it will retry *(it will not actually retry)*,
4. goes back to the top of the loop, where [the loop exits instead of retrying](https://github.com/webrecorder/browsertrix-crawler/blob/6898bcf7ae8a41b72b735917167c21fc7e00c409/src/util/worker.ts#L151) *(the loop condition is [`this.crawler.isCrawlRunning()`](https://github.com/webrecorder/browsertrix-crawler/blob/6898bcf7ae8a41b72b735917167c21fc7e00c409/src/crawler.ts#L1501-L1504), which returns false after [marking the browser as crashed](https://github.com/webrecorder/browsertrix-crawler/blob/6898bcf7ae8a41b72b735917167c21fc7e00c409/src/crawler.ts#L1782-L1785))*.

When we’ve entered this `if` block, It seems like step 2 is unnecessary and step 3 is inaccurate. The change here doesn’t fix any of that, though, as far as I can tell. It just adds more delay like step 2, which again seems unnecessary. And then there’s the comment about “retry again after killing browser,” but as far as I can tell, that’s not going to happen because of step 4 (at first I was worried `retry = 0` was going to create an infinite loop, but because of step 4 `retry` will never be checked).

Am I missing something?

ikreymer:
This is an attempt to recover when the browser gets into the 'can't open any new windows state'. I was able to recover from this by killing the main process and puppeteer seemed to be able to recover. It does need more testing. In our use case, we just exit (we run with --restartOnError) but this is an attempt to not have to do that, but clearly needs another pass.

Yes, the `markBrowserCrashed` should be moved into the if statement so that happens only when throwing, or it should clear that. Here's a branch that should address that, if you have a repro where this happens, would appreciate any testing to see if this makes it better:

https://github.com/webrecorder/browsertrix-crawler/compare/main...retry-before-marking-crashed

(we generally run with --restartsOnError so aren't hitting this but standalone users of crawler may see this occasionally)

Mr0grog:
Ah, OK, this additional change makes a lot more sense to me now.

I used to hit this problem somewhat regularly in a crawl of ~4,800 URLs, but have not run into it since I figured out that ArcGIS was the problem and separated out all the `arcgis.com` URLs (66 of them) into a separate crawl with only one worker. I *think* it was probably a memory pressure issue, but never had the time to investigate really deeply.

ikreymer:
It started happening periodically around Chromium 133 or so, others have reported it as well, seems there's not much we can do other than restart the browser when it happens, or just restart the crawl. I'll go ahead and merge the fix since that should result in it actually proceeding, and we'll continue to test this.

PR #818, opened 2025-04-08
Support for behaviors from 'recorder flow' JSON created in devtools

ikreymer:
New Feature:
- support 'flow behavior' from JSON specification
- detect .json files via --customBehaviors
- log behavior progress while running
- logging tweaks (via browsertrix-behaviors 0.8.4) to limit logging for custom behaviors
- differentiate logging for iframes, move more behavior messages to debug
- docs to be added in separate follow-up PR


src/util/browser.ts
```diff
@@ -341,7 +341,13 @@ export class Browser {
...
     }
 
-    logger.info("Run Script Started", details, contextName);
+    const isTopFrame = !frame.parentFrame();
+
+    if (isTopFrame) {
+      logger.debug("Run Script Started", details, contextName);
+    } else {
+      logger.debug("Run Script Started in iframe", details, contextName);
+    }
```
tw4l, line 350: Bit of a nit/can be cleaned up in a follow-up, but our casing for log messages is a bit all over the place. I'd suggest we move to sentence case (only capitalize first word and proper nouns, being selective about the latter)

PR #821, opened 2025-04-29
state: add trimqueue() redis command to trim queue / seen list

ikreymer:
useful to support dynamically lowering pageLimit when restarting a crawl
fixes issue raised in webrecorder/browsertrix#2514


tw4l (approved):
Looks good, tested as working. Looks like test failures are unrelated sitemap issues we can address in a separate PR.

Issue #817, opened 2025-04-07
Brave downloads spellcheck dictionary


bensteinberg:
Hi -- I noticed that some crawls I've made included an unexpected resource from dict.brave.com, and it turned out to be `https://dict.brave.com/edgedl/chrome/dict/en-us-10-1.bdic`, a file of 441K. Brave downloads a spellcheck dictionary on first run, apparently, when there isn't one already present. I've observed this on an Ubuntu machine but not on a Mac. In both cases, the command line was

```
docker run -v $PWD/crawl.txt:/app/seedFile.txt -v $PWD/crawls:/crawls/ webrecorder/browsertrix-crawler crawl --seedFile /app/seedFile.txt --combineWARC --depth 0 --workers 4 --behaviors autoplay,autofetch --collection test-crawl
```

Have you run into this? Do you know if there's a way to suppress this behavior?

ikreymer:
@bensteinberg Can you try this branch to see if that fixes it? https://github.com/webrecorder/browsertrix-crawler/tree/disable-spellcheck

Added disabling the spellcheck via enterprise policies: https://chromeenterprise.google/policies/#SpellcheckEnabled

bensteinberg:
Thanks! I'm trying it now, and I think I'm running into image caching problems. Is the complete sequence to check out a branch, then run `docker compose build --no-cache`, then the crawl? I've also been deleting images in between runs, not sure it makes a difference.

bensteinberg:
I now can't reproduce the problem, and I wonder if an upgrade to the underlying browser fixed this. I'm going to poke at this a little more, but I'll close the ticket for now. Thanks!

PR #822, opened 2025-04-30
Config Policy Update

ikreymer:
Fixes webrecorder/replayweb.page#416

Update enterprise policy to:
- Disable Spellcheck, which should include downloading spellcheck dictionary, possibly issue raised in #817 
- Disable automatic http->https redirects, which insert an extra 307 response, as raised in: webrecorder/replayweb.page#416

### DIFF
--- a/config/policies/chromium.json
+++ b/config/policies/chromium.json
@@ -7,5 +7,6 @@
     "URLBlocklist": [
         "file://*"
     ],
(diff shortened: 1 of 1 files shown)


tw4l (approved):
LGTM

Tested the example from https://github.com/webrecorder/replayweb.page/issues/416 and confirmed that there's now only the one 200 response for the index page in the resulting indices.

PR #825, opened 2025-05-01
support pause interrupt:

### LINKED ISSUE #824: Support Pausing Crawl
Add a new interrupt reason, enabled by Redis, similar to time, size limit, etc... to indicate crawl is 'paused'. Will exit the crawler, but also upload WACZ (as opposed to default interrupt)

Crawl is paused if `<id>:paused` key is set in redis (similar to stopping)

ikreymer:
- add new interrupt reason / exit code
- add isCrawlPaused() which checks redis <id>:paused key
- exit gracefully, upload WACZ file when paused

version: bump to 1.6.1
fixes #824

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "browsertrix-crawler",
-  "version": "1.6.0",
+  "version": "1.6.1",
   "main": "browsertrix-crawler",
   "type": "module",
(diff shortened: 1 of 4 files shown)


tw4l (approved):
Nice and simple, working well!

PR #834, opened 2025-05-10
lang code fixes:

### LINKED ISSUE #833: Validation of language codes
Crawler should validate that the code passed to `--lang` is actually valid, otherwise subtle errors may happen (eg. pages not actually loading).
Given an incorrect code, the crawler should fail instantly.

Also, the `--lang` code should be ignored if using profiles, to avoid overriding the profile settings.

ikreymer:
- validate --lang values, fail immediately with invalid iso-639-1 country code
- ignore --lang value when using profile, print warning that profile language takes precedence
- fixes #833

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "get-folder-size": "^4.0.0",
     "husky": "^8.0.3",
     "ioredis": "^5.3.2",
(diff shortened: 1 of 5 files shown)


tw4l (approved):
Could use a test or two, otherwise looks good

PR #835, opened 2025-05-11
optimization: normalize dedup status: treat 0 (response code not yet known) or 206 as 200…

ikreymer:
Avoids fetching duplicate content when fetched through different code path (eg. autoplay behavior calling fetch, vs video playing automatically)

### DIFF
--- a/src/util/state.ts
+++ b/src/util/state.ts
@@ -25,6 +25,15 @@ export enum QueueState {
 }
 
 // ============================================================================
+// treat 0 or 206 as 200 for purposes of dedup
+function normalizeDedupStatus(status: number): number {
+  if (status === 0 || status === 206) {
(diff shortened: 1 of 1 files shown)


tw4l (commented):
Is there a risk in treating 0 as 200 for purposes of dedupe, that content may actually not have been captured?

ikreymer:
> Is there a risk in treating 0 as 200 for purposes of dedupe, that content may actually not have been captured?

The check happens before the fetch starts, to see if it should be skipped. Sometimes the status code is already known (if coming from browser) while other times it is not yet (when doing a full fetch call in node). I guess technically it could still fetch headers to know what status is in the latter case, but that requires more refactoring (can be done as a follow-up)

PR #837, opened 2025-05-13
Support host-specific proxies with proxy config YAML

### LINKED ISSUE #836: Support host-specific proxy settings.
It is often useful to be able to set up host-specific proxy, to route certain traffic through a proxy, or use different proxies for different domains, matched by regex.

This can be configured through a separate proxy yaml file (or could be part of the same settings as the existing config with new keys added).

This can be implemented with proxy PAC scripts which are supported in Brave (and support socks5 auth!). The crawler can generate a proxy PAC script from a config file that and pass that to the browser.

The config file could look as follows, for example:

```yaml
matchHosts:
  # load all URLs from example.com through 'example-1-proxy'
  example.com/.*: example-1-proxy 

  # load all URLS from https://my-social.example.com/.*/posts/ through
  # a different proxy
  https://my-social.example.com/.*/posts/: social-proxy

  # optional default proxy
  "": default-proxy

proxies:
  # SOCKS5 proxy just needs a URL
  example-1-proxy: socks5://username:password@my-socks-5-proxy.example.com

  # SSH proxy also should have at least a 'privateKeyFile'
  social-proxy:
    url: ssh://user@my-social-proxy.example.com
    privateKeyFile: /proxies/social-proxy-private-key
    # optional
    publicHostsFile: /proxies/social-proxy-public-hosts

  default-proxy:
    url: ssh://user@my-social-proxy.example.com
    privateKeyFile: /proxies/default-proxy-private-key
```

ikreymer:
- Adds support for YAML-based config for multiple proxies, containing 'matchHosts' section by regex and 'proxies' declaration, allowing matching any number of hosts to any number of named proxies.
- Specified via --proxyServerConfig option passed to both crawl and profile creation commands.
- Implemented internally by generating a proxy PAC script which does regex matching and running browser with the specified proxy PAC script served by an internal http server.
- Also support matching different undici Agents by regex, for using different proxies with direct fetching
- Precedence: --proxyServerConfig takes precedence over --proxyServer / PROXY_SERVER, unless --proxyServerPreferSingleProxy is also provided
- Updated proxies doc section with example
- Updated tests with sample bad and good auth examples of proxy config

Fixes #836


docs/docs/user-guide/proxies.md
```diff
@@ -80,7 +80,56 @@ The above proxy settings also apply to [Browser Profile Creation](browser-profil
...
+
+```sh
+docker run -v $PWD/crawls:/crawls -v $PWD/proxies:/proxies -it webrecorder/browsertrix-crawler --url https://example.com/ --proxyServerConfig /proxies/proxyConfig.yaml
+```
+
+Note that if SSH proxies are provided, an SSH tunnel must be opened for each one before the crawl starts.
+The crawl will not start if any of the SSH proxy connections fail, even if a host-specific proxy is not actually used.
+SOCKS5 and HTTP proxy connections are attempted only on first use.
+
+The same `--proxyServerConfig` option can also available be used browser profile creation `create-login-profile` command in the same way.
```
tw4l:
```suggestion
The same `--proxyServerConfig` option can also be used in browser profile creation with the `create-login-profile` command in the same way.
```

src/util/argParser.ts
```diff
@@ -641,6 +642,12 @@ class ArgParser {
           type: "string",
         },
 
+        proxyServerConfig: {
+          describe:
+            "if set, path to yaml/json file that configures multiple path servers per URL regex",
```
tw4l, line 654: Same question about JSON support as above

src/create-login-profile.ts
```diff
@@ -123,6 +123,12 @@ function initArgs() {
         type: "string",
       },
 
+      proxyServerConfig: {
+        describe:
+          "if set, path to yaml/json file that configures multiple path servers per URL regex",
```
tw4l, line 128: Does this actually support JSON? From docs and code I'm only seeing YAML support

src/util/proxy.ts
```diff
@@ -54,31 +105,124 @@ export function getSafeProxyString(proxyString: string): string {
...
+      sshProxyKnownHostsFile,
+    );
+  }
+
+  if (!params.proxyMap?.matchHosts || !params.proxyMap?.proxies) {
+    if (defaultProxyEntry) {
+      logger.debug("Using Single Proxy", {}, "proxy");
+    }
+    return { proxyServer: defaultProxyEntry?.proxyUrl };
+  }
```
tw4l, line 136: Maybe I'm reading this wrong, but it seems like the single proxyServer will only be returned here if host-specific proxies aren't also passed? If we're giving precedence to single proxy servers, maybe the wider if clause here isn't necessary?

src/create-login-profile.ts
ikreymer: YAML is supposed to be a superset of JSON, as of 1.2 at least! (Although there are some edge-cases, but not sure we need to worry about, according to: https://news.ycombinator.com/item?id=30052633)
We could attempt to parse as JSON if YAML parsing fails.

FWIW, the main config is also parsed via a YAML parser but is stored as JSON in Browsertrix - we have not had any issues with it.

src/util/proxy.ts
ikreymer: Ah good catch, I think I updated this later to only override the default proxy (with no host configured), so the other configs still take precedence. I can see a use case for both, I suppose could add another option for this..

tw4l: Yeah, maybe another option would be good. We have clients who require all traffic to come from a specific IP or country, so I think we do need to have an option to make sure that's the case, even if we want to by default have the host-specific proxies override.

docs/docs/user-guide/proxies.md
```diff
@@ -80,7 +80,56 @@ The above proxy settings also apply to [Browser Profile Creation](browser-profil
...
+Note that if SSH proxies are provided, an SSH tunnel must be opened for each one before the crawl starts.
+The crawl will not start if any of the SSH proxy connections fail, even if a host-specific proxy is not actually used.
+SOCKS5 and HTTP proxy connections are attempted only on first use.
+
+The same `--proxyServerConfig` option can also be used in browser profile creation with the `create-login-profile` command in the same way.
+
+### Proxy Precedence
+
+If both `--proxyServerConfig` and `--proxyServer`/`PROXY_SERVER` env var are specified, the single `--proxyServer`
+option takes precedence.
```
tw4l:
```suggestion
If both `--proxyServerConfig` and `--proxyServer`/`PROXY_SERVER` env var are specified, the `--proxyServerConfig` option takes precedence on matching hosts. To have the single `--proxyServer` option always take precedence instead, pass the `--proxyServerPreferSingleProxy` option.
```

PR #852, opened 2025-06-17
Support downloading seed file from URL

### LINKED ISSUE #841: Support downloading seedFile from online source
Connected to https://github.com/webrecorder/browsertrix/issues/2312

Similar to custom behaviors, the crawler should be able to download a seed file from any accessible URL, simply by specifying a URL instead of filepath to `--seedFile`.

We should additionally validate the URLs passed in the seed file.

tw4l:
Fixes #841 

Crawler work toward long URL lists in Browsertrix. This PR moves seed handling from the arg parser's validation step to the crawler's bootstrap step in order to be able to async fetch the seed file from a URL.

### DIFF
--- a/src/util/file_reader.ts
+++ b/src/util/file_reader.ts
@@ -52,7 +52,7 @@ async function writeUrlContentsToFile(
   pathDefaultExt: string,
 ) {
   const res = await fetch(url, { dispatcher: getProxyDispatcher() });
-  const ct = (res.headers.get("content-type") || "").toLowerCase();
+  const ct = (res.headers.get("content-type") || "").toLowerCase().split(";")[0];
   if (!allowedMimes.includes(ct)) {
     throw new Error(
       `Invalid Content-Type: ${ct}, expected one of: ${allowedMimes.join(",")}`,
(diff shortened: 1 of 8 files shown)


src/util/file_reader.ts
```diff
@@ -24,6 +24,35 @@ export type FileSource = {
 
 export type FileSources = FileSource[];
 
+async function getTempFile(
+  filename: string,
+  dirPrefix: string,
+): Promise<string> {
+  const tmpDir = `/tmp/${dirPrefix}-${crypto.randomBytes(4).toString("hex")}`;
```
```suggestion
  const tmpDir = path.join(os.tmpdir(), `${dirPrefix}-${crypto.randomBytes(4).toString("hex")}`);
```

tw4l:
It looks like GitHub always returns a content-type of `text/plain` for raw files such as https://raw.githubusercontent.com/webrecorder/browsertrix-crawler/refs/heads/main/tests/custom-behaviors/custom-2.js, so we may want to compute the mime type from what we receive rather than relying on what the web server says it is

ikreymer (approved):

PR #856, opened 2025-06-28
Add option to save local/sessionStorage

### LINKED ISSUE #855: Add option to serialize local/sessionStorage
Add an option to save the localStorage and sessionStorage for each page.
AWP already supports this, and RWP can replay this, as long as its injected into the page, along with cookies.

ikreymer:
If --saveStorage is set, localStorage and sessionStorage will be serialized with the WARC record for the page.
If a page redirects, track what the current page URL is and save storage as part of the page's WARC record.

Fixes #855 

Note: should this be an option or enabled by default?

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -596,6 +596,12 @@ class ArgParser {
           default: [],
         },
 
+        saveStorage: {
+          describe:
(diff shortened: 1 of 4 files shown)


tw4l (approved):
I would definitely keep this an opt-in feature so that people don't accidentally share WACZs that have data they don't intend to share. Otherwise looks great!

PR #861, opened 2025-07-04
Support option to fail crawl on content check

### LINKED ISSUE #860: Add option to fail crawl after initial content check
Allows custom behaviors to check state of page and fail if the crawl if some condition is not met (eg. page is not logged in).

ikreymer:
- add --failOnContentCheck for quick fail if content check in behavior fails
- expose __bx_contentCheckFailed to cause an immediately failure from behavior
- only allow failing crawl due to content check from within awaitPageLoad() callback
- set a 'failReason' key to track that crawl failed due to a particular content check reason
- deps: update to browsertrix-behaviors 0.9.0, update to wabac.js (2.23.6)
- fixes #860

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -587,6 +587,13 @@ class ArgParser {
           default: false,
         },
 
+        failOnContentCheck: {
+          describe:
+            "If set, allows for behaviors to fail a crawl with custom reason based on content, (eg. logged out, limited, etc...)",
+          type: "boolean",
+          default: true,
+        },
+
         customBehaviors: {
           describe:
             "Custom behavior files to inject. Valid values: URL to file, path to file, path to directory" +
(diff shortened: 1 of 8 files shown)


src/util/argParser.ts
```diff
@@ -587,6 +587,13 @@ class ArgParser {
           default: false,
         },
 
+        failOnContentCheck: {
+          describe:
+            "If set, allows for behaviors to fail a crawl with custom reason based on content, (eg. logged out, limited, etc...)",
```
```suggestion
            "If set, allows for behaviors to fail a crawl with custom reason based on content (e.g. logged out)",
```

src/util/argParser.ts
```diff
@@ -587,6 +587,13 @@ class ArgParser {
...
+          type: "boolean",
+          default: true,
```
tw4l: I'm not sure we want to set the default to true, as this might be a breaking change for some folks' crawl configs (e.g. for posts that don't actually require being logged in to capture).

ikreymer: oops, good catch, that was just set for testing!

ikreymer:
```suggestion
          default: false,
```

tw4l (approved):

PR #868, opened 2025-07-23
url queueing: log skipped URLs as errors if depth === 0

### LINKED ISSUE #866: Log skipped / invalid URLs that are seeds as errors
To increase visibility in Browsertrix, log issues with seed URLs (eg. read from a list) as errors instead of as debug messages.

ikreymer:
- will ensure sees from URL list are reported as errors if skipped
- also set logging context to 'scope' instead of 'links'
- fixes #866

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -2459,25 +2459,35 @@ self.__bx_behaviors.selectMainBehavior();
       this.pageLimit,
     );
 
+    const logContext = depth === 0 ? "links" : "scope";
+    const logLevel = depth === 0 ? "error" : "debug";
+
     switch (result) {
       case QueueState.ADDED:
-        logger.debug("Queued new page url", { url, ...logDetails }, "links");
+        logger.logAsJSON(
+          "Queued new page url",
+          { url, ...logDetails },
+          logContext,
+          logLevel,
+        );
         return true;
 
       case QueueState.LIMIT_HIT:
-        logger.debug(
+        logger.logAsJSON(
           "Not queued page url, at page limit",
           { url, ...logDetails },
-          "links",
+          logContext,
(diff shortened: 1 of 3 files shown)


src/crawler.ts
```diff
@@ -2459,25 +2459,30 @@ self.__bx_behaviors.selectMainBehavior();
       this.pageLimit,
     );
 
+    const logContext = depth === 0 ? "links" : "scope";
```
```suggestion
    const logContext = depth === 0 ? "scope" : "links";
```

src/crawler.ts
```diff
@@ -2459,25 +2459,30 @@ self.__bx_behaviors.selectMainBehavior();
...
     switch (result) {
       case QueueState.ADDED:
-        logger.debug("Queued new page url", { url, ...logDetails }, "links");
+        logger.debug("Queued new page url", { url, ...logDetails }, logContext);
         return true;
 
       case QueueState.LIMIT_HIT:
-        logger.debug(
+        logger.logAsJSON(
           "Not queued page url, at page limit",
```
tw4l:
```suggestion
          "Page url not queued, at page limit",
```

^ Small wording fix, current wording is weird grammatically

src/crawler.ts
```diff
@@ -2459,25 +2459,30 @@ self.__bx_behaviors.selectMainBehavior();
...
+          logContext,
+          logLevel,
         );
         this.limitHit = true;
         return false;
 
       case QueueState.DUPE_URL:
-        logger.debug(
+        logger.logAsJSON(
           "Not queued page url, already seen",
```
tw4l:
```suggestion
          "Page url not queued, already seen",
```

tw4l (approved):

PR #871, opened 2025-07-26
quickfix: WACZ upload retry support:

ikreymer:
- if a failure occurs on failed upload, and crawler restarts on error, exit with 'interrupt' to allow for automatic restart (eg. in Browsertrix app)
- otherwise, a failed upload will exit the crawl with no WACZ, resulting in overall crawl failure

### Testing
This is a bit tricky to test, but noticed at least one error in uploads in Browsertrix.

One way to test it is to
1) Add a `if (Math.random() > 0.5) { throw new Error("test"); }` within generateWACZ()
2) Run a crawl in Browsertrix, if the error condition is hit, error logs will be printed one or more times, but crawl will eventually succeed.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1962,6 +1962,8 @@ self.__bx_behaviors.selectMainBehavior();
       logger.error("Error creating WACZ", e);
       if (!streaming) {
         logger.fatal("Unable to write WACZ successfully");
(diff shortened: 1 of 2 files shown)


tw4l (approved):
Thanks for the testing advice!

PR #873, opened 2025-07-29
Don't trim to limit if limit is default of 0

### LINKED ISSUE #872: `--pageLimit` of 0 removes all queued pages from loaded state
**Summary:**

Although it is not explicitly stated in the docs, it seems like the intended behavior of `--pageLimit` at its default (0) is for there to be no limit on the number of crawled pages. This works properly for me when starting a crawl from scratch, but if I load a previously saved crawl state, a `pageLimit` of 0 removes all queued pages from the queue after loading the state.

It seems like this is a result of [`trimToLimit`](https://github.com/webrecorder/browsertrix-crawler/blob/bc4d649307e23824ec058b2fbfd1f8c060991d68/src/util/state.ts#L460) not accounting for a `pageLimit` of 0.

**Expected behavior**:

`--pageLimit` of 0 should allow loading unlimited queued pages from a `.yaml` config file.

**Observed behavior**:

When `--pageLimit` is 0, all queued pages from a saved crawl state from a `.yaml` file are removed.

tw4l:
Fixes #872 

Fix for restarting crawl from saved state, where the default `--limit` value of 0 was incorrectly preventing any URLs from being re-queued.

## Testing instructions (problem)

1. Create a crawl from latest main branch of crawler with default limit
2. Interrupt the crawl early on
3. Restart crawl from saved state with `--config <FILE>`
4. Verify crawler exists with the queued pages from saved state not re-queued

## Testing instructions (fix)

1. Create a new crawl from this branch with default limit
2. Interrupt the crawl early on
3. Restart crawl from saved state with `--config <FILE>`
4. Verify crawler successfully re-queues pages and continues from where it left off

### DIFF
--- a/src/util/state.ts
+++ b/src/util/state.ts
@@ -458,6 +458,10 @@ return inx;
   }
 
   async trimToLimit(limit: number) {
+    if (limit === 0) {
+      return;
+    }
+
     const totalComplete =
       (await this.numPending()) +
       (await this.numDone()) +


ikreymer (approved):
Great! (Should probably set this default to -1 instead of 0 but not in this PR)

PR #874, opened 2025-07-30
behavior logging: remove last line dupe check for behavior logs

ikreymer:
Shouldn't skip multiple log messages, as this is unexpected behavior for user-defined behaviors.


tw4l (approved):

ikreymer:
Updated to 0.9.2 for behavior webrecorder/browsertrix-behaviors#104 fix

PR #877, opened 2025-08-13
feat: add extraChromeArgs support for passing custom Chrome flags

aponb:
This change introduces a new CLI option --extraChromeArgs to Browsertrix Crawler, allowing users to pass arbitrary Chrome flags without modifying the codebase.

This approach is future-proof: any Chrome flag can be provided at runtime, avoiding the need for hard-coded allowlists.
Maintains backward compatibility: if no extraChromeArgs are passed, behavior remains unchanged.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -609,7 +609,7 @@ export class Crawler {
     const extra = this.params.extraChromeArgs;
     if (Array.isArray(extra) && extra.length > 0) {
       for (const v of extra) {
-        if (v !== undefined && v !== null && v !== "") {
+        if (v) {
           args.push(String(v));
         }
       }
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -594,7 +594,7 @@ export class Crawler {
   }
 
   extraChromeArgs() {
-    const args = [];
+    const args: string[] = [];
     if (this.params.lang) {
       if (this.params.profile) {
         logger.warn(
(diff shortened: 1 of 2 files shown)


src/crawler.ts
```diff
@@ -600,6 +600,16 @@ export class Crawler {
         args.push(`--accept-lang=${this.params.lang}`);
       }
     }
+
+    const extra = this.params.extraChromeArgs;
+    if (Array.isArray(extra) && extra.length > 0) {
+      for (const v of extra) {
+        if (v !== undefined && v !== null && v !== "") {
```
```suggestion
        if (v) {
```
probably sufficient here?

aponb: true. a bit overcautious

aponb:
We’re running Browsertrix in a virtualized environment. Our biggest bottleneck is limited fast local storage, while the crawler writes a lot of small files.
Our main use case is to be able to redirect the Chrome cache to another storage location with
--disk-cache-dir, and also control related settings like --disk-cache-size. Having the flexibility to pass these flags directly lets us optimize performance without needing to patch Browsertrix itself every time.
So the initial motivation was the cache directory, but having a general extraChromeArgs option makes it future-proof: if other flags become useful later, we don’t need code changes.

aponb:
I rebased this, but some checks are still not successful. It seems there is a login to dockerhub required somewhere?

PR #880, opened 2025-08-25
Async Fetch Refactor

ikreymer:
- separate out reading stream response while browser is waiting (not really async) from actual async loading, this is not handled via fetchResponseBody()
- unify async fetch into first trying browser networking for regular GET, fallback to regular fetch() 
- load headers and body separately in async fetch, allowing for cancelling request after headers
- refactor direct fetch of non-html pages: load headers and handle loading body, adding page async, allowing worker to continue loading browser-based pages (should allow more parallelization in the future)
- unify WARC writing in preparation for dedup: unified serializeWARC() called for all paths, WARC digest computed, additional checks for payload added for streaming loading


src/crawler.ts
```diff
@@ -1280,6 +1265,10 @@ self.__bx_behaviors.selectMainBehavior();
   }
 
   async pageFinished(data: PageState, lastErrorText = "") {
+    // not yet finished
+    if (data.asyncLoading) {
+      return;
+    }
```
tw4l, line 1271: I assume we'd end up here if the page worker timeout is hit (or the worker crashes) before the page has finished loading async. Is there any tidying up we want to do in that case rather than just returning?

ikreymer: This is mainly for direct fetch path:
- to make sure that this call is ignored: https://github.com/webrecorder/browsertrix-crawler/blob/async-fetch-refactor/src/util/worker.ts#L314 (and worker can move on to next page)
- instead, the pageFinished() is called again from here after setting asyncLoading to false https://github.com/webrecorder/browsertrix-crawler/blob/async-fetch-refactor/src/util/recorder.ts#L1936

PR #888, opened 2025-09-30
Add option to respect robots.txt disallows

### LINKED ISSUE #631: Add option to fetch and respect disallows from `robots.txt`
It would be nice if the crawler could automatically fetch rules from `robots.txt` and add `exclusion` rules for every rule present in the `robots.txt` file.

I think this functionality should even be turned-on by default to avoid annoying servers which have clearly expressed what they do not want "external systems" to mess with.

At Kiwix, we have lots of non-tech users configuring zimit to do a browertrix crawl. In most cases, they have no idea what a `robots.txt` is, so having the switch turned-on by default would help a lot. That being said, I don't mind if it is off by default, we can do the magic to turn it on by default in zimit ^^

tw4l:
Fixes #631 

This PR introduces a `--robots` CLI flag to the crawler. When enabled, the crawler will attempt to fetch robots.txt for each host encountered during crawling, and if found will check if pages are disallowed before adding them to the crawl queue.

Fetched robots.txt bodies are cached by their URL in Redis using an LRU cache mechanism which retains the 100 most recently accessed ones in the cache to prevent memory usage from getting out of control.

Robots.txt bodies are parsed and checked for page allow/disallow status using the https://github.com/samclarke/robots-parser library, which is the most active and well-maintained implementation I could find with TypeScript types.

I have added basic tests that check the log lines to ensure robots.txt is being fetched and cached, but **I haven't been able to find a robots.txt on a Webrecorder-managed domain with disallows to test that disallowed URLs are not actually queued.** Maybe we could set up a single disallow on our website's robots.txt for this purpose?

## Manual testing

1. In `util/constants.ts`, set `ROBOTS_CACHE_LIMIT` to 2
2. Re-build the crawler image and run a crawl across multiple hosts, with `--robots` and `--logging debug` enabled and a mixture of URLs that do and do not have a robots.txt specified, and for those that do, a mixture of URLs that will be allowed and disallowed, e.g.:

```sh
docker compose build
docker compose run crawler crawl --url https://forums.gentoo.org --url https://forums.gentoo.org/search.php --url https://forums.gentoo.org/statistics.php --url https://forums.gentoo.org/viewforum-f-33.html\?sid\=ed69503a6d2b866c8bcc28307f8e7d59 --url https://webrecorder.net/browsertrix --url https://webrecorder.net --url https://bl.uk --url https://bl.uk/collection --url https://bl.uk/visit --url https://www.kb.dk --url https://www.kb.dk/README.txt --url https://www.kb.dk/search/ --url https://www.kb.dk/en/find-materials --url https://bitarchivist.net --limit 20 --scopeType page --collection robots-cache --robots --generateWACZ --logging debug,errors
```
3. Inspect the crawl log for relevant messages with `robots` and `links` context. What should be present:

*Log lines for fetching and caching robots.txt*

```json
{"timestamp":"2025-09-30T14:45:03.888Z","logLevel":"debug","context":"robots","message":"Fetching robots.txt","details":{"url":"https://forums.gentoo.org/robots.txt"}}
{"timestamp":"2025-09-30T14:45:04.226Z","logLevel":"debug","context":"robots","message":"Caching robots.txt body","details":{"url":"https://forums.gentoo.org/robots.txt"}}
```
*Log lines for using cached robots.txt*

```json
"timestamp":"2025-09-30T14:45:04.234Z","logLevel":"debug","context":"robots","message":"Using cached robots.txt body","details":{"url":"https://forums.gentoo.org/robots.txt"}}
```

*Log lines for not queueing page URLs disallowed by robots*

```json
{"timestamp":"2025-09-30T14:45:04.235Z","logLevel":"debug","context":"links","message":"Page URL not queued, disallowed by robots.txt","details":{"url":"https://forums.gentoo.org/search.php"}}
```

*Log lines for deleting least-recently accessed cached robots.txt when over cache limit*

```json
{"timestamp":"2025-09-30T14:45:04.843Z","logLevel":"debug","context":"robots","message":"Deleting cached robots.txt, over cache limit","details":{"url":"https://forums.gentoo.org/robots.txt"}}
```

*Log lines for hosts where robots.txt cannot be found or fetched*

```json
{"timestamp":"2025-09-30T14:45:05.619Z","logLevel":"debug","context":"robots","message":"Robots.txt not fetched","details":{"url":"https://bitarchivist.net/robots.txt","status":404}}
```

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1289,11 +1289,9 @@ self.__bx_behaviors.selectMainBehavior();
     url: string,
     logDetails: LogDetails,
   ): Promise<Robot | null> {
-    // Fetch robots.txt for url's host and return parser, caching robots
-    // bodies in Redis by their URL
-    // TODO: Consider using an LRU cache/only cache so many robots responses
-    // in Redis at one time and re-fetch if no longer in cache to avoid
-    // exhausting memory on very large crawls across many hosts
+    // Fetch robots.txt for url's host and return parser.
+    // Results are cached by robots.txt URL in Redis using an LRU cache
+    // implementation that retains the 100 most recently used values.
     const urlParser = new URL(url);
     const robotsUrl = `${urlParser.origin}/robots.txt`;
 
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -3,6 +3,8 @@ import path from "path";
 import fs, { WriteStream } from "fs";
 import os from "os";
 import fsp from "fs/promises";
+import { fetch as undiciFetch } from "undici";
+import robotsParser, { Robot } from "robots-parser";
 
 import {
   RedisCrawlState,
@@ -36,6 +38,7 @@ import { logger, formatErr, LogDetails, LogContext } from "./util/logger.js";
 import { WorkerState, closeWorkers, runWorkers } from "./util/worker.js";
 import { sleep, timedRun, secondsElapsed } from "./util/timing.js";
 import { collectCustomBehaviors, getInfoString } from "./util/file_reader.js";
+import { getProxyDispatcher } from "./util/proxy.js";
 
 import { Browser } from "./util/browser.js";
 
@@ -1249,6 +1252,98 @@ self.__bx_behaviors.selectMainBehavior();
     }
   }
 
+  async _fetchRobots(url: string) {
+    while (true) {
+      const resp = await undiciFetch(url, {
+        headers: this.headers,
+        dispatcher: getProxyDispatcher(url),
+      });
+
+      if (resp.ok) {
+        return resp;
+      }
+
+      const retry = resp.headers.get("retry-after");
+
+      if (retry) {
+        logger.debug(
+          "Robots.txt fetch: Retry after",
+          { url, retrySeconds: retry },
+          "robots",
+        );
+        await sleep(parseInt(retry));
+        continue;
+      }
+
+      logger.debug(
+        "Robots.txt not fetched",
+        { url, status: resp.status },
+        "robots",
+      );
+      return null;
+    }
+    return null;
+  }
+
+  async fetchAndParseRobots(
+    url: string,
+    logDetails: LogDetails,
+  ): Promise<Robot | null> {
+    // Fetch robots.txt for url's host and return parser, caching robots
+    // bodies in Redis by their URL
+    // TODO: Consider using an LRU cache/only cache so many robots responses
+    // in Redis at one time and re-fetch if no longer in cache to avoid
+    // exhausting memory on very large crawls across many hosts
+    const urlParser = new URL(url);
+    const robotsUrl = `${urlParser.origin}/robots.txt`;
+
+    const cachedRobots = await this.crawlState.getCachedRobots(robotsUrl);
+    if (cachedRobots) {
+      logger.debug(
+        "Using cached robots.txt body",
+        {
+          url: robotsUrl,
+          ...logDetails,
+        },
+        "robots",
+      );
+      return robotsParser(robotsUrl, cachedRobots);
+    }
+
+    try {
+      logger.debug(
+        "Fetching robots.txt",
+        { url: robotsUrl, ...logDetails },
+        "robots",
+      );
+      const resp = await this._fetchRobots(robotsUrl);
+      if (!resp) {
+        return null;
+      }
+      const content = await resp.text();
+
+      logger.debug(
+        "Caching robots.txt body",
+        { url: robotsUrl, ...logDetails },
+        "robots",
+      );
+      await this.crawlState.setCachedRobots(robotsUrl, content);
+
+      return robotsParser(robotsUrl, content);
+    } catch (e) {
+      // ignore
+    }
+    logger.warn(
+      "Failed to fetch robots.txt",
+      {
+        url: robotsUrl,
+        ...logDetails,
+      },
+      "robots",
+    );
+    return null;
+  }
+
   async awaitPageExtraDelay(opts: WorkerState) {
     if (this.params.pageExtraDelay) {
       const {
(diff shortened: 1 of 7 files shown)


src/util/state.ts
```diff
@@ -1025,6 +1036,38 @@ return inx;
...
+    await this._updateRobotsAccessTime(robotsUrl);
+    await this.redis.set(`${this.rkey}:${robotsUrl}`, body);
+
+    // prune least-recently used items in zset and robots cache if over limit
+    const cacheCount = await this.redis.zcard(this.lkey);
+    if (cacheCount > ROBOTS_CACHE_LIMIT) {
+      const diff = cacheCount - ROBOTS_CACHE_LIMIT;
+      const keysToDelete = await this.redis.zrange(this.lkey, 0, diff - 1);
+
+      for (const keyToDelete of keysToDelete) {
```
ikreymer, line 1054: I think this can actually be optimized further by using `this.redis.zremrangebyscore(this.lkey, 0, diff - 1)` which will remove all the matching keys.

ikreymer:
This handles the robots.txt matches for pages, but not for subpage resources. Perhaps we should have a:
`--robots all|pages` option, with `all` also intercepting all the requests (like blockrules) which can be handled with `interceptRequest()` similar to blockrules.

Perhaps we have a RobotsHandler that handles both interception + page queue filtering?

src/util/state.ts
```diff
@@ -200,7 +204,10 @@ export class RedisCrawlState {
   fkey: string;
   ekey: string;
   bkey: string;
+  rkey: string;
```
ikreymer, line 207: heh, maybe we should make these keys more user-friendly - I started a bad convention here, probably robotsKey and robotsLRUKey would be better...

src/crawler.ts
```diff
@@ -2462,6 +2555,18 @@ self.__bx_behaviors.selectMainBehavior();
       return false;
     }
 
+    if (this.params.robots) {
+      const robots = await this.fetchAndParseRobots(url, logDetails);
+      if (robots && robots.isDisallowed(url, "Browsertrix/1.0")) {
```
ikreymer: Should we make the robots useragent configurable (separate from crawling UA), eg. --robotsUA ?

ikreymer: I suppose it can default to `Browsertrix/<current package>` version

tw4l: This user agent currently isn't actually sent in the request. The library we're using requires a value here to check against the crawler user agents listed in the robots.txt, but at this point we're really just looking for the catch-all `*` rules.

If we want to use a Browsertrix user agent and send it with the request, I think best practice would have it be more consistent over time (so maybe just `"Browsertrix"` - [here is the list of Google common crawler user agents as a point of comparison](https://developers.google.com/search/docs/crawling-indexing/google-common-crawlers)) and ideally we'd publish it somewhere so people know what it is/can write rules for it. But I think there's a question there of whether we want that to happen. I'm a little back and forth on that myself. On the one hand, it seems in keeping with the intent of robots.txt. Arguments against might include:

1. We don't want to overwrite custom user agents that identify the actual institution doing the crawling rather than the tool they're using (which are used by some users to identify themselves for allowlists on particular sites)
2. It might lead to getting blocked or flagged as a bot more often than using the default browser user agent.

src/util/state.ts
tw4l: I'm not sure this is the way to go because we need to know what keys are being deleted so that we can also delete them from the actual cache (in the line above the `zrem` below), and `zremrangebyscore` doesn't return the delete keys but just a count of how many were deleted.

src/crawler.ts
tw4l: So can make it configurable, but I guess the question first is: do we actually want to send the request for robots.txt with this different user agent? Or are we just trying to use it in parsing the response (being a small enough fish for now that we're unlikely to ever come across any specific rules for Browsertrix)?

ikreymer: I was thinking just for the parsing library to test rules other than `*` for now

src/util/state.ts
ikreymer: right, yeah, would still need the `redis.del` call, but could eliminate the multiple zrem calls, but maybe not worth it..

ikreymer: since, in theory, something could get added in between and then there would be a left-over key, so probably safer to keep what you have.

src/crawler.ts
ikreymer: Implemented a --robotsAgent option for this, defaulting to `Browsertrix/1.x`

ikreymer:
For now, decided to keep just keep this focused on pages only, can add other options later.

Added some minor refactoring in: https://github.com/webrecorder/browsertrix-crawler/pull/888/commits/e9e373883194bc901452601c6e9834e11d4bd2e1:

- move robots logic to ./src/utils/robots.ts
- add --robotsAgent arg, defaulting to Browsertrix/1.x
- remove debug logging for 'using cached robots' as its too verbose (for every link), harder to see other message
- cache empty "" robots responses and treat as allow all
- treat non-200 and non-429 and non-503 responses as empty ""

Added the cacheing of empty "", and error responses as "", otherwise, these were getting fetched every time and not cached.

src/util/robots.ts
```diff
@@ -0,0 +1,122 @@
...
+    const resp = await fetch(url, {
+      headers,
+      dispatcher: getProxyDispatcher(url),
+    });
+
+    if (resp.ok) {
+      return await resp.text();
+    }
+
+    if (resp.status === 429 || resp.status === 503) {
```
ikreymer: Hm, while the retry mechanism is good and something we should perhaps add for required crawl resources (see: #921), here it could potentially lead to indefinite waiting just to fetch robots.txt, for every link. If a site is rate limiting access to robots.txt, that could potentially slow down the whole crawl.

I think we should reevaluate this in a later pass, and remove for now, also adding a timedRun() around the robots fetch, perhaps limiting to 10 seconds at most

ikreymer: Probably the way to handle this would be to fetch robots async, outside of the main crawler loop, and also avoid starting a fetch multiple times for the same domain.

ikreymer: Also added quick batching of multiple requests, to avoid sending multiple requests for robots.txt in one crawler.
Could further improve for multi-instance deployments, but should be good first pass.

PR #889, opened 2025-10-01
Dedup Initial Implementation

### LINKED ISSUE #884: Deduplication (Initial Support).
This is a top-level issue to cover core deduplication support in the crawler.

The general idea is as follows:

## Core Requirements

URL Revisit Record Support:
- [ ] The crawler will support hash-based deduplication via Redis, which could be the same redis as --redisStoreUrl (internal / external) or a secondary Redis instance (that can persist separately at a different lifetime). This redis will be specified via --redisDedupUrl.
- [ ] The Redis will store `digest->original url, original date` mapping. The first time a WARC-Payload-Digest is encountered, the mapping is added to this redis.
- [ ] When saving response records, after the WARC-Payload-Digest is computed, if an existing entry for the digest exists in the mapping, a revisit record is created instead.

This allows for URL agnostic deduplication across crawls.

Page Level Dedup / Skipping Support
- [ ] The digest can also be computed when the HTML is loaded as part of the Fetch handler. If the digest already exists (eg. HTML page is an exact duplicate), loading the whole page can be skipped, by returning an error code to the browser. A revisit record can be written for the page, but no other resources are loaded. The page list should also record this as a `duplicate` page. 
- [ ] A --minPageDedupDepth setting can be added to only do page dedup for seeds at a minimum depth, eg. the seed is never deduped, but perhaps pages level 2 or 3 can be deduped.

Populating the dedup index:
- [ ] It should be possible to populate a deduplication Redis from an existing set of WACZ file, or multi WACZ file, or multi WACZ collection manifest. The crawler will have have a special entry point that iterates over the WACZ files and processes all existing CDX(J) to set up the `digest->original url, original date` mapping.
- [ ] The state of the ingest should also be saved in the dedup Redis, so that the ingest can happen in parallel / can be retried until all the queued WACZ files are ingested.

## Additional Requirements

These will likely be done through follow-up issues, but listing them here:

- Option to control revisit vs skip within a single crawl. Currently, browser already dedups URL+status within the same crawl. Perhaps this could be made more customizable, eg. skip duplicate URLs for X minutes (by status, by hash?)
- Minimum dedup size. Only write revisit records if payload > minimum size
- Fuzzy dedup of page HTML: Apply a transformation regex to the page HTML to determine if its a duplicate, and skip if it is. Based on ideas from: https://github.com/natliblux/browsertrix-crawler-deduplication

ikreymer:
Fixes #884 

- Support for hash-based deduplication via a Redis provided with --redisDedupeUrl (can be same as default redis)
- Support for writing WARC revisit records for duplicates
- Support for new indexer mode which imports CDXJ from one or more WACZs (refactored from replay) to populate the dedup index
- Crawl and aggregate stats updated in dedupe index, including total urls, deduped URLs, conserved size (difference between revisit and response records), and estimated redundant size (aggregate) of duplicates not deduped.
- Track removed crawls on index update, support for --remove operation to purge removed crawls, otherwise removed crawl aggregate data is maintained.
- WIP on page-level dedupe (preempt loading entire pages) if HTML is a dupe/matches exactly.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -204,7 +204,7 @@ export class Crawler {
     | null = null;
 
   recording: boolean;
-  deduping = false;
+  isExternalDedupeStore = false;
 
   constructor() {
     const args = this.parseArgs();
@@ -346,7 +346,7 @@ export class Crawler {
     const redisUrl = this.params.redisStoreUrl || "redis://localhost:6379/0";
     const dedupeRedisUrl = this.params.redisDedupeUrl || redisUrl;
 
-    this.deduping = dedupeRedisUrl !== redisUrl;
+    this.isExternalDedupeStore = dedupeRedisUrl !== redisUrl;
 
     if (!redisUrl.startsWith("redis://")) {
       logger.fatal(
@@ -1629,7 +1629,7 @@ self.__bx_behaviors.selectMainBehavior();
   }
 
   async cleanupOnCancel() {
-    if (this.deduping) {
+    if (this.isExternalDedupeStore) {
       await this.crawlState.clearUncommitted();
     }
   }
@@ -1739,13 +1739,16 @@ self.__bx_behaviors.selectMainBehavior();
       this.storage = initStorage();
     }
 
-    if (this.params.generateWACZ && (this.storage || this.deduping)) {
+    if (
+      this.params.generateWACZ &&
+      (this.storage || this.isExternalDedupeStore)
+    ) {
       const filename = await this.crawlState.setWACZFilename();
-      if (this.deduping) {
+      if (this.isExternalDedupeStore) {
         await this.crawlState.addSourceWACZForDedupe(filename);
       }
     }
-    if (this.deduping) {
+    if (this.isExternalDedupeStore) {
       await this.crawlState.addUncommited();
     }
 
@@ -1949,7 +1952,7 @@ self.__bx_behaviors.selectMainBehavior();
       if (wacz) {
         await this.crawlState.clearWACZFilename();
 
-        if (this.deduping) {
+        if (this.isExternalDedupeStore) {
           await this.crawlState.updateDedupeSourceWACZ(wacz);
         }
 
@@ -1973,7 +1976,7 @@ self.__bx_behaviors.selectMainBehavior();
       return;
     }
 
-    if (this.deduping) {
+    if (this.isExternalDedupeStore) {
       // commit crawl data to main index
       logger.info("Committing dedupe index");
       await this.crawlState.commitDedupeDone();


src/util/recorder.ts
```diff
@@ -817,6 +823,25 @@ export class Recorder extends EventEmitter {
...
+      const { origUrl } = await this.crawlState.getHashDupe(hash);
+      if (origUrl) {
+        const errorReason = "BlockedByResponse";
+        await cdp.send("Fetch.failRequest", {
+          requestId,
+          errorReason,
+        });
+        return true;
+      }
+    }
```
tw4l: Help me understand how this works in relation to the revisits that we're otherwise writing - when do we want to write a revisit vs. block the page from loading? (might be good to add a comment here as well)

src/util/recorder.ts
```diff
@@ -1606,7 +1634,39 @@ export class Recorder extends EventEmitter {
...
+        returnPayloadOnlySize: true,
+      });
+    }
+
+    const hash = responseRecord.warcPayloadDigest || "";
+
+    // if (!(await this.crawlState.addIfNoDupe(WRITE_DUPE_KEY, url, hash))) {
+    //   serializer.externalBuffer?.purge();
+    //   return false;
+    // }
```
tw4l: Safe to remove?

src/util/redis.ts
```diff
@@ -34,6 +35,19 @@ export async function initRedis(url: string) {
   return redis;
 }
 
+export async function initRedisWaitForSuccess(redisUrl: string, retrySecs = 1) {
+  while (true) {
+    try {
+      return await initRedis(redisUrl);
+      break;
+    } catch (e) {
+      //logger.fatal("Unable to connect to state store Redis: " + redisUrl);
```
tw4l:
```suggestion
```

src/util/redis.ts
```diff
@@ -34,6 +35,19 @@ export async function initRedis(url: string) {
   return redis;
 }
 
+export async function initRedisWaitForSuccess(redisUrl: string, retrySecs = 1) {
+  while (true) {
```
tw4l, line 39: Wonder if we want to potentially add a timeout to this so it doesn't hang if redis can't resolve?

src/util/state.ts
```diff
@@ -185,7 +185,103 @@ export type SaveState = {
...
 
 // ============================================================================
-export class RedisCrawlState {
+export class RedisDedupIndex {
+  dedupRedis: Redis;
+
+  sourceDone = "src:d";
+  sourceQ = "src:q";
+  pendingQ = "pending:q";
+  sourceP = "src:p";
```
tw4l: Is `sourceP` used? I'm not seeing it.

src/util/recorder.ts
ikreymer: At this point, this is still fairly experimental, will add a comment, but idea is to short circuit page loading and write that in the pages.jsonl. But, perhaps should still write a revisit record, will check if that's happening.

src/util/state.ts
ikreymer: Good catch, removed!

src/util/redis.ts
ikreymer: Removed!

src/util/recorder.ts
ikreymer: Yep, done!

src/util/recorder.ts
ikreymer: Currently not writing a revisit record or recording in pages yet.. decided to comment this out for now, and leave in, so can come back to it. It's not quite ready to use yet, but can enable in a future update.
Can also just remove but maybe easier to comment out for now.

src/util/redis.ts
ikreymer: There's not much that crawler can do if it can't connect to either the dedupe or per-crawl redis so should just exit.
Could add that if restartsOnError and then just exit to allow k8s exponential backoff to kick-in.. but that could result in slower startup, but maybe worth it to save resources. Could add a max timeout of 30 or 60 seconds, before exponential backoff kicks in?

src/crawler.ts
```diff
@@ -343,32 +344,30 @@ export class Crawler {
 
   async initCrawlState() {
     const redisUrl = this.params.redisStoreUrl || "redis://localhost:6379/0";
+    const dedupeRedisUrl = this.params.redisDedupeUrl || redisUrl;
+
+    this.deduping = dedupeRedisUrl !== redisUrl;
```
tw4l: Is this a safe assumption? I thought that it was possible to use the same Redis URL for crawling and dedupe (in which case, this might be `false` when it should be `true`)

ikreymer: This was supposed to track if its an external dedupe redis, to determine if merging/committing crawls is needed.
Renamed the key to `isExternalDedupeStore` for clarity, may remove altogether later.

Mostly, it's there to optimize certain per-crawl actions that are not needed if not using external dedupe.

PR #892, opened 2025-10-07
flow behaviors: add scrolling into view

ikreymer:
Some page elements don't quite respond correctly if the element is not in view, so should add the setEnsureElementIsInTheViewport() to click, doubleclick, hover and change step locators.

Testing:
- This happens on the https://illuminateduniverse.org/ behavior at: https://github.com/webrecorder/custom-behaviors/blob/main/flow-behaviors/illuminated-universe/illuminated-universe.json
Without this change, the Older posts button stops being loaded, as the page is too far. This scrolls the page to each successive Older posts button.

### DIFF
--- a/src/util/flowbehavior.ts
+++ b/src/util/flowbehavior.ts
@@ -368,7 +368,7 @@ class Flow {
       case StepType.DoubleClick:
         await locator(step)
           .setTimeout(timeout * 1000)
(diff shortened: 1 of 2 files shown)


tw4l (approved):
Nice improvement! Tested in crawler with screencasting, working as expected.

PR #894, opened 2025-10-08
Profile Saving Improvements

ikreymer:
fix some observed errors that occur when saving profile:
- use browser.cookies instead of page.cookies to get all cookies, not just from page
- catch exception when clearing cache and ignore
- logging: log when proxy init is happening on all paths, in case error in proxy connection

### DIFF
--- a/src/util/browser.ts
+++ b/src/util/browser.ts
@@ -616,14 +616,14 @@ export class Browser {
     await page.setViewport(params);
   }
 
-  async getCookies(page: Page) {
-    return await page.cookies();
+  async getCookies() {
+    return (await this.browser?.cookies()) || [];
   }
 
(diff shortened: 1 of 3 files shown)


src/util/browser.ts
```diff
@@ -616,14 +616,14 @@ export class Browser {
...
 
-  async getCookies(page: Page) {
-    return await page.cookies();
+  async getCookies() {
+    return (await this.browser?.cookies()) || [];
   }
 
   // TODO: Fix this the next time the file is edited.
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   async setCookies(page: Page, cookies: any) {
```
tw4l: `page` arg seems unused now

ikreymer: fixed, thanks!

PR #903, opened 2025-10-26
add --saveProfile option to save profile after successful crawl

### LINKED ISSUE #898: [Feature] Option to auto-update profiles after a successful crawl.
For crawls that are run with a profile, it would be helpful to update the profile after a successful crawl, and either save the profile locally and/or re-upload to remote storage, just as if the profile were newly created.

This can help with maintaining updated session info after repeated crawling.

ikreymer:
- if --saveProfile is specified, attempt to save profile to same target as --profile
- if --saveProfile <target>, save to target
- save profile on finalExit if browser has launched
- supports local file paths and storage-relative path with '@' (same as --profile)
- also clear cache in first worker to match regular profile creation

fixes #898

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "browsertrix-crawler",
-  "version": "1.8.2",
+  "version": "1.9.0-beta.0",
   "main": "browsertrix-crawler",
   "type": "module",
   "repository": "https://github.com/webrecorder/browsertrix-crawler",


src/crawler.ts
```diff
@@ -1857,6 +1865,15 @@ self.__bx_behaviors.selectMainBehavior();
       }
     }
 
+    if (this.finalExit && generateFiles && this.params.saveProfile) {
+      console.log("SAVE PROFILE", this.params.saveProfile);
```
```suggestion
```

src/util/argParser.ts
```diff
@@ -848,6 +856,14 @@ class ArgParser {
       logger.info("Saving localStorage and sessionStorage");
     }
 
+    //remove default
+    if (!argv.saveProfile) {
```
ikreymer: TODO: remove this before merging, for testing in Browsertrix with option always enabled without any backend changes.

PR #907, opened 2025-11-05
fix: Skip proxy for seed file downloads in Self-Hosting environments

Percivalll:
When proxy is enabled in Self-Hosting deployments, seed file downloads may fail if the seed file URL is hosted on the internal network or local server. This change ensures that seed file downloads bypass the proxy configuration and connect directly to the source.

The proxy configuration (as documented in
https://docs.browsertrix.com/deploy/proxies/) is designed for crawling traffic, not for downloading seed files which are typically accessed from internal or local resources during the crawl initialization phase.

Changes:
- Modified collectOnlineSeedFile() to pass useProxy=false when downloading seed files from URLs
- Seed file downloads now use direct connection instead of proxy

### DIFF
--- a/src/util/file_reader.ts
+++ b/src/util/file_reader.ts
@@ -40,8 +40,9 @@ async function writeUrlContentsToFile(
   url: string,
   pathPrefix: string,
   pathDefaultExt: string,
+  useProxy: boolean = false,
 ) {
-  const res = await fetch(url, { dispatcher: getProxyDispatcher(url) });
+  const res = await fetch(url, { dispatcher: useProxy ? getProxyDispatcher(url) : undefined });
   const fileContents = await res.text();
 
   const filename =


src/util/file_reader.ts
```diff
@@ -40,8 +40,9 @@ async function writeUrlContentsToFile(
   url: string,
   pathPrefix: string,
   pathDefaultExt: string,
+  useProxy: boolean = true,
```
```suggestion
  useProxy: boolean = false,
```

I think we should actually just default this to false - this method is used to fetch data before the crawl starts, so should really not need to use the proxy here, but maybe good to leave as option for later.

src/util/file_reader.ts
```diff
@@ -54,7 +55,7 @@ async function writeUrlContentsToFile(
 
 export async function collectOnlineSeedFile(url: string): Promise<string> {
   try {
-    const filepath = await writeUrlContentsToFile(url, "seeds-", ".txt");
+    const filepath = await writeUrlContentsToFile(url, "seeds-", ".txt", false);
```
```suggestion
    const filepath = await writeUrlContentsToFile(url, "seeds-", ".txt");
```
Can leave unchanged if changing the default.

Issue #913, opened 2025-11-13
Huge slowdowns on epa.gov in v1.9.0


Mr0grog:
Unfortunately I don’t have a lot of detail here yet, but thought I’d open an issue in case there any obvious culprits or config changes I can/should make that might solve things. I will see what more detail I can uncover as time permits.

Yesterday, I upgraded [EDGI’s](https://envirodatagov.org) crawling to use browsertrix-crawler v1.9.0 (via the `webrecorder/browsertrix-crawler` Docker image), but then hit unworkable slowdowns on epa.gov URLs (other URLs may have also slowed down — I need to investigate — but didn’t slow down so much they outright broke things like epa.gov did). Specifically, we do `page`-scoped “crawls” of ~3,700 epa.gov URLs, portioned out over separate workers that each do 1k of the URLs. **It usually takes a worker 75–100 minutes to do these 1,000 URLs, but after upgrading to v1.9.0, these crawls started timing out at 6 hours, and having only covered about half (~550) of the URLs in that time.** Switching back to v1.8.1 succeeded in the same timeframe we used to have, so the slowdown is definitely do to something that changed in Browsertrix-crawler, not at epa.gov.

The most obvious potential culprits based on release notes look like updates to behaviors (#906, 74b6ad0ae0883229f344054f38dc886261b2d125) or Brave (#912), but I haven’t had time to look into any details.

The crawl configuration for these crawls is:

```yaml
warcinfo:
  operator: '"Environmental Data & Governance Initiative" <contact@envirodatagov.org>'
workers: 1
pageLoadTimeout: 120
rolloverSize: 8000000000
saveStateHistory: 1
scopeType: page
seeds:
- https://www.epa.gov/bedbugs
- https://www.epa.gov/chemical-data-reporting
# etc...
```

And being run with arguments:

```sh
docker run \
    --rm \
    --attach stdout --attach stderr \
    --volume "seeds.yaml:/app/config.yaml" \
    --volume "/path/to/data/dir:/crawls/" \
    "webrecorder/browsertrix-crawler:1.9.0" \
    crawl \
    --config /app/config.yaml \
    --collection "name-of-collection" \
    --saveState always
```

Mr0grog:
Looking at those logs, the really obvious thing is that v1.9.0 spends 26 seconds scrolling, while v1.8.1 logs "Skipping autoscroll, page seems to not be responsive to scrolling events".

[...]

PR #915, opened 2025-11-15
remove --disable-component-update flag, fixes shields not working

ikreymer:
should fix main cause of slow down in #913 
deps: update to brave 1.84.139, pptr 24.30.0
bump to 1.9.1


ikreymer:
> [brave/brave-core#30440](https://github.com/brave/brave-core/pull/30440) also introduced `--allow-brave-component-update`, which restores the previous behavior.
> 
> Passing both flags should register the components required for Shields, while excluding some other components:
> 
> ```
> '--allow-brave-component-update',
> '--disable-component-update'
> ```

Thanks! I missed that flag somehow when looking at that. Added both flags in #942

PR #916, opened 2025-11-18
netIdle cleanup + better default for pages where networkIdle timesout

ikreymer:
- set default networkIdle to 2
- add netIdleMaxRequests as an option, default to 1 (in case of long running requests)
- further fix for #913 
- avoid accidental logging


src/util/argParser.ts
```diff
@@ -518,7 +518,13 @@ class ArgParser {
           describe:
             "if set, wait for network idle after page load and after behaviors are done (in seconds). if -1 (default), determine based on scope",
```
```suggestion
            "number of seconds to wait for network idle after page load and after behaviors are done (default: 2)",
```

PR #917, opened 2025-11-19
better failure detection, allow update support for captcha detection via behaviors

ikreymer:
- allow fail on content check from main behavior
- update to behaviors 0.9.6 to support 'captcha_found' content check for tiktok
- allow throwing from timedRun
- call fatal() if profile can not be extracted

### DIFF
--- a/src/util/storage.ts
+++ b/src/util/storage.ts
@@ -167,6 +167,7 @@ export class S3StorageSync {
           {},
           "storage",
           true,
+          true,
         );
         break;
       } catch (e) {


src/util/timing.ts
```diff
@@ -14,8 +14,10 @@ export function timedRun(
   logDetails = {},
   context: LogContext = "general",
   isWarn = false,
+  throwOnTimeout = false,
```
tw4l, line 17: I'm not seeing this being used anywhere yet - is it just for future usage?

ikreymer: Good catch! Fixed!

PR #919, opened 2025-11-25
improvements to support pausing:

ikreymer:
- clear size to 0 immediately after wacz is uploaded
- if crawler is in paused, ensure upload of any data on startup
- fetcher q: stop queuing async requests if recorder is marked for stopping


tw4l (approved):
Thank you!

PR #921, opened 2025-11-26
Add downloads dir to cache external dependency within the crawl

### LINKED ISSUE #897: [Bug]: Request expired on seedfile
### Browsertrix Version

v1.19.1-e1f057e

Crawling large sites with huge premade seedlist, job fails after:

Image

Explanation from logfile:
Image

### Reproduction instructions

Add a 25mb seedlist and crawl it over long time (one crawlerwindow).

### Screenshots / Video

_No response_

### Environment

_No response_

### Additional details

_No response_

### LINKED ISSUE #920: [Change] Store downloaded inputs in the crawl directory instead of as temp files.
The crawler may download several files on input from URL, including:
- seed list
- profile
- custom behaviors

Currently, these are all placed in a temp directory, which means that upon a restart, the data is gone, and needs to be re-downloaded.
This is less than desirable, eg. seed list should not need to be re-fetched (see: #897), while profile and behaviors could be updated,
but crawl should be resumable with existing data, if present.

The crawler should store all of these in a new `/downloads/` directory in the crawler, and be able to restart (crawl from an existing crawl) if external resources are already present locally.

ikreymer:
Fixes #920 
- Downloads profile, custom behavior, and seed list to `/downloads` directory in the crawl
- Seed File: Downloaded into downloads. Never refetched if already exists on subsequent crawl restarts.
- Custom Behaviors: Git: Downloaded into dir, then moved to /downloads/behaviors/<dir name>. if already exist, failure to downloaded will reuse existing directory
- Custom Behaviors: File: Downloaded into temp file, then moved to /downloads/behaviors/<name.js>. if already exists, failure to download will reuse existing file.
- Profile: using `/profile` directory to contain the browser profile
- Profile: downloaded to temp file, then placed into /downloads/profile.tar.gz. If failed to download, but already exists, existing /profile directory is used
- Also fixes #897

### DIFF
--- a/src/util/browser.ts
+++ b/src/util/browser.ts
@@ -224,14 +224,13 @@ export class Browser {
           formatErr(e),
           "browser",
         );
-        this.customProfile = true;
-        return;
       } else {
         // remove the temp profile dir, likely empty
         await fsp.rm(tmpProfileDir, { recursive: true });
         logger.fatal("Profile setup failed", formatErr(e), "browser");
       }
     }
+    this.customProfile = true;
   }
 
   async loadProfile(


src/util/storage.ts
```diff
@@ -176,12 +176,7 @@ export class S3StorageSync {
           await sleep(5);
           logger.warn("Retry downloading profile", {}, "storage");
         } else {
-          logger.fatal(
-            "Could not download profile, exiting",
-            {},
-            "storage",
-            ExitCodes.Failed,
-          );
+          throw new Error("profile count not be downloaded");
```
ikreymer:
```suggestion
          throw new Error("Profile could not be downloaded");
```

src/util/file_reader.ts
```diff
@@ -113,30 +184,53 @@ async function collectGitBehaviors(gitUrl: string): Promise<FileSources> {
     pathToCollect = path.join(tmpDir, relPath);
   }
 
+  // Download behaviors to temp dir
```
```suggestion
  // Download behaviors to temp dir (in downloads directory)
```

src/util/browser.ts
```diff
@@ -190,61 +188,98 @@ export class Browser {
...
+      // replace old tarball with new tarball
+      await fsp.rename(tmpProfileDest, profileTarGz);
+    } catch (e) {
+      if (exists) {
+        logger.warn(
+          "Error updating profile, using existing profile",
+          formatErr(e),
+          "browser",
+        );
+        this.customProfile = true;
```
tw4l: I think this PR may have removed setting `this.customProfile` in the happy path, looks like it's now only set when an error is thrown here

src/util/file_reader.ts
```diff
@@ -25,39 +24,94 @@ export type FileSource = {
...
+    return targetFile;
+  }
+
+  try {
+    const res = await fetch(url, {
+      dispatcher: useProxy ? getProxyDispatcher(url) : undefined,
+    });
+    if (!res.ok) {
+      throw new Error(`Invalid response, status: ${res.status}`);
+    }
```
tw4l, line 86: Arguably out of scope/not a new problem, but we should probably really be doing retries here

src/util/file_reader.ts
```diff
@@ -113,30 +184,53 @@ async function collectGitBehaviors(gitUrl: string): Promise<FileSources> {
     pathToCollect = path.join(tmpDir, relPath);
   }
 
+  // Download behaviors to temp dir (in downloads directory)
   try {
```
tw4l, line 188: Another place we should probably be doing retries in case of network interruption, etc.

src/util/file_reader.ts
```diff
@@ -113,30 +184,53 @@ async function collectGitBehaviors(gitUrl: string): Promise<FileSources> {
...
+    }
   }
-  return [];
+
+  await replaceDir(pathToCollect, behaviorsDir, exists);
+
+  // remove the rest of the repo that we're not using
+  if (relPath) {
+    await fsp.rm(tmpDir, { recursive: true, force: true });
+  }
```
tw4l, line 217: Nice addition

src/util/file_reader.ts
ikreymer: we do want to check this, since its just getting the text, it'll succeed below, and possibly overwrite a valid seed list/behavior with the contents of a 404 page

ikreymer: But agreed on retries, that should be a follow-up!

src/util/browser.ts
ikreymer: Good catch, fixed!

Issue #928, opened 2025-12-01
Page Load Failed -> Docker Exit Code 10 for nike.com


gitreich:
Seed:    https://www.nike.com
Result: No archive file written
Docker Exit Code 10, Crawler seems to hit every possible Timeout

Expected Result:  Crawl of Page as it looks very valid in the browser

Reproduceable minimum example:

`
 docker run -d --name test_nike_secure -v /home/user/browsertrix/crawls/:/crawls/    webrecorder/browsertrix-crawler:1.9.2 crawl  --scopeType page --depth 0  --headless  --delay 0 --behaviorTimeout 60 --pageLoadTimeout 60 --waitUntil networkidle0  --saveState always   --logging stats,info   --sitemap  --url https://www.nike.com --collection nike_com
 `

Logs:

```
{"timestamp":"2025-12-01T12:05:05.824Z","logLevel":"info","context":"general","message":"Browsertrix-Crawler 1.9.2 (with warcio.js 2.4.7)","details":{}}
{"timestamp":"2025-12-01T12:05:05.825Z","logLevel":"info","context":"general","message":"Seeds","details":[{"url":"https://www.nike.com/","scopeType":"page","include":[],"exclude":[],"allowHash":false,"depth":-1,"sitemap":"<detect>","auth":null,"_authEncoded":null,"maxExtraHops":0,"maxDepth":0}]}
{"timestamp":"2025-12-01T12:05:05.825Z","logLevel":"info","context":"general","message":"Link Selectors","details":[{"selector":"a[href]","extract":"href","isAttribute":false}]}
{"timestamp":"2025-12-01T12:05:05.825Z","logLevel":"info","context":"general","message":"Behavior Options","details":{"message":"{\"autoplay\":true,\"autofetch\":true,\"autoscroll\":true,\"siteSpecific\":true,\"log\":\"__bx_log\",\"startEarly\":true,\"clickSelector\":\"a\"}"}}
{"timestamp":"2025-12-01T12:05:05.862Z","logLevel":"info","context":"sitemap","message":"Fetching sitemap","details":{"from":"<any date>","to":"<any date>"}}
{"timestamp":"2025-12-01T12:05:06.217Z","logLevel":"info","context":"sitemap","message":"Sitemap partially parsed, continue parsing large sitemap in the background","details":{"urlsFound":100}}
[...]

PR #930, opened 2025-12-02
sitemapper refactor to fix concurrency:

ikreymer:
- original implementation did not actually wait for sitemap to complete before queuing new ones, resulting in concurrency resource leak
- refactor to await completion of sitemap parser, replacing peding list with counter
- also, don't parse sitemap if single-page and no extra hops!
- fixes issues in #928

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -2714,8 +2714,17 @@ self.__bx_behaviors.selectMainBehavior();
       return;
     }
 
+    if (
+      (this.params.scopeType === "page" ||
+        this.params.scopeType === "page-spa") &&
+      !this.params.extraHops
+    ) {
+      logger.info("Single page crawl, skipping sitemap", {}, "sitemap");
+      return;
+    }
+
     if (await this.crawlState.isSitemapDone()) {
-      logger.info("Sitemap already processed, skipping", "sitemap");
+      logger.info("Sitemap already processed, skipping", {}, "sitemap");
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/util/sitemapper.ts
```diff
@@ -191,8 +189,10 @@ export class SitemapReader extends EventEmitter {
         { fullUrl, seedUrl },
         "sitemap",
       );
-      this._parseSitemapFromResponse(fullUrl, resp);
+      await this.parseSitemapFromResponse(fullUrl, resp);
```
ikreymer, line 192: This was the issue - the parseSitemapFromResponse would start parsing, driven by callbacks, and would not wait for parsing to complete before moving on to next sitemap. Now passing down resolve/reject and awaiting a promise via the events..

src/util/sitemapper.ts
```diff
@@ -218,23 +218,61 @@ export class SitemapReader extends EventEmitter {
...
+      resolve = res;
+      reject = rej;
+    });
+
+    this.pending++;
+
+    try {
+      this.doParseSitemapFromResponse(url, resp, resolve!, reject!);
+
+      await promise;
```
ikreymer, line 246: Here's where we await for the parsing to actually be done...

src/crawler.ts
```diff
@@ -2714,8 +2714,17 @@ self.__bx_behaviors.selectMainBehavior();
...
     }
 
+    if (
+      (this.params.scopeType === "page" ||
+        this.params.scopeType === "page-spa") &&
+      !this.params.extraHops
+    ) {
+      logger.info("Single page crawl, skipping sitemap", {}, "sitemap");
+      return;
+    }
```
tw4l, line 2724: Good addition, thanks

tw4l (approved):
Tested and seems to be working well. Thanks for the really helpful comments explaining the issue and solution :)

PR #932, opened 2025-12-02
Rename robots flag to --useRobots, keep --robots as alias

tw4l:
Follow-up to https://github.com/webrecorder/browsertrix-crawler/issues/631

Based on feedback from https://github.com/webrecorder/browsertrix/pull/3029

Renaming `--robots` to `--useRobots` will allow us to keep the Browsertrix backend API more consistent with similar flags like `--useSitemap`. Keeping `--robots` as it's a nice shorthand alias.


ikreymer (approved):
Makes sense. Also updated the cli-options.md

PR #933, opened 2025-12-05
Sitemaps: parse /sitemap.xml if no sitemap listed in robots.txt

ikreymer:
- Ensure the /sitemap.xml is parsed even if robots.txt exists, but no sitemaps listed there.
- Resolve relative URLs listed in robots.txt, eg. 'Sitemap: /my-sitemap.xml'
- Simplify sitemap detection logic, check robots first, then sitemap.xml OR alternate url if provided via --useSitemap <url>
- Have two main methods, parseSitemap() and parseSitemapFromRobots() that handle the parsing.
- follow-up to #930

### DIFF
--- a/src/util/sitemapper.ts
+++ b/src/util/sitemapper.ts
@@ -73,7 +73,7 @@ export class SitemapReader extends EventEmitter {
         const ct = resp.headers.get("content-type");
         if (expectedCT && ct && !expectedCT.includes(ct.split(";")[0])) {
           logger.debug(
-            "Not loading sitemap: invalid content-type"
+            "Not loading sitemap: invalid content-type",
             { ct },
             "sitemap",
           );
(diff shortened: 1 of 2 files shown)


src/util/sitemapper.ts
```diff
@@ -62,14 +62,23 @@ export class SitemapReader extends EventEmitter {
...
       const resp = await fetch(url, {
         headers: this.headers,
         dispatcher: getProxyDispatcher(url),
       });
 
       if (resp.ok) {
+        const ct = resp.headers.get("content-type");
+        if (expectedCT && ct && !expectedCT.includes(ct.split(";")[0])) {
+          logger.debug(
+            "Loading sitemap: invalid content-type",
```
tw4l:
```suggestion
            "Not loading sitemap: invalid content-type"
```

Maybe good to make it clearer that this means we won't use this sitemap

src/util/sitemapper.ts
```diff
@@ -85,148 +94,95 @@ export class SitemapReader extends EventEmitter {
...
-      return null;
-    }
-  }
-
-  async tryFetch(url: string, expectedCT?: string[] | null) {
-    try {
       logger.debug(
-        "Detecting Sitemap: fetching",
-        { url, expectedCT },
+        "Loading Sitemap: invalid status code",
```
tw4l:
```suggestion
        "Not loading sitemap: invalid status code",
```
Same comment as above

src/util/sitemapper.ts
```diff
@@ -85,148 +94,95 @@ export class SitemapReader extends EventEmitter {
...
-        resp = await this.tryFetch(fullUrl, XML_CONTENT_TYPES);
-        if (resp) {
-          isSitemap = true;
-        }
+      const robotsUrl = new URL("/robots.txt", seedUrl).href;
+      const done = await this.parseRobotsForSitemap(robotsUrl);
+
+      if (!done) {
+        const sitemapUrl = new URL("/sitemap.xml", seedUrl).href;
+        await this.parseSitemap(sitemapUrl);
```
tw4l: We should probably also log a `"Sitemap not found"` message to debug here if auto-detect doesn't find a sitemap via /robots.txt or /sitemap.xml, like we do below for when a specific sitemap URL is provided

PR #935, opened 2025-12-05
sort query args before queuing URLs

ikreymer:
- use 'normalize-url' package to avoid differently sorted query args that are the same url
- configure other options, such as keeping www. and trailing slashes, only using this for query arg sorting

### DIFF
--- a/src/util/state.ts
+++ b/src/util/state.ts
@@ -11,7 +11,7 @@ import {
 import { ScopedSeed } from "./seeds.js";
 import { Frame } from "puppeteer-core";
 import { interpolateFilename, UploadResult } from "./storage.js";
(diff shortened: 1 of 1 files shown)


src/util/state.ts
```diff
@@ -11,6 +11,7 @@ import {
 import { ScopedSeed } from "./seeds.js";
 import { Frame } from "puppeteer-core";
 import { interpolateFilename, UploadResult } from "./storage.js";
+import normalizeUrl, { Options } from "normalize-url";
```
tw4l: Not a big deal but Options is a pretty generic name, maybe we should do something like `{ Options as NormalizeUrlOptions }` here?

PR #942, opened 2025-12-24
follow-up to #915, add --allow-brave-component-update flag

ikreymer:
can still disable most other component updates, but enable brave components, including shields, based on:
https://github.com/webrecorder/browsertrix-crawler/pull/915#issuecomment-3689772878


ikreymer:
Tested profile creation to ensure shields are in fact working

PR #943, opened 2025-12-28
Replace minio client with aws client-s3 + lib-storage for multi-part upload

ikreymer:
Extends work in #547 adds Upload via @aws-sdk/lib-storage library:
- Replaces minio client with official aws s3 client
- Uses @aws-sdk/lib-storage for multi-part upload support

Testing:
- This should address issues from #479 and webrecorder/browsertrix#2925 ideally
- Tested with all the major S3 implementations: VersityGW, RustFS, SeaweedFS, Garage as well as Minio


ikreymer:
@mguella @wvengen @HeliosLHC can you test to see if this implementation meets your needs and works with the different S3 providers that you were using, and hopefully addresses the issues in #479 also (if its still relevant)?
At this point, it makes sense to move away from minio client libraries, given the minio is no longer being maintained.

ikreymer:
Tested latest changes with VersityGW, RustFS and SeaweedFS and existing Minio

src/util/storage.ts
```diff
@@ -128,58 +121,56 @@ export class S3StorageSync {
...
 
-    // for backwards compatibility, keep 'bytes'
-    return { path, size, hash, bytes: size };
+    await uploader.done();
   }
 
   async downloadFile(srcFilename: string, destFilename: string) {
-    let count = 0;
     logger.debug("Downloading profile", { srcFilename }, "storage");
-    while (true) {
```
ikreymer, line 158: Retries handled by the S3Client itself, with maxAttempts set in constructor

src/util/storage.ts
```diff
@@ -128,58 +121,56 @@ export class S3StorageSync {
...
+    const size = await getFileSize(srcFilename);
+
+    // for backwards compatibility, keep 'bytes'
+    return { path, size, hash, bytes: size };
+  }
+
+  private async doUpload(targetFilename: string, Body: Readable) {
+    const Bucket = this.bucketName;
+
+    const Key = this.objectPrefix + targetFilename;
```
tw4l, line 141: Why the upper-cased `const`s here?

ikreymer: just for clarity, so can pass in `params: { Bucket, Key, Body }` below

tw4l:
Tested locally and on dev with large crawls. After adding `requestChecksumCalculation: "WHEN_REQUIRED"` to S3 client config in https://github.com/webrecorder/browsertrix-crawler/pull/943/commits/9cae2905e0a5e7959eb904d4df68cd06c16d08ce, appears to be working well. The AWS s3 client library authors seem uninterested in preventing breaking changes with other s3-compatible providers so there is a chance that we could have future issues like that one where new AWS features introduce breaking changes, but after looking at the options it still seems like the best library available to us.

Issue #923, opened 2025-11-27
Connection leak


ato:
We found some long running crawls were eventually exhausting all available TCP ports on the host running browsertrix-crawler causing every process trying to connect to something to fail with "Cannot assign requested address".

How to reproduce:

```
podman run webrecorder/browsertrix-crawler:1.9.2 crawl --url https://www.library.gov.au/
```

Then watch the number of TCP sockets of the browsertrix-crawler node process slowly grow over time:

```bash
pid=$(pgrep ^node$)

sudo lsof -p $pid

# or enter the network namespace so lsof can see the addresses
sudo nsenter -n -t $pid lsof -p $pid
```

In my testing browsertrix-crawler 1.7.1 immediately closes the sockets so it seems to have been introduced in 1.8.0.

PR #946, opened 2026-01-02
Replace fetch() with optimized undici request()

ikreymer:
Per https://undici.nodejs.org/#/?id=benchmarks request() is supposed to be much more performant compared to fetch() with almost the same interface.

This PR replaces all of the fetch() calls (both using proxy dispatcher and regular fetch) with undici request() 
The migration is fairly simply as shown in https://undici.nodejs.org/#/?id=migration-guide

The migration eliminates various web streams to node streams conversions.

Also updates to latest undici

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -1841,6 +1841,7 @@ class AsyncFetcher {
     dispatcher = dispatcher.compose((dispatch) => {
       return (opts, handler) => {
         if (opts.headers) {
+          // store full actual headers that are sent for the request
           reqresp.requestHeaders = opts.headers as Record<string, string>;
         }
         return dispatch(opts, handler);


src/util/recorder.ts
```diff
@@ -1813,10 +1814,9 @@ class AsyncFetcher {
   }
 
   async doCancel() {
-    const { abort } = this;
-    if (abort) {
-      abort.abort();
-      this.abort = undefined;
+    const { body } = this;
+    if (body) {
+      body.destroy();
```
tw4l, line 1821: Is this sufficient to cancel the request the same way it did previously? Was doing a bit of reading in https://github.com/nodejs/undici/issues/3353 but it's a bit hard to tease apart.

src/util/recorder.ts
```diff
@@ -1840,23 +1840,32 @@ class AsyncFetcher {
...
+    //this.abort = new AbortController();
 
-    const resp = await fetch(url!, {
-      method,
+    const resp = await request(url!, {
+      method: (method || "GET") as Dispatcher.HttpMethod,
       headers,
       body: reqresp.postData || undefined,
-      redirect: this.manualRedirect ? "manual" : "follow",
+      //redirect: this.manualRedirect ? "manual" : "follow",
```
tw4l: I assume we still want to be able to set the redirect policy following this change?

tw4l (commented):
In general seems good, and the performance advantage of switching from `fetch()` to `request()` seems worth it, but left a few in-line questions. Happy to chat on Discord if that's easier.

src/util/recorder.ts
ikreymer: yea, i think that's destroy() works same as abort, but perhaps simpler?
See section in: https://undici.nodejs.org/#/docs/api/Dispatcher?id=example-2-aborting-a-request that says destroy() is equivalent.

Also tested active connections as I did with fix for #923 to ensure the connections are not leaking with this.

src/util/recorder.ts
ikreymer: Good catch, was left as placeholder to fix - fixed by setting `maxRedirections`

src/util/recorder.ts
tw4l: That's great! And thanks for the docs link :)

src/util/recorder.ts
tw4l: It seems like `maxRedirections` isn't an option on the `request` method - the docs show a few ways of setting this, but the most straightforward seems like the `redirect` interceptor: https://undici.nodejs.org/#/docs/api/Dispatcher?id=redirect

PR #955, opened 2026-01-29
ensure redirects are followed for sitemap, robots, other requests converted from fetch()

### LINKED ISSUE #954: Default to following redirects for most fetches
The change to undici.request() #946 resulted in a regression where requests that were following redirects (default with fetch()) are no longer, as that's not the default behavior with request().

Should default to same semantics, unless disabled (eg. in recorder)

ikreymer:
Refactor dispatcher apis:
- Use `getProxyDispatcher(withRedirect = true)` to follow redirects by default, with option to disable (eg. in recorder). This dispatcher also ignores TLS errors, to match current browser config. Used for fetching archival content.
- Use `getFollowRedirectsDispatcher()` with follows redirects but does not ignore TLS and does not use proxies, for fetching non-archival configs (profiles, behaviors, etc...)

Fixes #954, regression from #946 

Testing:
- run crawler on a seed that redirects, ex:
`docker run --rm webrecorder/browsertrix-crawler crawl --url https://eff.org/ --useSitemap  --logging debug --context sitemap`
- without this PR, will see errors like: `"Not loading sitemap: invalid status code","details":{"status":301}}`
- with this PR, the redirect should be followed and sitemap loaded correctly.

### DIFF
--- a/src/util/proxy.ts
+++ b/src/util/proxy.ts
@@ -48,7 +48,7 @@ const contentDispatcher = new Agent(contentAgentOpts);
 const contentRedirectDispatcher = addRedirectInterceptor(contentDispatcher);
 
 // dispatcher for following redirects, non-archival content, trust SSL
-const followRedirectDispatcher = new Agent(baseOpts);
+const followRedirectDispatcher = addRedirectInterceptor(new Agent(baseOpts));
 
 export type ProxyServerConfig = {
   matchHosts?: Record<string, string>;


src/util/proxy.ts
```diff
@@ -17,13 +17,39 @@ const SSH_PROXY_LOCAL_PORT = 9722;
...
+};
+
+// dispatcher for archival content without following redirects
+const contentDispatcher = new Agent(contentAgentOpts);
+
+// dispatcher for archival content with following redirects
+const contentRedirectDispatcher = addRedirectInterceptor(contentDispatcher);
+
+// dispatcher for following redirects, non-archival content, trust SSL
+const followRedirectDispatcher = new Agent(baseOpts);
```
tw4l: Without the redirect interceptor added here, what is making this dispatcher follow redirects?

ikreymer: Good catch! Fixed!

PR #962, opened 2026-02-02
Fix default user-agent to not include minor version + set sec-ua-ch-* headers

ikreymer:
- Only use major version from, set rest to 0.0.0 to match Brave/Chrome behavior
- Store major version in Browser
- Also set some `sec-ua-ch-*` headers to match brave default
- Don't disable cache when creating profile to avoid sending different Cache-Control headers when in profile creation (cache cleared before profile created anyway)

### DIFF
--- a/src/create-login-profile.ts
+++ b/src/create-login-profile.ts
@@ -236,7 +236,6 @@ async function main() {
 
   const waitUntil: PuppeteerLifeCycleEvent = "load";
 
-
   if (!params.automated) {
     await browser.setupPage({ page, cdp });
 


src/crawler.ts
```diff
@@ -328,13 +328,12 @@ export class Crawler {
...
-    if (!this.emulateDevice.userAgent) {
-      this.emulateDevice.userAgent = this.browser.getDefaultUA();
-    }
-
     // suffix to append to default userAgent
     if (this.params.userAgentSuffix) {
+      // get default if not set
+      if (!this.emulateDevice.userAgent) {
+        this.emulateDevice.userAgent = this.browser.getDefaultUA();
+      }
```
emma-sg: Bit of a simpler way to do this!

```suggestion
      this.emulateDevice.userAgent ||= this.browser.getDefaultUA();
```

src/util/browser.ts
```diff
@@ -679,18 +686,25 @@ export class Browser {
...
-        // TODO: Fix this the next time the file is edited.
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        await page.emulate(device as any);
-      } else if (device.userAgent) {
-        await page.setUserAgent(device.userAgent);
+      if (device.viewport) {
+        await page.setViewport(device.viewport);
+      }
+      if (device.userAgent) {
+        // set via emulation domain for, may be more complete
```
ikreymer:
```suggestion
        // set via Emulation instead of Network, may be more complete override
```

src/create-login-profile.ts
```diff
@@ -236,7 +236,7 @@ async function main() {
 
   const waitUntil: PuppeteerLifeCycleEvent = "load";
 
-  await page.setCacheEnabled(false);
+  //await page.setCacheEnabled(false);
```
ikreymer:
```suggestion
```

PR #963, opened 2026-02-03
Fix browser network loading

### LINKED ISSUE #960: Crawling does not fetch all images of srcset
I am using browsertrix-crawler:1.7.0 in a docker container.
On my local windows machine I have no problems and all images are fetched, 
but when I try to crawl on a onpremise machine only images fitting the breakpoint for the default windowSize 1360,1020 are fetched.

The images per breakpoint are defined in a srcset.

Do you have a clue, why the behaviour deviates and know a fix, so that all images are crawled?

ikreymer:
When available, async fetch should try to load via the browser network, especially for in-page discovered URLs, to ensure proper credentials are used (may fix #960):

- adds missing CDP param that resulted in browser network being skipped!
- try browser network for direct fetch too, if page is available, but then fallback to node fetch
- default to node fetch when network loading failed, or if in browser (non-page) context for request interception, eg. in a worker.

This complements change webrecorder/browsertrix-behaviors#117 that prefers in-browser fetch when possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@novnc/novnc": "1.4.0",
     "@puppeteer/replay": "^3.1.3",
     "@webrecorder/wabac": "^2.24.5",
-    "browsertrix-behaviors": "^0.9.7",
+    "browsertrix-behaviors": "^0.9.8",
     "client-zip": "^2.4.5",
     "css-selector-parser": "^3.0.5",
     "fetch-socks": "^1.3.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2744,10 +2744,10 @@ browserslist@^4.24.0:
     node-releases "^2.0.18"
     update-browserslist-db "^1.1.1"
 
-browsertrix-behaviors@^0.9.7:
-  version "0.9.7"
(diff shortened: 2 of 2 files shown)


src/crawler.ts
```diff
@@ -1201,6 +1201,11 @@ self.__bx_behaviors.selectMainBehavior();
     opts.pageBlockUnload = false;
 
     await this.awaitPageExtraDelay(opts);
+
+    // wait for any browser network fetches to finish before moving on to another page
+    if (recorder) {
+      await recorder.browserFetchQ.onIdle();
+    }
```
tw4l, line 1208: Since `crawlPage` has the possibility of timing out before we reach here, should we also be checking when we first start to crawl a new page that the queue is empty?

ikreymer: Hm, I don't think so, as that would eat into the time for that page, as the queue may not be empty.
What will happen is the page-related loading will fail with 'execution context destroyed' errors if they're tied to the previous page, though any not-yet-started requests will be retried async with node fetch anyway, which is similar to current behavior (so may not succeed), but we can avoid those if there is time to wait still allotted for current page.

PR #964, opened 2026-02-05
frame behaviors: use frame.evaluate() instead of custom evaluteWithCLI()

ikreymer:
Removes our custom evaluateWithCLI() call in favor of using standard `frame.evaluate()`.
The custom method became more tricky to use, and doesn't work for all iframes.
The main benefit of it was to inject `getEventListeners()`, which is now only used in Autoscroll to potentially skip scrolling, and not needed for any other behaviors.

We could potentially use this workaround: https://stackoverflow.com/questions/75517220/geteventlistener-function-support-for-latest-puppeteer-versions/75581410#75581410 to bring back `getEventListeners` or, adding a custom callback, like `hasEventListener()` would suffice for that check.

This simplifies the codebase and ensures that running behaviors is more reliable.
Also, adds a callback for new frames that may get added to page, such as during scrolling, an ensures behaviors are called on those iframes as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@novnc/novnc": "1.4.0",
     "@puppeteer/replay": "^3.1.3",
     "@webrecorder/wabac": "^2.24.5",
-    "browsertrix-behaviors": "^0.9.7",
+    "browsertrix-behaviors": "^0.9.8",
     "client-zip": "^2.4.5",
     "css-selector-parser": "^3.0.5",
     "fetch-socks": "^1.3.2",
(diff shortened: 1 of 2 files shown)


ikreymer:
> Tested with autoscroll and https://raw.githubusercontent.com/webrecorder/custom-behaviors/refs/heads/main/behaviors/timeline.js. I can see log messages indicating that behavior scripts are run and finished, but the actual behaviors themselves don't seem to happen - in the screencast I don't see any scrolling happening on pages where it should, and for the custom behavior I don't see the behavior running in the screenshare or any log lines from the custom behavior as expected.

Good catch, fixed, removing unnecessary function wrapping which didn't actually await the promise..

PR #966, opened 2026-02-05
Add option to write JSONL file with data on skipped pages

### LINKED ISSUE #965: Add option to produce report of skipped pages
A someone common reporting/QA requirement we've seen is to be able to list URLs that were not crawled because they were out of scope or not queued for other reasons, such as a limit being hit.

In order to meet this requirement, we should have an option to generate a page list like `extraPages.jsonl` for URLs that were not queued.

We may additionally want to log URLs that were queued but never crawled if the crawler gracefully stops (e.g. due to hitting a limit) as well.

tw4l:
Fixes #965 

Add `--reportSkipped` argument, which will enable the creation of a `reports/skippedPages.jsonl` file with the following elements for each URL encountered that was not queued:

- `url`
- `seedUrl`
- `depth`
- `reason` (one of `outOfScope`, `pageLimit`, `robotsTxt`, or `redirectToExcluded`)
- `ts`

The `reports/` directory is new and will likely be expanded with other crawl-time reporting moving forward.


tests/skipped_pages.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    pageCount++;
+
+    const validReasons = ["outOfScope", "pageLimit", "robotsTxt"];
+
+    expect(data).toHaveProperty("url");
+    expect(data).toHaveProperty("seedUrl");
+    expect(data).toHaveProperty("depth");
+    expect(data).toHaveProperty("reason");
+    expect(validReasons.includes(data.reason)).toBe(true);
+    expect(data).toHaveProperty("ts");
```
ikreymer, line 45: Since the page is skipped, it isn't expected to have a `ts` property, I don't think

tw4l: You're right it wasn't getting written, but just because of a bug. It's now generating a ts at the time the URL is written into `skippedPages.jsonl`. ts is one of only two required fields in a pages JSONL file [as we defined them in the WACZ spec](https://specs.webrecorder.net/wacz/1.1.1/#pages-jsonl), and it may also be helpful in looking at the report to see when various URLs were encountered.

ikreymer: ah, makes sense!

ikreymer (approved):
Made a few more changes:
- added test for pageLimit skipped page
- added redirectToExcluded skip reason and add test

Should be good to go now!

PR #970, opened 2026-02-11
add decompress() interceptor, support undici.request() without decompression + keep content-encoding if no decompression

### LINKED ISSUE #971: [Bug] Regression from undici: need to add decompress interceptor to match existing decompression behavior
undici.request() does not auto-decompress content-encoding, which is good for recording original payload, however, the handler currently assumes that content-encoding and transfer-encoding is removed! To match existing behavior, the undici handlers should be wrapped in a decompressing interceptor.

ikreymer:
Fixes #971 
- Add decompression interceptor as default for getProxyDispatcher(), no need to decompress sitemap explicitly
- Add option to not decompress, keep content-encoding header and don't add `x-orig-` when using node async fetch
- Create three dispatcher variants per proxy: redirect + decompress, redirect + no decompress, no redirect + no decompress

### DIFF
--- a/src/util/proxy.ts
+++ b/src/util/proxy.ts
@@ -20,7 +20,7 @@ const SSH_WAIT_TIMEOUT = 30000;
 type ProxyEntry = {
   proxyUrl: string;
   dispatcher: Dispatcher;
-  decompressDispatcher: Dispatcher;
+  decompressRedirectDispatcher: Dispatcher;
   redirectDispatcher: Dispatcher;
 };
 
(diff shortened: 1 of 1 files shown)


tw4l (approved):
Tested again and we do seem to be getting the expected results and fixed replay.

I do find it a little odd that request bodies may be decompressed or not depending on whether it is handled through the browser or through async fetch (which itself might just be determined by timing of the page in the crawler), but given that starting point, the solution you've landed on here makes sense to me.

PR #972, opened 2026-02-12
Add OpenContainers labels to Dockerfile to support Dependabot

Mr0grog:
I recently made some changes to how I depend on Browsertrix Crawler’s Docker image (through a slightly customized Dockerfile) that had the nice side-effect of Dependabot being able to provide automated updates. However, those updates come with no release notes or other information (integrating those right in the PR description is, IMO, one of Dependabot’s nicest features):

Screenshot of Dependabot PR to update Browsertrix Crawler that is missing release notes

It turns out that [Dependabot](https://github.blog/changelog/2023-04-13-dependabot-now-supports-fetching-release-notes-and-changelogs-for-docker-images/), [Renovate](https://docs.renovatebot.com/modules/datasource/docker/), etc. find this info by following the `org.opencontainers.image.source` label in an image, so this PR adds that. While I was at it, I figured I’d throw in a few other [standardized OCI labels](https://specs.opencontainers.org/image-spec/annotations/) that seemed obvious, although there might be others you want to add!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,11 @@ ARG BROWSER_IMAGE_BASE=webrecorder/browsertrix-browser-base:brave-${BROWSER_VERS
 
 FROM ${BROWSER_IMAGE_BASE}
 
+LABEL org.opencontainers.image.vendor="Webrecorder <https://webrecorder.net/>"
+LABEL org.opencontainers.image.source="https://github.com/webrecorder/browsertrix-crawler"
+LABEL org.opencontainers.image.documentation="https://crawler.docs.browsertrix.com/"
+LABEL org.opencontainers.image.licenses="AGPL-3.0-or-later"
+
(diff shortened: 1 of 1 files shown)


ikreymer:
Sure, happy to add these if its helpful!

Mr0grog:
Circling back here with a final comment/brief update:

It turns out this didn’t work because the Dependabot docs are not clear on an important detail: *this only works for packages published to the GitHub package/container registry (`ghcr.io`), and not for other registries like Docker Hub.* https://github.com/dependabot/dependabot-core/issues/15417

I sure wasted a lot of time testing various setups before I figured that out!

I’ll work up another PR to set the existing labels up better (e.g. the issue noted above with `version` bleeding through from the underlying Ubuntu image), but it turns out there’s nothing we can do here to get Dependabot to show release notes. 😞

PR #973, opened 2026-02-15
Interrupt instead of fail crawl when not fatal

ikreymer:
- Add logger.interrupt() which results in crawler being restarted, but without failing the crawl, unlike logger.fatal()
- Switch a few logger.fatal() -> logger.interrupt(), when crawl could be retried, only use fatal() when initial crawl config is invalid for the most part
- Don't increment 'failing' and set to 'failed' after unexpected errors, set to interrupted as well to allow for retry (and exponential backoff if in k8s)
- Allow logger to manage opening/closing its open external log file, via setOutputFile() and closeLog()
- Move closeLog() and setStatusAndExit() to logger itself so logger.interrupt() and logger.fatal() can call these
- Use the newer logger.interrupt() for: proxy connection unavailable, out of disk space, interrupt signal and WACZ upload failed

### DIFF
--- a/src/util/logger.ts
+++ b/src/util/logger.ts
@@ -279,12 +279,7 @@ class Logger {
         await this.crawlState.setStatus(status);
       }
     } catch (e) {
-      this.logAsJSON(
-        "Error shutting down, exiting anyway",
-        e,
-        this.defaultLogContext,
-        "error",
-      );
+      this.error("Error shutting down, exiting anyway", e);
     } finally {
       process.exit(exitCode);
     }


tw4l (commented):
Makes sense. I pushed a commit to improve typing for log levels so that we can also list the options in the CLI arg help text, especially since we've gone further off convention now.

Seems like a safe change for Browsertrix Crawler users, since no exit codes will change. What's unclear to me is the effect this will have in Browsertrix - I think without related changes in that codebase, we might just get the crawler stuck in an endless loop of retrying at this point?

src/util/logger.ts
```diff
@@ -225,21 +236,56 @@ class Logger {
...
+      if (this.crawlState && status) {
+        await this.crawlState.setStatus(status);
+      }
+    } catch (e) {
+      this.logAsJSON(
+        "Error shutting down, exiting anyway",
+        e,
+        this.defaultLogContext,
+        "error",
+      );
```
tw4l:
```suggestion
      this.error("Error shutting down, exiting anyway", e);
```

PR #975, opened 2026-02-17
Fix streaming response / retry mechanism when loading from browser

ikreymer:
A previous refactor broke the fetchResponseBody() streaming option, as a result this function always fails with 'invalid interception id'. (Mentioned in https://github.com/webrecorder/browsertrix-crawler/issues/960#issuecomment-3932156590) This function is called to read the response as a stream when the response size is unknown and the resource is 'not essential' to the page. A side effect is that this response is not rendered in the browser, unless its served back to the browser (which we do if response is below a certain size, 5M currently) or if its an 'essential' resource (html/js/css):
This fix:
- Passes in the correct id for this to work in the first place, as it was always failing
- Fixes expectedSize to -1 instead of 0, to avoid expecting 0 size response instead of unknown
- If the response is incomplete, allow for refetching the response async
- Update serializeToWARC() to return three states: success, skipped, or aborted. If aborted (eg. incomplete record, actual size mismatch with expected size), the response can be retried if coming from the fetchResponseBody()

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -815,9 +815,12 @@ export class Recorder extends EventEmitter {
 
       streamingConsume = await this.fetchResponseBody(requestId, reqresp, cdp);
 
+      // reqresp object should no longer be used for network events:
+      // either will be consumed (true) or will be tried async (false)
+      this.removeReqResp(networkId);
+
       // if not consumed via takeStream, attempt async loading
       if (!streamingConsume) {
-        this.removeReqResp(networkId);
         const opts: AsyncFetchOptions = {
           reqresp,
           expectedSize: contentLen,
@@ -1509,7 +1512,6 @@ export class Recorder extends EventEmitter {
         logger.warn("Error Serializing to WARC", e, "recorder");
         return false;
       }
-      this.removeReqResp(reqresp.requestId);
       return true;
     } catch (e) {
       logger.debug(


src/util/recorder.ts
```diff
@@ -1482,21 +1501,28 @@ export class Recorder extends EventEmitter {
...
+        ) {
+          return false;
+        }
+      } catch (e) {
+        logger.warn("Error Serializing to WARC", e, "recorder");
         return false;
       }
-
-      this.removeReqResp(requestId);
+      this.removeReqResp(reqresp.requestId);
```
tw4l: Sort of seems like we're using `requestId and reqresp.requestId` interchangeably here (e.g. using `requestId` to start the stream, then `reqresp.requestId` here now) - are there any cases where they should be different from each other? Looks like elsewhere we have a check to treat the requestId as invalid if so.

src/util/recorder.ts
tw4l, line 1918: Just double checking this still holds with the change of defaulting to `-1`

src/util/recorder.ts
ikreymer: You're right, and that was part of the confusion here - there are actually two different requestIds:
https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-RequestId
https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#type-RequestId

The `reqresp` object and most other places we deal with the Network.requestId, while in handleFetchRequest(), we use the Fetch.requestId
-- while the other is called networkId. Passing the wrong id caused the initial bug here :(

Perhaps using more precise typing can help, but yeah, this needs to be the reqresp.requestId to remove the request in our tracking

src/util/recorder.ts
ikreymer: Yes, should be more correct now, as this should only happen if content-length was set to 0

src/util/recorder.ts
ikreymer: Got rid of this altogether, now just calling `this.removeReqResp(networkId)` in calling function, consistent with other uses!

tw4l: Thanks for the clarification!

PR #977, opened 2026-02-18
Support QA with deduped crawls

ikreymer:
deps: update to RWP 2.4.2 to support QA for deduped crawls, update wabac.js 2.25.4
errors: always mark 404 as 0% match: if no replay resources returned but crawled resources exist, set text and screenshot match to 0

### DIFF
--- a/src/replaycrawler.ts
+++ b/src/replaycrawler.ts
@@ -636,7 +636,7 @@ export class ReplayCrawler extends Crawler {
     resourceCounts.replayBad = replayBad;
 
     // If no resources replay, implies a 404 and set other matches to 0
-    if (replayGood === 0 && (crawlGood + crawlBad) > 0) {
+    if (replayGood === 0 && crawlGood + crawlBad > 0) {
       logger.error("Replay Error, no resources replayed");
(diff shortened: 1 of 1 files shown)


src/replaycrawler.ts
```diff
@@ -635,6 +635,13 @@ export class ReplayCrawler extends Crawler {
     resourceCounts.replayGood = replayGood;
     resourceCounts.replayBad = replayBad;
 
+    // If no resources replay, implies a 404 and set other matches to 0
+    if (replayGood === 0 && crawlGood > 0) {
```
tw4l:
```suggestion
    if (replayGood === 0 && crawlGood >= 0) {
```
In case the crawler also encountered a 404 page that was just plain HTML with no resources?

ikreymer: Good point! maybe it should be this then? An archived 404 page would be counted as crawlBad:

```suggestion
    if (replayGood === 0 && (crawlGood + crawlBad) > 0) {
```

tw4l: Ah true!, Yeah I think that makes sense :)

PR #978, opened 2026-02-19
Add Indexer options to commit/cancel single crawl

ikreymer:
Add a --commitCrawlId and --cancelCrawlId to indexer which will either:
- merge the the data for a single crawl into the main index. This can take a variable amount of time, depending on size of index.
- improved logging for commit, and ensure counts are never incremented twice, even if long-running commit job is restarted.
- delete data for canceled crawl from uncommitted crawls list (should be very fast).

Both of these tasks also done as part of the crawl itself, and is still useful for single-crawler operation. However, with Browsertrix controlled in k8s, these tasks are better handled as separate jobs, especially since crawl may be committed without restarting the crawler container (eg. for paused crawls).

### DIFF
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -76,6 +76,7 @@ export class CrawlIndexer {
 
   async run() {
     logger.setDebugLogging(true);
+    logger.setDefaultLogContext("dedupe");
 
     process.on("SIGINT", () => this.handleInterrupt("SIGINT"));
 
@@ -84,7 +85,6 @@ export class CrawlIndexer {
     logger.info(await getInfoString());
 
     const params = this.initArgs();
-    logger.info("Running with Args", params);
 
     const redis = await initRedisWaitForSuccess(params.redisDedupeUrl);
     const dedupeIndex = new RedisDedupeIndex(redis, "");
--- a/src/util/logger.ts
+++ b/src/util/logger.ts
@@ -80,8 +80,13 @@ class Logger {
   logLevels: string[] = [];
   contexts: LogContext[] = [];
   excludeContexts: LogContext[] = [];
+  defaultLogContext: LogContext = "general";
   crawlState?: RedisCrawlState | null = null;
   fatalExitCode: ExitCodes = ExitCodes.Fatal;
+
+  setDefaultLogContext(value: LogContext) {
+    this.defaultLogContext = value;
+  }
 
   setDefaultFatalExitCode(exitCode: number) {
     this.fatalExitCode = exitCode;
@@ -186,19 +191,35 @@ class Logger {
     }
   }
 
-  info(message: string, data: unknown = {}, context: LogContext = "general") {
+  info(
+    message: string,
+    data: unknown = {},
+    context: LogContext = this.defaultLogContext,
+  ) {
     this.logAsJSON(message, data, context, "info");
   }
 
-  error(message: string, data: unknown = {}, context: LogContext = "general") {
+  error(
+    message: string,
+    data: unknown = {},
+    context: LogContext = this.defaultLogContext,
+  ) {
     this.logAsJSON(message, data, context, "error");
   }
 
-  warn(message: string, data: unknown = {}, context: LogContext = "general") {
+  warn(
+    message: string,
+    data: unknown = {},
+    context: LogContext = this.defaultLogContext,
+  ) {
     this.logAsJSON(message, data, context, "warn");
   }
 
-  debug(message: string, data: unknown = {}, context: LogContext = "general") {
+  debug(
+    message: string,
+    data: unknown = {},
+    context: LogContext = this.defaultLogContext,
+  ) {
     if (this.debugLogging) {
       this.logAsJSON(message, data, context, "debug");
     }
@@ -207,7 +228,7 @@ class Logger {
   fatal(
     message: string,
     data = {},
-    context: LogContext = "general",
+    context: LogContext = this.defaultLogContext,
     exitCode = ExitCodes.Success,
   ) {
     exitCode = exitCode || this.fatalExitCode;


docs/docs/develop/dedupe.md
```diff
@@ -49,6 +49,8 @@ to the merged index. The following key is used to track uncomitted crawls:
 
 If a crawl is canceled, these 3 keys should be removed and the crawl id should be removed from `uncommittedcrawls`
 
+The indexer entry point a dedicated `indexer --cancelCrawlId <crawlid>` to perform the cancelation operation.
```
tw4l:
```suggestion
The indexer entrypoint includes a dedicated `indexer --cancelCrawlId <crawlid>` option to perform the cancellation operation.
```

docs/docs/develop/dedupe.md
```diff
@@ -126,6 +131,13 @@ the existing crawls, by rerunning the crawl 'commit' process for each existing c
 
 The result is that all data related to removed crawls is purged, and the `removeCrawlSize` and `removeCrawls` counts are reset to 0.
 
+### Committing and Canceling
+
+The indexer entrypoint also includes options to commit or cancel a crawl, as explained above. These options,
+`--commitCrawlId <crawlid>` and `--cancelCrawlId <crawlid>` are useful when the crawler is controlled via k8s, eg. with --restartsOnError is set.
```
tw4l:
```suggestion
`--commitCrawlId <crawlid>` and `--cancelCrawlId <crawlid>` are useful when the crawler is controlled via k8s, e.g. with `--restartsOnError` set.
```

docs/docs/develop/dedupe.md
```diff
@@ -126,6 +131,13 @@ the existing crawls, by rerunning the crawl 'commit' process for each existing c
...
+
+When this flag is set, it is expected that the commit/cancel operations will be run separately, otherwise, they are attempted as part of the crawl (after WACZ upload or upon cancellation).
```
tw4l:
```suggestion
When the `--restartsOnError` flag is set, it is expected that the commit/cancel operations will be run separately, and so they are not run as part of the crawl. Otherwise, they are attempted as part of the crawl (after WACZ upload or upon cancellation).
```

Just to be as explicit as possible.

src/util/state.ts
```diff
@@ -307,8 +316,28 @@ export class RedisDedupeIndex {
       }
     }
 
-    // add counts
-    await this.addCrawlCounts(crawlId);
+    logger.info("Crawl committed to merged index!", {
```
tw4l:
```suggestion
    logger.info("Crawl committed to merged index", {
```

The exclamation point is fun but unnecessary ;)

src/util/state.ts
```diff
@@ -273,19 +273,28 @@ export class RedisDedupeIndex {
...
+  async commitDedupeDone(crawlId?: string, uncommitted_key = DUPE_UNCOMMITTED) {
     crawlId = crawlId || this.crawlId;
 
-    // set this first, as crawl is considered 'committed' as soon
-    // as data is partially merged in
-    await this.dedupeRedis.srem(DUPE_UNCOMMITTED, crawlId);
-    await this.dedupeRedis.sadd(DUPE_ALL_CRAWLS, crawlId);
+    // check if already committed
+    if (!(await this.dedupeRedis.sismember(uncommitted_key, crawlId))) {
+      logger.warn("Crawl not found for committing, or already committed", {
```
tw4l: Might be worth adding a new `dedupe` log context for this and the other logger messages here? When called by the indexer it's less important, but is also called by the crawler and the context would be very helpful in the crawl logs to help understand what these messages are referring to/support filtering like we do for other log messages.

src/util/state.ts
```diff
@@ -307,8 +316,28 @@ export class RedisDedupeIndex {
...
+      newHashes,
+      numWacz,
+    });
+
+    // move to all crawls here
+    if (
+      !(await this.dedupeRedis.smove(uncommitted_key, DUPE_ALL_CRAWLS, crawlId))
+    ) {
+      // if already moved, return here to avoid duplicating counts
+      logger.info("Crawl counts not updated, already added to allcounts");
```
tw4l: Maybe better as warn or debug?

src/util/state.ts
```diff
@@ -307,8 +316,28 @@ export class RedisDedupeIndex {
...
+      return;
+    }
+
+    const counts = await this.dedupeRedis.hgetall(`h:${crawlId}:counts`);
+    for (const [key, value] of Object.entries(counts)) {
+      await this.dedupeRedis.hincrby(DUPE_ALL_COUNTS, key, Number(value));
+    }
+
+    logger.info("Crawl counts added to allcounts");
```
tw4l: Similarly, seems more of a debug message, and could use dedupe log context. Otherwise might be hard to understand what this log line is saying for someone not well-versed in our dedupe architecture.

src/indexer.ts
```diff
@@ -73,10 +84,32 @@ export class CrawlIndexer {
     logger.info(await getInfoString());
 
     const params = this.initArgs();
+    logger.info("Running with Args", params);
```
tw4l:
```suggestion
```

Probably worth removing before merging

ikreymer: can make debug

ikreymer: eh, I guess don't really need it at this point.

src/util/state.ts
ikreymer: Done!

src/util/state.ts
ikreymer: good idea, switched to warn.

src/util/state.ts
ikreymer: switched to debug.

PR #979, opened 2026-02-19
fix regression where behaviors are run prematurely in new pages due to remaining 'framenavigated' listener

ikreymer:
- regression caused by detecting new iframes while behaviors are running, added in #964
- stop running behaviors in new iframes after behavior run has finished
- use page.off() and also page.removeAllListeners() even if runBehaviors() times out
- caused behaviors to scroll too early, also affecting QA tests

Testing:
Run: `docker run -p 9037:9037 -it webrecorder/browsertrix-crawler crawl --url https://old.webrecorder.net/ --screencastPort 9037  --postLoadDelay 10 --screenshot view --limit 5`
Observe in screencast that second page has already scrolled down after while waiting for the 10 seconds after load, before behaviors have run. With this fix, that does not happen.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1275,9 +1275,6 @@ self.__bx_behaviors.selectMainBehavior();
           true,
         );
 
-        // in case runBehaviors() times out, remove all listeners here
-        page.removeAllListeners("framenavigated");
-
         data.contentCheckAllowed = false;
 
         await this.netIdle(page, logDetails);
--- a/src/util/worker.ts
+++ b/src/util/worker.ts
@@ -138,6 +138,8 @@ export class PageWorker {
         { reuseCount: this.reuseCount, ...this.logDetails },
         "worker",
       );
+      // remove any frame listeners added in behaviors and page timed out and didn't remove them before
+      this.page!.removeAllListeners("framenavigated");
       this.opts!.data = pagestate;
       return this.opts!;
     } else if (this.page) {


src/crawler.ts
```diff
@@ -1275,6 +1275,9 @@ self.__bx_behaviors.selectMainBehavior();
           true,
         );
 
+        // in case runBehaviors() times out, remove all listeners here
+        page.removeAllListeners("framenavigated");
```
tw4l: A bit edge case-y becuase there should always be ~10 seconds between the behaviors timing out and the page as a whole timing out, but if we wanted to be really truly certain to remove these listeners even if the entire page crawl (and not just behaviors) timed out before reaching this point, we could potentially do this near the top of `PageWorker.initPage()`.

ikreymer: yeah, we've seen all sorts of edge cases, so actually not a bad idea. Will move that there in the `reuse` condition.

src/util/worker.ts
```diff
@@ -138,6 +138,8 @@ export class PageWorker {
         { reuseCount: this.reuseCount, ...this.logDetails },
         "worker",
       );
+      // remove any frame listeners added in behaviors and page timed out and didn't remove them before
```
tw4l, line 141:
```suggestion
      // remove any frame listeners added in behaviors in case page timed out before removing
```

PR #981, opened 2026-02-20
rollover test: expect at least 5 WACZs, not 6, 

ikreymer:
with rollover fix in #974, which avoids an empty WACZ from being created, there may only be 5 WACZs with the test


ikreymer:
Hm, maybe this is no longer needed, was happening at one point, will close for now.

PR #986, opened 2026-02-28
Request original url, store normalized url for comparison check/dedupe

### LINKED ISSUE #982: Seed URL normalization in v1.11.2 changes what URLs are fetched and recorded
I probably should have caught this in the release notes (so, my fault), but I discovered today that a bunch of crawled pages are not getting properly recorded in my infrastructure because the normalization changes in #959 appear to alter the URL that is actually requested when crawling (or at least the URL that is recorded in WARCs). Within my infra, that’s a change I can handle, but I am a little concerned about the change because a URL’s query component is not actually unordered key/value pairs. Most servers treat it that way, but not all, and changing the order may break some URLs and cause the server to return something different (I have a *lot* of URLs to go back and check to make sure they are still OK after realizing I missed this change in the release notes).

I understand the use of sorting for doing various matching tasks around scoping and the value of loose(ish) comparisons there, but I think it’s somewhat important that the requested URL and the seed are actually equivalent (so you could do things like lower-case the scheme and hostname or standardize whether 80/443 port numbers are included, but not re-ordering query parameters). Is there any change the URL can be treated different for scoping/deduping/etc. purposes than for what will actually be requested and recorded?

ikreymer:
Fixes #982 

This should ensure the original URL is requested, while the normalized URL is stored in the seed, used for scope check, added to seen list, and used for dedupe.

Testing: Adds to ensure a URL is requested as is, while avoiding duplicates when normalized.
Attempt to crawl`https://example-com.webrecorder.net/?B=1&A=2` and `https://example-com.webrecorder.net/?A=2&B=1` which normalize to the same URL, `https://example-com.webrecorder.net/?A=2&B=1` after sorting query args. Before this PR, the URL crawled would be the normalized one, `https://example-com.webrecorder.net/?A=2&B=1` but with this PR `https://example-com.webrecorder.net/?B=1&A=2` will be crawled as its the first URL requested.

### DIFF
--- a/tests/sitemap-parse.test.js
+++ b/tests/sitemap-parse.test.js
@@ -27,13 +27,17 @@ async function waitContainer(containerId) {
   }
 }
 
+let startPort = 36701;
+
 async function runCrawl(numExpected, url, sitemap="", limit=0, numExpectedLessThan=0, extra="") {
-  const command = `docker run -d -p 36381:6379 -e CRAWL_ID=test webrecorder/browsertrix-crawler crawl --url ${url} --sitemap ${sitemap} --limit ${limit} --context sitemap --logging debug --debugAccessRedis ${extra}`;
+  const port = startPort++;
+
+  const command = `docker run --rm -d -p ${port}:6379 -e CRAWL_ID=test webrecorder/browsertrix-crawler crawl --url ${url} --sitemap ${sitemap} --limit ${limit} --context sitemap --logging debug --debugAccessRedis ${extra}`;
   const containerId = child_process.execSync(command, {encoding: "utf-8"});
 
   await sleep(3000);
 
-  const redis = new Redis("redis://127.0.0.1:36381/0", { lazyConnect: true, retryStrategy: () => null });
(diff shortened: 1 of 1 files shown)


ikreymer:
@Mr0grog do you think this works for your use case?

src/util/recorder.ts
```diff
@@ -2041,7 +2044,6 @@ class AsyncFetcher {
       result.resource;
 
     if (!success || !stream) {
-      //await this.recorder.crawlState.removeDupe(ASYNC_FETCH_DUPE_KEY, url);
```
tw4l, line 2044: Should this not have been here to begin with?

ikreymer: I think this is leftover from different approach to due checking...

tw4l: Maybe better to just remove rather than commenting out? Or leave a descriptive comment if you want to come back to it. It seemed out of place to me as well, and doesn't match the function signature of `crawlState.removeDupe`.

tw4l (approved):

src/util/recorder.ts
tw4l: Nevermind, I was reading the diff incorrectly!

PR #992, opened 2026-03-11
Implement text extraction from raw HTML + client-side rendering (CSR) detection in QA

emma-sg:
Adds option (`--qaDetectClientSideRendering`) to analyze replayed html content to detect CSR clues such as framework indicators, hydration data, web components, and DOM manipulation patterns to identify client-side rendered pages. Output is written to the `info` warcs along with the rest of the page info (pages, comparison info, etc).

Will follow up with extracted text analysis using & not using additional page resources.


src/replaycrawler.ts
```diff
@@ -591,6 +603,112 @@ export class ReplayCrawler extends Crawler {
...
+  }
+
+  async extractRawTextFromArchive(
+    page: Page,
+    url: string,
+    date: Date,
+  ): Promise<string | undefined> {
+    const timestamp = date.toISOString().slice(0, 19).replace(/[T:-]/g, "");
+    // `if_` suffix ensures wabac.js serves the unaltered source
+    const replayUrl = REPLAY_PREFIX + `${timestamp}if_/${url}`;
```
ikreymer:
```suggestion
    // `id_` suffix ensures wabac.js serves the unaltered source
    const replayUrl = REPLAY_PREFIX + `${timestamp}id_/${url}`;
```

This should be id_, if_ is for iframe rewriting

PR #994, opened 2026-03-11
Make fatal() and interrupt() async

### LINKED ISSUE #993: Possible race condition due to fatal() and interrupt() changing state async
These functions actually create a promise to change state, then exit.
It is possible for state to be changed in between, resulting in unexpected behavior.
These functions should be made async and awaited whenever possible.

ikreymer:
- Fix #993 
- fatal() and interrupt() give false impression that fatal or interrupt happens right away, instead of in a promise after state is updated. Make them async to clear this up.
- most uses of these functions are already in async functions, add 'void' elsewhere to keep current behavior (mostly in arg validation)
- fix race condition where state could be changed after a call to fatal() or interrupt() as they call async function
- Also add CrawlStatus type for clarity of available types
- bump to 1.12.2

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -48,6 +48,7 @@ import {
   InterruptReason,
   BxFunctionBindings,
   MAX_JS_DIALOG_PER_PAGE,
+  CrawlStatus,
 } from "./util/constants.js";
 
 import { AdBlockRules, BlockRuleDecl, BlockRules } from "./util/blockrules.js";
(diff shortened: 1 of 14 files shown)


src/crawler.ts
```diff
@@ -381,9 +382,7 @@ export class Crawler {
       logger.setLogBehaviorsToRedis(true);
     }
 
-    if (this.params.logErrorsToRedis || this.params.logBehaviorsToRedis) {
-      logger.setCrawlState(this.crawlState);
-    }
+    logger.setCrawlState(this.crawlState);
```
tw4l, line 385: Nice catch here, since it's necessary for `setStatusAndExit`

Issue #1004, opened 2026-03-24
Deduplication across multiple crawls in the same wacz file seems to be broken


Merlin04:
I'm using browsertrix-crawler to regularly download the contents of a website into a wacz file. I'd like to be able to have past versions of the site contained in the same archive file, and to use deduplication so that I don't waste space duplicating the page over and over again when nothing changes.

This is the config I'm using:

```yaml
seeds:
  - "https://example.com"
combineWARC: false
generateWACZ: true
collection: public
profile: /crawls/profiles/profile.tar.gz
redisDedupeUrl: redis://redis:6379/1
```

The first crawl works fine, but the second crawl produces a warc file where all the pages/resources appear in replayweb.page but clicking on any of them gives the error "Sorry, this page was not found in this archive". I've attached the WACZ file resulting from two crawls of example.com (renamed to a zip so github will allow it) here: [archive.wacz.zip](https://github.com/user-attachments/files/26228261/archive.wacz.zip)

ikreymer:
Thanks for reporting! This is actually a replay issue, to be fixed in wabac.js, as it looks like you created a WACZ file that was called public.wacz, which the next crawl assumes may exist, but doesn't.

eg: in datapackage.json there is:

```
  "relation": {
    "requires": [
      {
        "filename": "public.wacz",
        "hash": "9539dda087a484196f7470ec28d8e6aa59ef20ce0397e52d45f613c5d7f269fd",
        "size": 10548,
        "crawlId": "6f0d1624e4b6-public"
      }
    ]
  }
```

The replay should ignore this non-existent file, which will be fixed in wabac.js.

For your use case, since you're creating one WACZ always, perhaps we should have a way to turn off this dependency tracking, since I assume you're trying to avoid linking to an external WACZ, and always have just one WACZ?

PR #1005, opened 2026-03-26
track crawlIds included in each --collection directory

ikreymer:
- track crawlIds that are included in each crawl via crawls/ids.txt list, one crawlId per line
- when generating wacz, don't reference crawls that are part of current wacz in relation.requires list
- tests: add test to ensure included crawls not references as external
- fixes issue in #1004

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -153,7 +153,7 @@ export class Crawler {
   warcCdxDir: string;
   indexesDir: string;
 
-  crawlsDir: string;
+  crawlIdsDir: string;
 
   downloadsDir: string;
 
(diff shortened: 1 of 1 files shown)


src/crawler.ts
```diff
@@ -374,6 +379,10 @@ export class Crawler {
       dedupeRedis,
     );
 
+    if (!this.params.dryRun) {
+      await this.crawlState.loadAndAddIncludedCrawlIds(this.crawlsDir);
```
tw4l: The crawler is attempting to create this file before the collection crawls directory has been created, which is resulting in a "no such file or directory" error being thrown.

tw4l: Now fixed

ikreymer: Thanks! was a last minute refactor that should've tested..

PR #1015, opened 2026-04-07
Detection and Skip Rate Limited Pages

### LINKED ISSUE #758: Retry Improvements + Rate Limit Support
Following up to #132 (and also #392, #360) , we need a more sophisticated retry strategy, also considering what do with rate limiting status code.
We already have --failOnInvalidStatus, --maxPageRetries, --failOnFailedSeed and --failOnFailedLimit and probably need to add a few more flags.

This is getting slightly messy, but hopefully there's a clear path to figure this out.

There's a few options to consider:
- Which status code should be counted as page failures, for purposes of ending crawl
- Which status codes should result in retrying the page
- Should capture of pages with invalid status codes be skipped when they will be retried.
- Which status code should result in slowing down the crawl / adding a delay before loading those pages again if retrying..

It's probably useful to list the various use cases:
- The crawler should treat 4xx and 5xx as failed, possibly customizing which status codes are included?
- The crawler should fail the crawl if a certain number of pages have failed or if any of the seeds have failed.
- The crawler should retry failed pages a certain number of times, possibly customizing which status codes are eligible for retries.
- The crawler should not write any data for pages that are being retried, until the final retry.

With this in mind, probably should add at least a:
- --retryStatusCodes flag which indicates which status codes will be retried.
- Is there a need to also specify --invalidStatusCodes that is separate from --retryStatusCodes? Leaning against it.
- Is there a need to also specify if failed pages that are being retried should be captured to WARC? Sort of leaning against it as well, since retries are part of the capture process
- How to handle rate limiting, eg. add exponential backoff via pageExtraDelay for certain status codes, like 429, 503 maybe 403.. Possibly using `Retry-After`, if available (from #392)

ikreymer:
Always Rate Limit Status Codes:
  - detect 403, 429, 503 status codes as possible rate limited, skip pages that have these status codes from being archived
  - add --rateLimitStatusCodes to customize these status codes
  - can use --no-rateLimitStatusCodes to disable rate limiting altogether
    
Conditional Rate Limiting by text match:
  - add --rateLimitOn200MatchText to allow configuring text that is considered rate limit page even with 200 status
  - default to incapsula string

Rate Limit Retries:
  - Track and retry rate limited pages at the end of the queue, similar to failed pages
  - By default, retry indefinitely or up to --rateLimitMaxRetries retries if set to >-1
    
Rate Limit Timeout:
  - add --rateLimitTimeout for how long rate limit should be tracked when at least one rate limited pages is detected
  - If Retry-After header is provided, use that value for retry seconds instead, otherwise use setting from rateLimtTimeout (default to 5 min)
  
Optional Crawler Interrupt:
  - add --rateLimitInterruptCount to indicate number of rate-limited pages at which the crawler should be interrupted with new exit code (18)
  - defaults to -1 to never interrupt
  - set rateLimited key for easier detection in browsertrix app when a crawler is rate limited and will exit on next start
  
Direct Fetch Cleanup:
  - separate rate limit for direct fetch, skip direct fetch if it is rate limited, but don't interrupt crawl
  - direct fetch: return status code of direct fetch attempt to track rate limits on direct fetch by status
  - use special constants for STATUS_UNKNOWN_ERROR and STATUS_IS_HTML_NO_DIRECT_FETCH to indicate that direct fetch is to be skipped and regular browser load performed
  
Fixes #758


src/util/argParser.ts
```diff
@@ -738,11 +741,19 @@ class ArgParser {
...
+            "Consider page rate limited given the following matches by status code and text",
+          type: "array",
+          default: RATE_LIMIT_MATCH_200,
+=======
         reportSkipped: {
           describe:
             "If set, write information about URLs encountered but not queued to reports/skippedPages.jsonl",
           type: "boolean",
           default: false,
+>>>>>>> main
```
```suggestion
        rateLimitOn200MatchText: {
          describe:
            "Consider page rate limited given the following matches by status code and text",
          type: "array",
          default: RATE_LIMIT_MATCH_200,
        },

        reportSkipped: {
          describe:
            "If set, write information about URLs encountered but not queued to reports/skippedPages.jsonl",
          type: "boolean",
          default: false,
```

src/util/recorder.ts
```diff
@@ -1454,20 +1494,23 @@ export class Recorder extends EventEmitter {
...
     const mime = reqresp.getMimeType() || "";
     // cancel if not 200 or mime is html
-    if (reqresp.status !== 200 || isHTMLMime(mime)) {
+    if (reqresp.status !== 200) {
       await fetcher.doCancel();
-      return false;
+      return reqresp.status;
+    }
+    if (isHTMLMime(mime)) {
+      return 600;
```
tw4l: This could use a comment to explain why we're giving it a 600 error code I think

src/util/state.ts
```diff
@@ -47,6 +47,11 @@ export function normalizeDedupeStatus(status: number): string {
   return status + "";
 }
 
+// ============================================================================
+// Rate Limit Constants
+const RATE_LIMIT_TIME = 300;
+const RATE_LIMIT_MAX = 3;
```
tw4l: I think it would be good to make these values configurable via the arg parser to give people more control over how sensitive they want the crawler to be re: rate limits, especially if adding custom 200 status code matches

tw4l (commented):
Just leaving some comments as I look more closely and start to test.

Could use a new docs section to explain the motivation and mechanisms behind this change as well as how to configure it. Good to have in general but especially as this is likely to result in peoples' crawls behaving differently without any changes to their configs, which might catch some people off guard.

(Bringing from separate offline conversation: Also worth considering if this should be disabled by default and opt-in)

src/util/state.ts
ikreymer: Made these configurable, --rateLimitInterruptCount which defaults to -1, can be set to how many concurrent rate limited pages will result in an interrupt. The --rateLimitTimeout is set to 300.

src/util/recorder.ts
ikreymer: Replaced this with a constant STATUS_IS_HTML_NO_DIRECT_FETCH to indicate a custom error code.

src/util/state.ts
tw4l: Thank you!

src/util/constants.ts
```diff
@@ -124,4 +126,18 @@ export enum SkippedReason {
   RobotsTxt = "robotsTxt",
   RedirectToExcluded = "redirectToExcluded",
   Duplicate = "duplicate",
+  RateLimit = "RateLimited",
```
tw4l:
```suggestion
  RateLimit = "rateLimited",
```

To keep consistent casing with other reasons

src/util/constants.ts
```diff
@@ -124,4 +126,18 @@ export enum SkippedReason {
...
 }
+
+// Direct Fetch Error Constants
+export const STATUS_IS_HTML_NO_DIRECT_FETCH = 600;
+
+export const STATUS_UNKNOWN_ERROR = 999;
+
+// ============================================================================
+// Rate Limit Constants
+export const RATE_LIMIT_TTL = 300;
```
tw4l:
```suggestion
export const RATE_LIMIT_TTL_SECS = 300;
```

src/util/argParser.ts
```diff
@@ -744,6 +749,33 @@ class ArgParser {
...
+
+        rateLimitStatusCodes: {
+          describe:
+            "Consider responses with these status codes to be treated as rate-limited/blocked responses",
+          type: "array",
+          default: [403, 429, 503],
+        },
+
+        rateLimitTimeout: {
+          describe: "Time to track rate limited count for before resetting",
```
```suggestion
          describe: "Time in seconds to track rate limited count for before resetting",
```

We don't specify what unit of time this is anywhere yet - I assume seconds, since we're treating it as Retry-After?

src/util/argParser.ts
```diff
@@ -19,6 +19,8 @@ import {
   DEFAULT_MAX_RETRIES,
   BxFunctionBindings,
   DEFAULT_CRAWL_ID_TEMPLATE,
+  RATE_LIMIT_MATCH_200,
+  RATE_LIMIT_TTL,
```
```suggestion
  RATE_LIMIT_TTL_SECS,
```

src/util/state.ts
```diff
@@ -13,6 +13,7 @@ import {
   DUPE_UNCOMMITTED,
   CrawlStatus,
   SkippedReason,
+  RATE_LIMIT_TTL,
```
tw4l:
```suggestion
  RATE_LIMIT_TTL_SECS,
```

src/util/state.ts
```diff
@@ -840,6 +848,9 @@ export class RedisCrawlState extends RedisDedupeIndex {
 
     this.sitemapDoneKey = this.crawlId + ":sitemapDone";
 
+    this.rateLimitTTL = rateLimitTTL || RATE_LIMIT_TTL;
```
tw4l:
```suggestion
    this.rateLimitTTL = rateLimitTTL || RATE_LIMIT_TTL_SECS;
```

src/util/recorder.ts
```diff
@@ -1471,20 +1503,25 @@ export class Recorder extends EventEmitter {
...
-    if (reqresp.status !== 200 || isHTMLMime(mime)) {
+    if (reqresp.status !== 200) {
       await fetcher.doCancel();
-      return false;
+      return reqresp.status;
+    }
+    if (isHTMLMime(mime)) {
+      // custom value to indicate direct fetch is skipped
+      // since this is an HTML page
+      return STATUS_IS_HTML_NO_DIRECT_FETCH;
```
tw4l, line 1519: Do we still want to cancel the fetcher in this case as we do for non-200?

src/util/state.ts
```diff
@@ -1025,16 +1038,21 @@ return inx;
...
 
-  async markFailed(url: string, noRetries = false) {
+  async markFailed(url: string, noRetries = false, alwaysRetry = false) {
     return await this.redis.requeuefailed(
       this.pkey,
       this.qkey,
       this.fkey,
       url,
-      noRetries ? 0 : this.maxRetries,
+      noRetries ? 0 : alwaysRetry ? -1 : this.maxRetries,
```
tw4l:
```suggestion
      noRetries ? 0 : (alwaysRetry ? -1 : this.maxRetries),
```

Just to help a bit with readability

tw4l (commented):
A few comments/suggestions for now.

Going to continue testing but so far it also seems that it's possible `rateLimitInterruptCount` isn't getting enforced as expected - I'm seeing hundreds of log messages indicating a domain is likely rate-limiting without the crawl being interrupted even with `--rateLimitInterruptCount 3` set.

tw4l (approved):
With latest push, it's now working as expected in testing.

Outstanding comments still need attention (feel free to ignore the TTL constant renaming ones if you prefer) but approving so you're unblocked to address those and then merge.

We definitely need to add documentation and crawler tests before including this work in a release.

src/util/recorder.ts
ikreymer: yes, as we only do direct fetch if its not HTML.

Issue #1023, opened 2026-04-21
Issue with hash URL settings --allowHashUrls / allowHash: true


philippwittwer:
Using Browsertrix-Crawler 1.12.4 on a Windows-VM in Docker 
appears to ignore hash URL settings  despite using  --allowHashUrls or a config.yaml with allowHash: true

Example-Website:

example.com using:

docker run --rm -it -p 9037:9037 -v "C:\browsertrix-webarchive:/Crawls" webrecorder/browsertrix-crawler crawl --url "https://example.com"    --screenshot view --text to-warc --generateWACZ  --allowHashUrls --screencastPort 9037

Observed in startup log:
"allowHash": false

Log:
"2026-04-21T12:25:00.027Z","logLevel":"info","context":"general","message":"Browsertrix-Crawler 1.12.4 (with warcio.js 2.4.9)","details":{}}
{"timestamp":"2026-04-21T12:25:00.030Z","logLevel":"info","context":"general","message":"Seeds","details":[{"url":"https://example.com/","normUrl":"https://example.com/","scopeType":"prefix","include":["/^https?:\\/\\/example\\.com\\//"],"exclude":[],"allowHash":false,"depth":-1,"sitemap":null,"auth":null,"_authEncoded":null,"maxExtraHops":0,"maxDepth":1000000}]}
{"timestamp":"2026-04-21T12:25:00.031Z","logLevel":"info","context":"general","message":"Link Selectors","details":[{"selector":"a[href]","extract":"href","isAttribute":false}]}
{"timestamp":"2026-04-21T12:25:00.031Z","logLevel":"info","context":"general","message":"Behavior Options","details":{"message":"{\"autoplay\":true,\"autofetch\":true,\"autoscroll\":true,\"siteSpecific\":true,\"log\":\"__bx_log\",\"startEarly\":true,\"clickSelector\":\"a\"}"}}
{"timestamp":"2026-04-21T12:25:06.929Z","logLevel":"info","context":"worker","message":"Creating 1 workers","details":{}}
{"timestamp":"2026-04-21T12:25:06.930Z","logLevel":"info","context":"worker","message":"Worker starting","details":{"workerid":0}}
{"timestamp":"2026-04-21T12:25:07.256Z","logLevel":"info","context":"worker","message":"Starting page","details":{"workerid":0,"page":"https://example.com/"}}
[...]

PR #1025, opened 2026-04-22
Fix allowHashUrls option and scope checking for hash URLs

ikreymer:
allow #-tag urls to be treated as distinct URLs for scope types other than custom:
- fix --allowHashUrls option being ignored altogether
- if --allowHashUrls global or --allowHash per seed is set, don't reset to false if scopeType != 'custom'
- tests: update scope tests to ensure --allowHash and --allowHashUrls now work as expected
- improve typing for ScopedSeed, add ScopedSeedInitOpts

fixes part of #1023


tw4l (approved):
@ikreymer You might want to double-check tests after rebasing on main as there was some overlap between this PR and the relative URL one, but appears to be working well and tests are passing.

PR #1026, opened 2026-04-24
support scope check for relative URLs using current page:

ikreymer:
Previously, the assumption was that only absolute URLs are provided from the link extraction.
However, with custom link selectors and user provided behaviors, relative URLs are possible.
This change ensures that relative URLs can be resolved to the current page:
- pass pageUrl to seed.isIncluded() to allow resolving scope of relative URLs to current page, if provided
- resolves any relative URLs found added from custom link extraction
- fixes part of #1023

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -765,12 +765,14 @@ export class Crawler {
       depth,
       extraHops,
       noOOS,
+      pageUrl,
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Nice, thanks for the tests!

PR #1028, opened 2026-05-07
Add flag to disable content deduplication

aponb:
This adds a new boolean crawl option, dedupe, which defaults to true and can be disabled on the CLI with --no-dedupe.

When deduplication is disabled, the crawler no longer writes WARC revisit records and ignores dedupe-specific behavior such as cross-crawl dedupe via --redisDedupeUrl and page-level dedupe via
  --dedupePagesMinDepth.

The change is intentionally scoped to dedupe behavior only:

- content hash lookup for revisits is skipped
- dedupe hash state is not written
- page dedupe logic is disabled
- external dedupe store usage is bypassed when dedupe is off

Testing:

- added a regression case to tests/dedupe-basic.test.ts that runs a crawl with --no-dedupe and verifies that no revisit records are written, even when --redisDedupeUrl is set

Notes:

- CLI usage is --no-dedupe
- config/YAML usage is dedupe: false

### DIFF
--- a/docs/docs/user-guide/dedupe.md
+++ b/docs/docs/user-guide/dedupe.md
@@ -3,6 +3,11 @@
 With version 1.12, the crawler includes full support for deduplication ("dedupe") of crawled content by content hash,
 avoiding saving the same content multuple times. When a duplicate is encountered, the crawler saves a reference to the original content, instead of the full response. The crawler supports deduplication in several ways.
 
+To disable content-based deduplication entirely, pass `--no-dedupe`. When
+disabled, the crawler will not write `revisit` records and will ignore
+dedupe-specific options such as `--redisDedupeUrl` and
+`--dedupePagesMinDepth`.
+
 ## Automatic deduplication within a single crawl
 
 By default, the crawler applies the following deduplication automatically:
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -266,7 +266,9 @@ export class Crawler {
     // and additional pages needed from the queue
     // otherwise, queuePageLimit == pagesLimit
     this.queuePageLimit =
-      this.params.dedupePagesMinDepth >= 0 ? 0 : this.pageLimit;
+      this.params.dedupe && this.params.dedupePagesMinDepth >= 0
+        ? 0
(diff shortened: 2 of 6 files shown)


ikreymer:
Before we add this, is it possible that there are other dedupe rules that could be useful instead of turning dedupe off completely?

For example, --dedupeMinSize could be added to only dedupe records > min size.

The page dedupe and external dedupe are already off by default, if those flags aren't provided, they are not used.

Or, do you absolutely need to have no revisit records at all written?

The crawler may be writing revisit records in some other circumstance, such as a resource that is previously captured and then again listed as a page (added in #1074).

aponb:
Yes, for our use case we really need to disable revisit records completely.

Our current workflow crawls without deduplication and does the deduplication afterwards as a separate post-processing step. We'd like to keep that workflow.

The reason is mostly historical. Our post-processing is based on a legacy NetarchiveSuite mechanism from the ARC days, before WARC revisit records existed. It is still in production, so for compatibility we need full records and do the deduplication afterwards.

I think more fine-grained deduplication options are definitely useful for different workflows. I just don't see them as a replacement for being able to switch deduplication off completely. I think disabling it is also a valid workflow.

ikreymer:
@aponb apologies for delay in responding. Got it, I understand that this makes sense for your use case and I agree the workflow is valid. Just thinking about the best way to support this, as I think there is another valid option here: enabling deduplication, but without revisit records, this could be useful if running two crawls concurrently when the data is to be merged, but revisit records may not be wanted.

Wanted to consider if that should be added as part of this option, or maybe a separate options. The options could be:
- --dedupe disabled - no deduplication, no revisit records
- --dedupe with-revisits - deduplication, revisit records
- --dedupe no-revisits - deduplication, no revisit records (skip writing duplicates)

Alternative, this could be a separate option, but we already have so many options so want to consider this approach, in case we need to add more variations in the future as well..

aponb:
yes I think flags like --dedupe disabled or --dedupe with-revisits are a good starting point for different needs in future. Why not to have --dedupe no-revisits , when you just want to save storage without need to have any meta information about deduplication!

PR #1031, opened 2026-05-26
External command tweaks

### LINKED ISSUE #1032: Unify subcommand execution
We've been using a mix of different subprocess methods from node's `child_process` module; we should standardize on a single family instead of different methods with different semantics except when we absolutely have to.

mistydemeo:


src/util/browser.ts
```diff
@@ -273,19 +273,20 @@ export class Browser {
...
-      this.removeSingletons();
+      await this.removeSingletons();
     } catch (e) {
       throw new Error(`Profile ${profileLocalSrc} not a valid tar.gz`);
     }
   }
 
-  removeSingletons() {
+  async removeSingletons() {
+    const singletons = await Array.fromAsync(fsp.glob("./Singleton*"));
```
mistydemeo: Seems like node didn't get `Array.fromAsync` until version 22, so I'll swap this over to `globSync`.

mistydemeo:
For the record, the CI failure is because, it turns out, the whole family of `glob` functions was added in node 22. We should either back that change our until we can upgrade, or use a third party library implementing it.

ikreymer (approved):

PR #1036, opened 2026-06-03
deps: bump to base image 1.90.128

ikreymer:
bump to 1.13.0-beta.2


mistydemeo:
We should probably also update CI to run on the same node version as the Docker image, right?

ikreymer:
> We should probably also update CI to run on the same node version as the Docker image, right?

done! yeah probably good to be consistent

mistydemeo (approved):

PR #1037, opened 2026-06-03
crawler: treat addLink links as in scope unless excluded

### LINKED ISSUE #1033: Allow addLink() to ignore scope include rules, but not exclusions.
To make it easier to allow behaviors to add new links, perhaps the crawler should support ignoring the 'include' rules for addLink, but not 'exclude' rules.

This would allow for behaviors to add new links, unless they were explicitly excluded, even if the links are not otherwise in scope.

mistydemeo:
Previously, links added via behaviours could be rejected as out of scope. This could be confusing to users, who may have expected these URLs to end up being included. For example, when archiving an Instagram profile, we add URLs to archive the stories, but these would be skipped if the profile was added via certain scopes. This adds a new `ignoreScope` parameter which, when present, causes us to ignore the usual out of scope check and only consider explicit exclusions. It also uses a new method to add these URLs to the queue which doesn't perform scope checks and only looks for explicit exclusions.

Fixes #1033.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -67,7 +67,7 @@ import {
 } from "puppeteer-core";
 import { Recorder } from "./util/recorder.js";
 import { SitemapReader } from "./util/sitemapper.js";
-import { ScopedSeed, parseSeeds } from "./util/seeds.js";
+import { LinkEntry, ScopedSeed, parseSeeds } from "./util/seeds.js";
 import { WARCWriter, createWARCInfo, setWARCInfo } from "./util/warcwriter.js";
 import { isHTMLMime, isRedirectStatus } from "./util/reqresp.js";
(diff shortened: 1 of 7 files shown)


src/crawler.ts
```diff
@@ -2592,12 +2600,11 @@ self.__bx_behaviors.selectMainBehavior();
     const { seedId, depth, extraHops = 0, filteredFrames, callbacks } = data;
 
     callbacks.addLink = async (url: string) => {
-      await this.queueInScopeUrls(
+      await this.queueNonExcludedUrls(
```
ikreymer: Hm, instead of adding a new queueNonExcludedUrls, I think should just extend queueInScopeUrls with the extra option.. Probably should convert the params to a dict since there's more options now..

PR #1038, opened 2026-06-08
ci: update docker actions

mistydemeo:
This is similar to #1029, but updates the Docker-managed actions versions. These were also warning us about node changes. I don't *think* any of these changes will break things for us; I read the changelogs.


tw4l (approved):
Thanks!

PR #1039, opened 2026-06-09
ci: update docker buildx actions

mistydemeo:
Somehow totally missed this one in #1038.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -25,7 +25,7 @@ jobs:
           platforms: arm64
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v4
       - name: Login to DockerHub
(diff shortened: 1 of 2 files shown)


mistydemeo:
Tested with https://github.com/webrecorder/browsertrix-crawler/actions/runs/27221036075.

.github/workflows/release.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
           platforms: arm64
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v2
```
tw4l: Should we bump this one to `docker/setup-buildx-action@v4` as well?

mistydemeo: Whoops, yes, good catch. That was a typo.

tw4l (approved):

Issue #726, opened 2024-11-26
Headful interactive profile creation broken after v1.4.0 update


halvir:
The headful interactive profile creation is not functional after updating Browsertrix-Crawler to version 1.4.0 release.
Error displayed on command line is following:

```
node:fs:448
    return binding.readFileUtf8(path, stringToFlags(options.flag));
                   ^

Error: ENOENT: no such file or directory, open '/app/node_modules/@novnc/novnc/core/rfb.js'
    at Object.readFileSync (node:fs:448:20)
    at InteractiveBrowser.handleRequest (file:///app/dist/create-login-profile.js:499:29)
    at Server.<anonymous> (file:///app/dist/create-login-profile.js:329:65)
    at Server.emit (node:events:518:28)
    at parserOnIncoming (node:_http_server:1141:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:118:17) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/app/node_modules/@novnc/novnc/core/rfb.js'
}

Node.js v20.18.1
```

Headless interactive profile creation seems to be working as intended.

ikreymer:
Thanks for the report, novnc should not have been updated, will add a fix to pin it to 1.4.0 shortly.

PR #1042, opened 2026-06-10
fix: resolve dependency CVEs, fix ws v8 screencast regression, prune devDependencies from image

royteeuwen:
## Summary

Dependency security maintenance: `yarn audit` reported 33 unique advisories (2 critical, 19 high); this PR resolves all but one (details below) and removes dev tooling from the published Docker image.

### Direct dependency upgrades
- `undici` ^7.16.0 → ^7.18.0 (resolves 7.27.2) — CVE-2026-1525, CVE-2026-22036
- `ws` ^7.4.4 → ^8.20.1 — CVE-2026-45736
- `uuid` 8.3.2 → ^11.1.1 — CVE-2026-41907 (`@types/uuid` dropped; types are bundled since v10)
- `@aws-sdk/client-s3` / `lib-storage` refreshed in-range — pulls `fast-xml-parser` 5.7.3 (CVE-2026-25896 critical, + 5 more)

### Transitive re-resolutions (lockfile only, within existing ranges)
`fast-xml-parser` 4.5.5+, `tar-fs`, `ip-address` 10.x (via `socks`), `minimatch`, `picomatch`, `ajv`, `js-yaml`, `@babel/helpers`, `basic-ftp`, `flatted`, `lodash`, `qs`, `follow-redirects`, `brace-expansion`. One resolutions entry forces `minimatch` 9.0.7 for `@typescript-eslint/typescript-estree` (pinned exact at 9.0.3).

### ws v8 runtime fix + new test
ws v8 no longer exposes `Server` on the default export — the screencast server crashed at startup with `ws.Server is not a constructor` (tsc did not catch it). Fixed by using the `WebSocketServer` named export, and added `tests/screencast.test.ts`, which runs a crawl with `--screencastPort` and asserts a WebSocket client receives the `init` message — this path previously had no test coverage and the new test catches the regression.

### Docker image: prune devDependencies
The image ran a plain `yarn install`, shipping jest, eslint, lighthouse, http-server and their transitive trees in production. The Dockerfile now prunes to production dependencies after `tsc` and cleans the yarn cache (runtime code only imports production deps — verified). Smaller image, smaller vulnerability surface.

### Known remaining advisory (accepted)
`fast-xml-parser` CVE-2026-41650 (moderate) via `@webrecorder/wabac`: the fix only exists on the 5.x line and wabac latest still pins `^4.5.4`. Will resolve when wabac moves to fxp 5.

## Testing
- `yarn audit`: 33 unique advisories → 1 moderate (above)
- `yarn tsc`, `eslint`, `prettier` clean
- e2e against the rebuilt image: `basic_crawl.test.ts`, new `screencast.test.ts`, plus S3 upload (versitygw), SOCKS5 proxy (3proxy), and yaml config crawls verified


royteeuwen:
Manual verification of the noVNC 1.4.0 → 1.7.0 upgrade, beyond the new automated test:

Ran `create-login-profile` in the built image with ports 9223/6080 exposed and opened the VNC viewer in Chrome:
- all 53 noVNC ES modules loaded with HTTP 200 via the `/vnc/` route (core, decoders incl. h264, crypto, input, vendored pako)
- RFB WebSocket connected to x11vnc on 6080, password auth accepted
- live framebuffer of the containerized Brave session rendered correctly in the VNC canvas

The RFB API used by `html/vnc_lite.html` (constructor, `sendCredentials`, event listeners) is unchanged between 1.4 and 1.7.

ikreymer:
Thank you for this and the additional testing! We will review today/tomorrow and should be able to do a new release with these updates soon!

mistydemeo:
> ws v8 no longer exposes Server on the default export — the screencast server crashed at startup with ws.Server is not a constructor (tsc did not catch it).

Huh, strange. The release that introduced it suggested it was meant to just be an alias: https://github.com/websockets/ws/releases/tag/8.2.0

And I don't see anything else in the changelog that suggests that's changed. That said the alias is fully supported by them so no reason *not* to use it! I'm just surprised.

mistydemeo:
Tessa dug up the reason we reverted novnc originally - it was #726. This was caused by an undocumented change in the novnc package layout in 1.5.0 which was subsequently reverted (also undocumented) in 1.7.0. We're definitely safe to upgrade to this new version then.

mistydemeo (approved):

ikreymer:
Nice work, thank you for this!
I think the image size may not actually be smaller, we may need to switch to multi-stage builds to benefit from removing dependencies, but still a good step to reduce vulnerability surface, as you mention.

Dockerfile
```diff
@@ -40,6 +40,9 @@ ADD src /app/src
 
 RUN yarn run tsc
 
+# prune devDependencies from the shipped image; runtime only needs production deps
+RUN yarn install --production --frozen-lockfile --network-timeout 1000000 && yarn cache clean
```
mistydemeo, line 44: Just as a note because this switched to prod dependencies and cleaned the cache in a separate layer, it didn't actually reduce the size of the image. We'd need to have done this in the same step that introduced the dependencies for it to have an effect.

royteeuwen: thanks, I missed that one :)

PR #1052, opened 2026-06-20
Add option to combine docker layers and minimize build size

ikreymer:
Follow up to #1045, add a MINIMIZE_IMAGE_SIZE option to combine layers and minimize build size for production builds, eg with `--build-arg MINIMIZE_IMAGE_SIZE=1`
Otherwise, by default, keep separate layers to have faster build time (but larger image size).
Remove puppeteer and lighthouse from devDependencies, not needed.


mistydemeo:
Hm, is there a reason we wouldn't want to always be running this together?

mistydemeo (approved):

PR #1054, opened 2026-06-22
deps: lighthouse 13.4.0/puppeteer 25.2.0

mistydemeo:
#1049 bumped puppeteer-core, but left puppeteer in the devDependencies on the previous version. It looks like our lighthouse dependency was also using puppeteer 24.x. Upgrading both of these gets everything onto the same version.


tw4l:
It looks like https://github.com/webrecorder/browsertrix-crawler/pull/1052/ may be removing `lighthouse` entirely, as it appears to be an unused dependency.

mistydemeo:
Oh, and it's removing puppeteer too. In that case, I'll just close this.

PR #1055, opened 2026-06-22
restrict auto-adding redirected seeds

### LINKED ISSUE #1051: Restrict seed redirects to only same scheme/www. differences
In #476 we added support for automatically adding new seeds if a seed page redirects to a different URL.
However, the current implementation is too broad.
This was intended for, say `http://www.example.com/` -> `https://example.com/` type redirect involving changes in scheme or www. prefix.

Currently, this allows unrelated redirects, such as `https://some-site.example.com/` -> `https://facebook.com/some-site`, which if the user is intending to crawl with `scopeType: prefix`, could result in all of Facebook being crawled. This is undesired behavior.

We should only add a new seed in limited circumstances, otherwise can crawl just the redirect page that was already loaded and not anything else from that seed.

ikreymer:
allow redirected seeds to be added as new seeds only if new URL differs by www. prefix or scheme:
- add new normalizedRedirectSeedUrl() which strips only www. and scheme
- if different redirect, print warning that new seed is not added
- if new seed page URL is in scope of original, continue as before, as crawl should be able to continue with original seed scope

fixes #1051

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -74,6 +74,7 @@ import { initProxy } from "./util/proxy.js";
 import { initFlow, nextFlowStep } from "./util/flowbehavior.js";
 import { isDisallowedByRobots, setRobotsConfig } from "./util/robots.js";
 import { request } from "undici";
+import { normalizedRedirectSeedUrl } from "./util/normalize.js";
 
 const btrixBehaviors = fs.readFileSync(
   new URL(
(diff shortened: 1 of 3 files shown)


src/crawler.ts
```diff
@@ -2505,17 +2506,40 @@ self.__bx_behaviors.selectMainBehavior();
...
-        origUrl: url,
-        newUrl: respUrl,
-        seedId: data.seedId,
-      });
+      const newUrl = respUrl;
+      const origUrl = url;
+      const seedId = data.seedId;
+
+      if (
+        normalizedRedirectSeedUrl(origUrl) == normalizedRedirectSeedUrl(newUrl)
```
tw4l, line 2514: I think we likely also want to pass the new URL to `addExtraSeed` if it isn't not the same seed URL post-normalization but is otherwise in scope, as might be the case for a few reasons (e.g. content moved from one subdomain to another in crawl with `host` scopeType, or to a domain that's covered by `include`s).

tw4l: In these cases I believe the links discovered on the new URL should be in scope either way, but probably better to be consistent, especially since we track which pages come from which seeds.

ikreymer: Hm, but it still comes from the original seed that was redirected? I guess could go either way on this, eg. if `example.com/page1.html` 301 redirects to `example.com/page2.html` with `host` scope, new URLs are should we add page2.html as a seed? since it was technically discovered from both page1.html by way of page2.html (since loading page1 will always redirect to page2)

tw4l: Hm, yeah I guess that's true. I was thinking more of a case where `original.example.com` redirects to `new.example.com`, which may be in scope depending on selected scope type. But that's admittedly edge case-y, and not adding `new.example.com` as a seed in this case if in scope wouldn't break anything.

tw4l (approved):
Changes make sense.

Tested:

- https://www.webrecorder.net: Added new seed of https://webrecorder.net and continued crawl as expected
- http://www.webrecorder.net: Added new seed of https://webrecorder.net and continued crawl as expected
- https://browsertrix.cloud: Crawled redirected page https://webrecorder.net/browsertrix but did not add new seed as expected
- https://browsertrix.com: Crawled redirected page https://webrecorder.net/browsertrix and other pages on browsertrix.com but did not add new seed as expected

A bit tricky to add integration tests here but a unit test for `normalizedRedirectSeedUrl` seems like it would be a good addition.

PR #1058, opened 2026-06-24
Add option to disable media sources

mistydemeo:
This uses wabac's media source script to try to disable media types we currently have trouble playing back, for example YouTube's more complex video formats.

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -792,6 +792,15 @@ class ArgParser {
           type: "number",
           default: -1,
         },
+
+        disableMediaSource: {
+          describe:
+            "If set, runs a script to disable AV1 media types for video site captures",
+          type: "boolean",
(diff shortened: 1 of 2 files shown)


src/util/argParser.ts
```diff
@@ -792,6 +792,15 @@ class ArgParser {
           type: "number",
           default: -1,
         },
+
+        disableMediaSource: {
+          describe:
+            "If set, runs a script to disable AV1 media types for video site captures",
```
ikreymer, line 798:
```suggestion
            "If set, runs a script to disable Media Source API for all pages (canPlayType() returns false)",
```

ikreymer:
Handling this via behaviors instead (webrecorder/browsertrix-behaviors#142 so will not need a new flag, just a doc update via #1062

PR #1065, opened 2026-06-26
dedupe: add --dedupeConcurrent to be able to dedupe concurrent crawls

### LINKED ISSUE #1064: Allow Dedupe for Concurrently Running Crawls
The dedupe implementation requires that crawls be fully finished and 'committed' before the index is available for dedupe, however, in certain circumstances it may be desirable to support concurrent dedupe (with the trade-off that cancelled / failed crawls may result in missing data as other concurrent crawls won't be also crawl the same data that is presumed to be in the first crawl).

There should be a flag to enable this approach.

ikreymer:
crawl cancelation with dedupeConcurrent: track of canceled concurrent crawls in 'allcanceled', to be treated as removed crawls and purged docs: update docs for --dedupeConcurrent trade-offs and update cli page dedupe: account for page urls that redirect by checking for finalPageUrl fixes #1064

### DIFF
--- a/docs/docs/develop/dedupe.md
+++ b/docs/docs/develop/dedupe.md
@@ -62,8 +62,8 @@ Once a crawl is finished, the merged index keys are updated as follows:
 
 - `allcounts`: A sum of all the `h:${crawlid}:counts` for all crawl ids in `allcrawls`. The `allcounts` fields are incremented by each new `h:${crawlid}:counts`.
 
-- `allcanceled`: Used with concurrent crawls only to store crawl ids of crawls that were cancelled or have failed
-while being deduped. If this key exists, it indicates there may be data missing, if it has not been re-crawled.
+- `canceledcrawls`: Use with concurrent crawls only to store crawl ids of crawls that were cancelled or have failed
+while being deduped. If this key exists, indicates there may be data missing, if it has not been re-crawled.
 
 Committing the crawl may take a long time, depending on the size of the crawl. For this reason, the indexer entrypoint
 includes a dedicated command to commit a single crawl, `indexer --commitCrawlId <crawlid>`.
@@ -71,7 +71,7 @@ includes a dedicated command to commit a single crawl, `indexer --commitCrawlId 
 ### Concurrent Crawl Support
 
 If running with `--dedupeConcurrent`, each new hash is automatically committed to `allhashes` and `allcounts` is updated
-for every new hash. If a concurrent crawl is canceled, it is additionally tracked in `allcanceled` to indicate that it
+for every new hash. If a concurrent crawl is canceled, it is additionally tracked in `canceledcrawls` to indicate that it
 needs to be included in the removed crawl count.
 
 #### Dedupe Lookup
--- a/docs/docs/develop/dedupe.md
+++ b/docs/docs/develop/dedupe.md
@@ -65,6 +65,12 @@ Once a crawl is finished, the merged index keys are updated as follows:
 Committing the crawl may take a long time, depending on the size of the crawl. For this reason, the indexer entrypoint
 includes a dedicated command to commit a single crawl, `indexer --commitCrawlId <crawlid>`.
 
+### Concurrent Crawl Support
+
+If running with `--dedupeConcurrent`, each new hash is automatically committed to `allhashes` and `allcounts` is updated
+for every new hash. If a concurrent crawl is canceled, it is additionally tracked in `allcanceled` to indicate that it
+needs to be included in the remove crawl count
+
 #### Dedupe Lookup
 
 To write a revisit record, two Redis lookups are needed:
--- a/docs/docs/develop/dedupe.md
+++ b/docs/docs/develop/dedupe.md
@@ -62,8 +62,8 @@ Once a crawl is finished, the merged index keys are updated as follows:
 
 - `allcounts`: A sum of all the `h:${crawlid}:counts` for all crawl ids in `allcrawls`. The `allcounts` fields are incremented by each new `h:${crawlid}:counts`.
 
-- `allcanceled`: Use with concurrent crawls only to store crawl ids of crawls that were cancelled or have failed
-while being deduped. If this key exists, indicates there may be data missing, if it has not been re-crawled.
+- `allcanceled`: Used with concurrent crawls only to store crawl ids of crawls that were cancelled or have failed
+while being deduped. If this key exists, it indicates there may be data missing, if it has not been re-crawled.
 
 Committing the crawl may take a long time, depending on the size of the crawl. For this reason, the indexer entrypoint
 includes a dedicated command to commit a single crawl, `indexer --commitCrawlId <crawlid>`.
--- a/docs/docs/develop/dedupe.md
+++ b/docs/docs/develop/dedupe.md
@@ -60,7 +60,10 @@ Once a crawl is finished, the merged index keys are updated as follows:
 
 - `allhashes`: `{[hash]: [crawlid]}` - The main merged index. For each hash in `h:${crawlid}`, an entry is added mapping `hash` to the `crawlid`, indicating the hash is found in this crawl. The full data from `h:${crawlid}` is not copied to save space.
 
-- `allcounts`: A sum of all the `h:${crawlid}:counts` for all crawlids in `allcrawls`. The `allcounts` fields are incremented by each new `h:${crawlid}:counts`
+- `allcounts`: A sum of all the `h:${crawlid}:counts` for all crawl ids in `allcrawls`. The `allcounts` fields are incremented by each new `h:${crawlid}:counts`.
+
+- `allcanceled`: Use with concurrent crawls only to store crawl ids of crawls that were cancelled or have failed
+while being deduped. If this key exists, indicates there may be data missing, if it has not been re-crawled.
 
 Committing the crawl may take a long time, depending on the size of the crawl. For this reason, the indexer entrypoint
 includes a dedicated command to commit a single crawl, `indexer --commitCrawlId <crawlid>`.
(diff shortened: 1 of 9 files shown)


docs/docs/develop/dedupe.md
```diff
@@ -65,6 +65,12 @@ Once a crawl is finished, the merged index keys are updated as follows:
 Committing the crawl may take a long time, depending on the size of the crawl. For this reason, the indexer entrypoint
 includes a dedicated command to commit a single crawl, `indexer --commitCrawlId <crawlid>`.
 
+### Concurrent Crawl Support
+
+If running with `--dedupeConcurrent`, each new hash is automatically committed to `allhashes` and `allcounts` is updated
+for every new hash. If a concurrent crawl is canceled, it is additionally tracked in `allcanceled` to indicate that it
+needs to be included in the remove crawl count
```
tw4l:
```suggestion
needs to be included in the removed crawl count.
```

I think this could use a little additional information, it's not immediately clear to me what is tracked in `allcanceled` (the hashes? the crawl?), or what the removed crawl count being referred to here is.

src/util/argParser.ts
```diff
@@ -476,6 +476,13 @@ class ArgParser {
           default: -1,
         },
 
+        dedupeConcurrent: {
+          describe:
+            "If set, immediately add deduped URLs to the dedupe index instead to allow dedupe of concurrently running crawls, but makes cancelling crawls potentially lossy",
```
```suggestion
            "If set, immediately add deduped URLs to the main dedupe index instead of to uncommittedcrawls. This allows dedupe of concurrently running crawls, but makes cancelling or failed crawls potentially lossy",
```

Will also need to regen the cli-options docs

docs/docs/user-guide/cli-options.md
```diff
@@ -203,6 +203,12 @@ Options:
                                             duplicate pages can be skipped. -1 m
                                             eans never skip duplicate pages
                                                           [number] [default: -1]
+      --dedupeConcurrent                    If set, immediately add deduped URLs
```
tw4l, line 206: Regen from argParser changes

docs/docs/user-guide/dedupe.md
```diff
@@ -26,11 +26,17 @@ been archived, it will not be saved again, instead a [`revisit` record will be c
   
     While the index needs to be Redis-compatible, any Redis-compatible database can be used as well without additional changes. [Apache Kvrocks](https://kvrocks.apache.org/) is a good choice for the dedupe index database as it persists the data on disk, instead of keeping it all in memory like Redis. 
 
-When multiple crawls are running at the same time, the resources from one crawl are not yet available to the other crawls.
-This is to account for crawls that may be cancelled or fail.
+
+### Crawl Cancellation and Concurrent Crawl support
+
+Bt default, when multiple crawls are running at the same time, the resources from one crawl are not yet available to the other crawls.
```
tw4l:
```suggestion
By default, when multiple crawls are running at the same time, the resources from one crawl are not yet available to the other crawls.
```

docs/docs/user-guide/dedupe.md
```diff
@@ -26,11 +26,17 @@ been archived, it will not be saved again, instead a [`revisit` record will be c
...
+This is to account for crawls that may be cancelled or fail. If the crawl is cancelled/failed, the data is not entered into the shared dedupe index.
 
 Once a crawl is complete, its data is fully 'committed' to the index and available to be deduplicated against by future crawls. This happens automatically when running the crawler via command line but can also be triggered [via a special indexer command](#committing-finished-crawls-to-the-index).
 
+Sometimes it is necessary to run dedupe concurrently running crawls. The `--dedupeConcurrent` flag can be used to ensure
```
tw4l:
```suggestion
Sometimes it is necessary to deduplicate multiple concurrently running crawls. The `--dedupeConcurrent` flag can be used to ensure
```

docs/docs/user-guide/dedupe.md
```diff
@@ -26,11 +26,17 @@ been archived, it will not be saved again, instead a [`revisit` record will be c
...
+the dedupe index is updated immediately while the crawl is running. As a trade-off, if the crawl is canceled/failed/deleted, there may be data loss as crawler does not write the data on the assumption that duplicate is stored in a concurrently running crawl, but that crawl is canceled/failed, that data may not be there. For this reason, this is not the default behavior, but may be enabled if concurrent crawls will not be canceled and will be retried on failure.
```
tw4l:
```suggestion
the dedupe index is updated immediately while the crawl is running. As a trade-off, there may be data loss if the crawl is cancelled or fails, as other crawls may deduplicate against records from a crawl whose data is discarded and not stored.
For this reason, concurrent deduplication is not the default behavior, but may be enabled if concurrent crawls will not be canceled and will be retried on failure.
```

docs/docs/user-guide/dedupe.md
```diff
@@ -98,9 +104,13 @@ docker run -it webrecorder/browsertrix-crawler indexer --commitCrawlId <crawl-id
 
 The commit process may take a long time if it was a large crawl, but generally should finish quickly.
 
+If running with `--dedupeConcurrent`, no commit operation is needed as the data is automatically committed.
+
 ### Handling interrupted or canceled crawls
 
-If a crawler is interrupted, eg. with SIGINT, the dedupe data stored for that crawl will not yet committed.
+If a crawler is interrupted, eg. with SIGINT, the dedupe data stored for that crawl will not yet committed
```
tw4l:
```suggestion
If a crawler is interrupted, eg. with SIGINT, the dedupe data stored for that crawl will not yet be committed
```

docs/docs/user-guide/dedupe.md
```diff
@@ -114,6 +124,14 @@ docker run -it webrecorder/browsertrix-crawler indexer --cancelCrawlId <crawl-id
 This operational is not required - the data from the interrupted crawl will not be used in further dedupe, it will
 simply free up the data on the Redis.
 
+#### With concurrent crawl support
+
+If the crawl was run with `--dedupeConcurrent`, and it is canceled/failed, the crawl is tracked in a list of removed crawls that
+should be purged. The crawl will be removed when the `indexer --remove` operation is performed.
```
tw4l:
```suggestion
If the crawl was run with `--dedupeConcurrent` and is cancelled or fails, the crawl is tracked in a list of removed crawls that
should be purged. The crawl's data will be removed from the index when the `indexer --remove` operation is performed.
```

docs/docs/user-guide/dedupe.md
```diff
@@ -114,6 +124,14 @@ docker run -it webrecorder/browsertrix-crawler indexer --cancelCrawlId <crawl-id
...
+
+Since other crawl data may have been deduped against the failed crawl, it is possible that there is missing data,
+if the crawl is marked as dependency of another crawl. Rerunning the crawl may patch the missing data in this case.
```
tw4l:
```suggestion
Since other crawl data may have been deduplicated against the failed crawl, it is possible that cancelled or failed crawls may result in missing data
if the crawl is marked as dependency of another crawl. Rerunning the crawl may patch the missing data in this case.
```

^ Relatedly, if a crawl is tracking another concurrent crawl as a dependency, how does that work in relation to the WACZ datapackage dependency tracking and `WARC-Refers-to-Container` headers? Those will just point to nonexistent files?

ikreymer: Right, since we write that the data is in those files, which do not exist due to crawl failing.

docs/docs/develop/dedupe.md
```diff
@@ -65,6 +65,12 @@ Once a crawl is finished, the merged index keys are updated as follows:
 Committing the crawl may take a long time, depending on the size of the crawl. For this reason, the indexer entrypoint
 includes a dedicated command to commit a single crawl, `indexer --commitCrawlId <crawlid>`.
 
+### Concurrent Crawl Support
+
+If running with `--dedupeConcurrent`, each new hash is automatically committed to `allhashes` and `allcounts` is updated
+for every new hash. If a concurrent crawl is canceled, it is additionally tracked in `allcanceled` to indicate that it
```
ikreymer:
```suggestion
for every new hash. If a concurrent crawl is canceled, the crawl is also added to `allcanceled` to indicate that it
```

docs/docs/user-guide/cli-options.md
ikreymer: Done!

docs/docs/develop/dedupe.md
ikreymer: Updated the docs to indicate `allcanceled` contains list of crawl ids that were cancelled in concurrent mode. If this key exists, there's possibility of data missing somewhere. A full purge should be able to determine if that's true or not (and reset this field)

tw4l: Thanks!

docs/docs/develop/dedupe.md
```diff
@@ -60,11 +60,20 @@ Once a crawl is finished, the merged index keys are updated as follows:
 
 - `allhashes`: `{[hash]: [crawlid]}` - The main merged index. For each hash in `h:${crawlid}`, an entry is added mapping `hash` to the `crawlid`, indicating the hash is found in this crawl. The full data from `h:${crawlid}` is not copied to save space.
 
-- `allcounts`: A sum of all the `h:${crawlid}:counts` for all crawlids in `allcrawls`. The `allcounts` fields are incremented by each new `h:${crawlid}:counts`
+- `allcounts`: A sum of all the `h:${crawlid}:counts` for all crawl ids in `allcrawls`. The `allcounts` fields are incremented by each new `h:${crawlid}:counts`.
+
+- `allcanceled`: Use with concurrent crawls only to store crawl ids of crawls that were cancelled or have failed
```
tw4l:
```suggestion
- `allcanceled`: Used with concurrent crawls only to store crawl ids of crawls that were cancelled or have failed
```

docs/docs/develop/dedupe.md
```diff
@@ -60,11 +60,20 @@ Once a crawl is finished, the merged index keys are updated as follows:
...
+while being deduped. If this key exists, indicates there may be data missing, if it has not been re-crawled.
```
tw4l, line 66:
```suggestion
while being deduped. If this key exists, it indicates there may be data missing, if it has not been re-crawled.
```

tw4l (approved):

PR #1066, opened 2026-06-29
Further improvements to awaitPageResources to avoid stuck requests

ikreymer:
- In #1063, addressed requests that were being removed too quickly, but also need to be more careful with pending requests being stuck
- Unset `.intercepting` after handleFetchResponse() since its now being checked before early cleanup
- As an extra precaution, remove requests add a way to track requests that have been 'stuck', eg. have size > 0, but have not changed for 20+ seconds after page has been closed. Hopefully this will not be hit much at all, but adding as extra check.
- Don't mark request as fetchContinued if its a redirect, as request will be repeated with same id.


src/util/recorder.ts
```diff
@@ -1157,6 +1171,33 @@ export class Recorder extends EventEmitter {
         if (reqresp.resourceType) {
           entry.resourceType = reqresp.resourceType;
         }
+        entry.size = reqresp.payload?.length;
+        entry.status = reqresp.intercepting
```
mistydemeo: The double ternary's a little hard to read here; maybe there's an alternate way to represent this?

src/util/recorder.ts
```diff
@@ -1157,6 +1171,33 @@ export class Recorder extends EventEmitter {
...
+          ? "intercepting"
+          : reqresp.asyncLoading
+          ? "async"
+          : "unknown";
+
+        if (reqresp.lastSize === entry.size) {
```
mistydemeo: Worth adding a comment what we're checking for here?

mistydemeo (commented):
Looks like the test's failing.

src/util/recorder.ts
tw4l: Agreed - some indentation or parentheses would go a long way, but I think prettier might currently strip them out when added

ikreymer: yes, the prettier strips that out :/

mistydemeo: Maybe just rework this into a nested `if/else` then?

ikreymer: refactored in toJSON

tw4l (approved):
Nice! Thanks for getting to the redirect issue/failed tests as well

mistydemeo (approved):

PR #1071, opened 2026-07-02
Add --addRedirectedSeeds setting to make seed redirect policy configurable

### LINKED ISSUE #1070: More flexible out of scope redirect handling
In #1051, we restricted out-of-scope seed redirect handling to only changes in www. / scheme, however, this may be something that should instead be configurable.

There could be several options:
- redirectSeedOutOfScope: block: never allow
- redirectSeedOutOfScope strict: only allow for www/scheme
- redirectSeedOutOfScope add: allow for any redirects

Then there's the existing behavior with exclusions what happens to a page that's redirect out of scope
Currently, if a page (not a seed) redirects to out-of-scope, it is loaded, unless specifically excluded.

This perhaps also be configured to:
- allow loading pages that are out of scope, unless excluded
- block pages from loading out of scope, even if not excluded

This could be as separate task, but mentioning here for completeness.

tw4l:
Fixes #1070 

Makes the policy for how to handle seed redirects configurable, while keeping the new behavior introduced in https://github.com/webrecorder/browsertrix-crawler/issues/1051 as the default.

As mentioned in the issue, we may also want to add an option so that the redirected page is not loaded/captured, even if the new seed is not added to the crawl. That has not yet been added in this PR but could be in a follow-up if we think that's an option that will be important to users.

## Testing

Integration tests are a bit tricky since we rely on sites on the live web and we don't necessarily want CI crawling our main website (which can interfere with analytics). The following are some crawl commands and their expected outcomes that can be used to validate the behavior is as expected.

- `docker compose run crawler crawl --limit 3 --url https://browsertrix.com --addRedirectedSeeds always`: Adds new redirected seed https://webrecorder.net/browsertrix due to always policy
- `docker compose run crawler crawl --limit 3 --url https://browsertrix.com`: Does not add redirected seed https://webrecorder.net/browsertrix due to strict policy (implicit)
- `docker compose run crawler crawl --limit 3 --url https://browsertrix.com --addRedirectedSeeds strict`: Does not add redirected seed https://webrecorder.net/browsertrix due to strict policy (explicit)
- `docker compose run crawler crawl --limit 3 --url https://www.webrecorder.net`: Adds redirected seed https://webrecorder.net due to strict policy (implicit)
- `docker compose run crawler crawl --limit 3 --url https://www.webrecorder.net --addRedirectedSeeds strict`: Adds redirected seed https://webrecorder.net due to strict policy (explicit)
- `docker compose run crawler crawl --limit 3 --url https://www.webrecorder.net --addRedirectedSeeds never`: Does not add redirected seed https://webrecorder.net due to never policy

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -803,7 +803,7 @@ class ArgParser {
 
         redirectSeedOutOfScope: {
           describe:
-            "Policy for how to handle seeds that redirect to a URL out of scope. Default (strict): add the redirect URL as a seed only if it differs in scheme or www subdomain. Other options: allow (always add redirect URL as new seed), block (never add redirect URL as new seed).",
+            "Policy for how to handle seeds that redirect to a URL out of scope. Default (strict): add the redirect URL as a seed only if it differs in scheme or www subdomain. Other options: add (always add as new seed), block (never add as new seed)",
           type: "string",
           choices: REDIRECT_OUT_OF_SCOPE_OPTS,
           default: "strict",
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -803,7 +803,7 @@ class ArgParser {
 
         redirectSeedOutOfScope: {
           describe:
-            "Policy for how to handle seeds that redirect to a URL out of scope. Default (strict): add the redirect URL as a seed only if it differs in scheme or www subdomain. Other options: add (always add as new seed), block (never add as new seed)",
+            "Policy for how to handle seeds that redirect to a URL out of scope. Default (strict): add the redirect URL as a seed only if it differs in scheme or www subdomain",
           type: "string",
           choices: REDIRECT_OUT_OF_SCOPE_OPTS,
           default: "strict",
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -21,7 +21,7 @@ import {
   DEFAULT_CRAWL_ID_TEMPLATE,
   RATE_LIMIT_MATCH_200,
   RATE_LIMIT_TTL_SECS,
-  ADD_REDIRECT_SEEDS_OPTS,
+  ADD_REDIRECTED_SEEDS_OPTS,
 } from "./constants.js";
 import { interpolateFilename } from "./storage.js";
 import { screenshotTypes } from "./screenshots.js";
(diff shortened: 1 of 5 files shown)


src/util/argParser.ts
```diff
@@ -799,6 +800,14 @@ class ArgParser {
           type: "number",
           default: -1,
         },
+
+        redirectSeedOutOfScope: {
+          describe:
+            "Policy for how to handle seeds that redirect to a URL out of scope. Default (strict): add the redirect URL as a seed only if it differs in scheme or www subdomain",
```
mistydemeo, line 806: I realize it's redundant with the crawl-scope docs, but I feel like it may be helpful to document what `add`/`block` do here.

mistydemeo (approved):

ikreymer:
Nice, impl looks good, though perhaps the terms should be `always`, `never`, and `strict`?
and flag could be `--addRedirectOutOfScopeSeeds` or something like that, eg so the flag reads `--addRedirectOutOfScopeSeeds always|never|strict` ?

For the blocking/never, the page isn't actually blocked from loading, it'll still load the single page, like other out of scope pages, but is not added as a new seed.

This is where the other flag mentioned in #1070 could come in where we control that, maybe `--loadRedirectOutOfScopePages load|block` where that specifies what happens to out of scope pages..

Could also add a shortcut to 'OOS' for these flag, eg. `--addRedirectOOSSeeds` perhaps?

ikreymer:
> ... renaming idea

Took a quick pass at renaming the opts, here: https://github.com/webrecorder/browsertrix-crawler/pull/1079

The load redirect blocking can come back to, as realized loading of seed page is needed to determine what sort of redirect it is..

tw4l:
@ikreymer I merged your suggestions but then looked at our other flags and noticed they pretty much all start with the action - e.g. `failOnContentCheck`, not `contenCheckFailCrawl`.

So rather than `outOfScopeRedirectAddSeeds`, what if we kept it simpler and more consistent with something like `addRedirectSeeds`? I think the "out of scope" is implied and documented here so doesn't have to be in the name necessarily - otherwise there'd be no reason to add a seed.

ikreymer:
> @ikreymer I merged your suggestions but then looked at our other flags and noticed they pretty much all start with the action - e.g. `failOnContentCheck`, not `contenCheckFailCrawl`.
> 
> So rather than `outOfScopeRedirectAddSeeds`, what if we kept it simpler and more consistent with something like `addRedirectSeeds`? I think the "out of scope" is implied and documented here so doesn't have to be in the name necessarily - otherwise there'd be no reason to add a seed.

Sure, yeah, I guess can go with `addRedirectSeeds` or `addRedirectingSeeds`. Also need to update the docs and cli opts to match.

src/util/argParser.ts
ikreymer: Now the options are always, never and strict, so more self-explanatory imo

ikreymer (approved):

ikreymer:
`addRedirectedSeeds` makes sense, clearer that seeds already redirected.

tw4l:
> `addRedirectedSeeds` makes sense, clearer that seeds already redirected.

Glad you agree!

mistydemeo (approved):

PR #1072, opened 2026-07-02
docs: add explicit documentation for site-specific behaviours

mistydemeo:
We previously pointed users at the browsertrix-behaviors repo, but that repo didn't actually have substantive documentation about what they could do. Let's unify the docs here in one place along with the other places that users of the crawler are consulting.

### DIFF
--- a/docs/docs/user-guide/behaviors.md
+++ b/docs/docs/user-guide/behaviors.md
@@ -16,12 +16,66 @@ There is also a built-in 'main' behavior, which runs to completion (or until a t
 
 ## Site-Specific Behaviors
 
-Browsertrix also comes with several 'site-specific' behaviors, which run only on specific sites. These behaviors will run instead of Autoscroll and will run until completion or timeout. Currently, site-specific behaviors include major social media sites.
+Browsertrix also comes with several 'site-specific' behaviors, which run only on specific sites. These behaviors will run instead of Autoscroll and will run until completion or timeout. Currently, site-specific behaviors include the following major social media sites:
 
-Refer to [Browsertrix Behaviors](https://github.com/webrecorder/browsertrix-behaviors) for the latest list of site-specific behaviors.
+- Bluesky
+- Facebook
+- Instagram
+- Telegram
+- TikTok
+- Twitter/X
 
 User-defined custom behaviors are also considered site-specific.
- 
+
+### Bluesky
+
+The Bluesky behavior loads additional JavaScript artifacts in order to ensure that Bluesky posts can be replayed with the user interface in any language.
+
+### Facebook
+
+The Facebook behavior is able to automate the following features:
+
+- Viewing posts from the timeline for an individual, organization, or group page
+- Expanding the comment section in a post or photo in order to load additional comments
+- Browsing the photos grid and viewing full size versions of posts and their descriptions
+- Browsing all reels (videos) on a page
+
+The Facebook behavior will use `addLink` to add single-page versions of posts encountered when crawling a profile to the crawl queue. For more information on `addLink` and scoping, [see below](#additional-links-from-behaviors).
+
+### Instagram
+
+The Instagram behavior is able to automate the following features:
+
+- Viewing all photos/videos on a single post
+- Viewing all media from a profile
+- Viewing the currently-active story (requires a logged-in profile)
+- Viewing story highlights
+- Expanding the comment section in order to load additional comments
+
+The Instagram behavior will use `addLink` to add stories and single-page versions of posts encountered when crawling a profile to the crawl queue. For more information on `addLink` and scoping, [see below](#additional-links-from-behaviors).
+
+### Telegram
+
+The Instagram behavior is intended to work with public Telegram channels. It supports the following features:
+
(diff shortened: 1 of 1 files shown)


docs/docs/user-guide/behaviors.md
```diff
@@ -16,12 +16,66 @@ There is also a built-in 'main' behavior, which runs to completion (or until a t
...
+
+The Instagram behavior is able to automate the following features:
+
+- Viewing all photos/videos on a single post
+- Viewing all media from a profile
+- Viewing the currently-active story (requires a logged-in profile)
+- Viewing story highlights
+- Expanding the comment section in order to load additional comments
+
+Additionally, the Facebook behavior will use smart scoping to crawl single-page versions of posts encountered when crawling a profile.
```
ikreymer:
```suggestion
Additionally, the Instagram behavior will use smart scoping to crawl single-page versions of posts encountered when crawling a profile.
```

ikreymer: Should we list additional pages that are added as single pages from Instagram?

ikreymer: Also, maybe its worth defining what we mean by 'Smart Scoping' in a section as part of behaviors, eg. expanding the scope as needed to include from addLink, in

mistydemeo: I can link out to the main browsertrix docs for it once those are in: https://github.com/webrecorder/browsertrix/pull/3443/files#diff-ce197e712fb7beaa93c9d9caf3f94ef267d4274bb60cc919dd7e5a1cb7f319cdR195

docs/docs/user-guide/behaviors.md
```diff
@@ -16,12 +16,66 @@ There is also a built-in 'main' behavior, which runs to completion (or until a t
...
-Refer to [Browsertrix Behaviors](https://github.com/webrecorder/browsertrix-behaviors) for the latest list of site-specific behaviors.
+- Bluesky
+- Facebook
+- Instagram
+- Telegram
+- TikTok
+- Twitter/X
 
 User-defined custom behaviors are also considered site-specific.
-
```
ikreymer, line 24: I think somewhere in this file we should also explain the `--alwaysAddBehaviorLinks` flag, maybe something like this?

```
### Additional Links from Behaviors

Many of the behaviors include queue additional pages that should be loaded for more complete capture, via the `addLink()` function. However, these page may not always be in scope, and user may not know what these are. The `--alwaysAddBehaviorLinks` flags allows for behaviors to extend scope as needed, including behavior specific links (for example additional pages for single page crawl), otherwise links are only added if they are already in the scope of the crawl. This is referred to as 'smart scoping' in the Browsertrix app.
```

mistydemeo: Pushed a change.

docs/docs/user-guide/behaviors.md
```diff
@@ -16,12 +16,66 @@ There is also a built-in 'main' behavior, which runs to completion (or until a t
...
+
+The Instagram behavior is able to automate the following features:
+
+- Viewing all photos/videos on a single post
+- Viewing all media from a profile
+- Viewing the currently-active story (requires a logged-in profile)
+- Viewing story highlights
+- Expanding the comment section in order to load additional comments
+
+When `--alwaysAddBehaviorLinks` is passed, the Instagram behavior will use `addLink` to crawl single-page versions of posts encountered when crawling a profile.
```
ikreymer: addLink is always called even if alwaysAddBehaviorLinks is not provided, maybe can just say something
lie:
Additional Pages: Uses addLink to crawl single-page versions of posts...

mistydemeo: How about:

> The Instagram behavior will use `addLink` to add single-page versions of posts encountered when crawling a profile to the crawl queue. Note that, depending on the scope, these may not be crawled unless `--alwaysAddBehaviorLinks` is passed.

ikreymer: Could also say 'See below on more info about addLink` and link to that section so we don't have to repeat that..

mistydemeo: Done!

docs/docs/user-guide/behaviors.md
```diff
@@ -16,12 +16,66 @@ There is also a built-in 'main' behavior, which runs to completion (or until a t
...
+
+The Instagram behavior is able to automate the following features:
+
+- Viewing all photos/videos on a single post
+- Viewing all media from a profile
+- Viewing the currently-active story (requires a logged-in profile)
+- Viewing story highlights
+- Expanding the comment section in order to load additional comments
+
+The Instagram behavior will use `addLink` to add single-page versions of posts encountered when crawling a profile to the crawl queue. For more information on `addLink` and scoping, [see below](#additional-links-from-behaviors).
```
ikreymer: Should we mention everything that is added via addLink, eg. stories?

mistydemeo: Good call, done!

ikreymer (approved):

PR #1074, opened 2026-07-03
Ensure direct fetched resources always result in revisit records

### LINKED ISSUE #1073: Non-HTML pages which are duplicates are treated as failed pages.
If a non-HTML resources (eg. PDF/image) is captured in the course of crawling, it is skipped within the same crawl.
However, if the resource is later identified as a page, it is erroneously treated as a failed page, since no WARC record is written.
Probably the best option is to always write a revisit record (that's what they're for, after all!) if a previously captured resource is also treated as a top-level page, and thus avoid a failed page, and ensure each page corresponds to a WARC record.

ikreymer:
- Refactors serializeWARC to differentiate between 'skipped not writing' and 'skipped already written' (duplicate) to clarify intent
- Adds 'alwaysWriteRevisit' to ensure a revisit is always written when 'skipped already written' would be returned.
- For direct fetches, always writes revisit records since the page will correspond to that record.
- Fixes #1073 

Testing:
- Run a crawl of a site that has lots of .pdf images linked as pages, but also crawled internally
- Observe these pages are in the CDX, but are added to extraPages.json with loadState 0 (failed)
- With this PR, should no longer be the case.

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -122,8 +122,11 @@ enum SerializeRes {
   // WARC record writing aborted due to incomplete, should retry
   Aborted = 1,
 
-  // WARC record skipped (eg. dupe, cached) and should not be retried
-  Skipped = 2,
+  // WARC record skipped because it is already been written
+  SkippedDupe = 2,
+
+  // WARC record skipped because it should not be written
+  SkippedNoWrite = 3,
 }
 
 // =================================================================
@@ -1726,8 +1729,9 @@ export class Recorder extends EventEmitter {
     skipPageInfo = false,
     matchHash?: string,
   ): Promise<SerializeRes> {
-    if (this.skipRecordingPage) {
-      return SerializeRes.Skipped;
+    // skip here if skipping recording page, and not asyncLoading outside of page
+    if (this.skipRecordingPage && !reqresp.asyncLoading) {
+      return SerializeRes.SkippedNoWrite;
     }
 
     // always include in pageinfo record if going to serialize to WARC
@@ -1746,7 +1750,7 @@ export class Recorder extends EventEmitter {
         { url, status },
         "recorder",
       );
-      return SerializeRes.Skipped;
+      return SerializeRes.SkippedDupe;
     } else if (!iter && reqresp.shouldSkipSave()) {
       logger.debug(
         "Skipping writing request/response",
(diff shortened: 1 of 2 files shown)


src/util/recorder.ts
```diff
@@ -906,7 +909,10 @@ export class Recorder extends EventEmitter {
           requestId,
           errorReason: "BlockedByResponse",
         });
-        await this.serializeToWARC(reqresp, undefined, false, true, hash);
+        await this.serializeToWARC(reqresp, {
+          canRetry: true,
```
ikreymer:
```suggestion
          skipPageInfo: true,
```

src/util/recorder.ts
```diff
@@ -1571,8 +1577,10 @@ export class Recorder extends EventEmitter {
       try {
         // if aborted, allow retrying
         if (
-          (await this.serializeToWARC(reqresp, iter, true)) ===
-          SerializeRes.Aborted
+          (await this.serializeToWARC(reqresp, {
+            iter,
+            skipPageInfo: true,
```
ikreymer:
```suggestion
            canRetry: true,
```

src/util/recorder.ts
```diff
@@ -122,8 +122,11 @@ enum SerializeRes {
   // WARC record writing aborted due to incomplete, should retry
   Aborted = 1,
 
-  // WARC record skipped (eg. dupe, cached) and should not be retried
-  Skipped = 2,
+  // WARC record skipped because it is already been written
+  SkippedDupe = 2,
```
mistydemeo, line 126: Is it safe to redefine this item in the enum? Not sure what compatibility looks like.

ikreymer: This particular enum is not exported so it's for internal use only within this file.

src/util/recorder.ts
```diff
@@ -1721,13 +1729,23 @@ export class Recorder extends EventEmitter {
 
   async serializeToWARC(
     reqresp: RequestResponseInfo,
-    iter?: AsyncIterable<Uint8Array>,
-    canRetry = false,
-    skipPageInfo = false,
-    matchHash?: string,
+    {
```
tw4l, line 1732: This is a nice change

tw4l (approved):
Tested, changes look good. Thanks for the test updates!

PR #1078, opened 2026-07-04
fault tolerance: allow ignoring failed per-host proxies

### LINKED ISSUE #1076: Crawler fault tolerance: allow ignoring failed matchHosts proxies
Using the matchHosts proxy system, multiple proxies may be specified. It should be possible to ignore these proxies if they are not available, since they only apply to certain hosts (which may not even be part of a crawl).
A global setting in the proxies config file alongside matchHosts should be sufficient.

ikreymer:
For failed proxy connections:
- interrupt instead of failing crawl

For per-host proxies specified in matchHosts:
- only attempt to init host proxies that are actually in use
- add optional `ignoreOnFailedSSHTunnel` to each proxy config to allow ignoring SSH proxies specified via `matchHosts` proxies if they are not available, and just connecting directly, instead of interrupting crawl. This only applies to SSH proxies and only when used with `matchHosts`

fixes #1076

### DIFF
--- a/docs/docs/user-guide/proxies.md
+++ b/docs/docs/user-guide/proxies.md
@@ -128,7 +128,7 @@ The same `--proxyServerConfig` option can also be used in browser profile creati
 
 ### SSH Tunnels
 
-For any SSH proxy that is listed as a possible proxy in `matchHosts` configuration, the crawler will attempt to set up an SSH tunnel before starting the crawl. By default, the crawl will not start and will be interrupted if any of the SSH tunnel connections fail. To allow ignoring failing SSH proxies (and perform a direct connection) the `ignoreOnFailedSSHTunnel: true` can be set for each SSH proxy that should be ignored on failure.
+For any SSH proxy that is listed as a possible proxy in `matchHosts` configuration, the crawler will attempt to set up an SSH tunnel before starting the crawl. By default, the crawl will not start and will be interrupted if any of the SSH tunnel connections fail. To allow ignoring failing SSH proxies and perform a direct connection instead, `ignoreOnFailedSSHTunnel: true` can be set for each SSH proxy that should be ignored on failure.
 
 SOCKS5 and HTTP proxy connections are made by the browser only on first use, and are not checked for successful connection until they are used.
 
--- a/docs/docs/user-guide/proxies.md
+++ b/docs/docs/user-guide/proxies.md
@@ -121,14 +121,15 @@ files, the crawler can be started with:
 docker run -v $PWD/crawls:/crawls -v $PWD/proxies:/proxies -it webrecorder/browsertrix-crawler --url https://example.com/ --proxyServerConfig /proxies/proxyConfig.yaml
 ```
 
-Note that if SSH proxies are provided, an SSH tunnel must be opened for each one before the crawl starts.
-The crawl will not start if any of the SSH proxy connections fail, even if a host-specific proxy is not yet used.
-The `--proxyServerConfigIgnoreFailedProxies` flag can be provided to silently ignore
-any failed proxies and connect directly.
-
-SOCKS5 and HTTP proxy connections are attempted only on first use.
-
 The same `--proxyServerConfig` option can also be used in browser profile creation with the `create-login-profile` command in the same way.
+
+### SSH Tunnels
+
+For any SSH proxy that is listed as a possible proxy in `matchHosts` configuration, the crawler will attempt to set up an SSH tunnel before starting the crawl. By default, the crawl will not start if any of the SSH tunnel connections fail and the crawl
+will be interrupted. However, the `--proxyServerConfigIgnoreFailedProxies` flag can be provided to ignore any failing
+proxies and use a direct connection instead.
+
+SOCKS5 and HTTP proxy connections are made by the browser only on first use, and are not checked for successful connection until/if they are used.
 
 ### Proxy Precedence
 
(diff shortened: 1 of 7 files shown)


src/util/argParser.ts
```diff
@@ -703,10 +703,17 @@ class ArgParser {
...
         proxyServerConfig: {
           describe:
-            "if set, path to yaml/json file that configures multiple path servers per URL regex",
+            "if set, path to yaml/json file that configures multiple proxy servers per URL regex",
           type: "string",
         },
 
+        proxyServerConfigIgnoreFailedProxies: {
+          describe:
+            "if set, and --proxyServerConfig is used, any of the per URL regex proxy servers if they are not accessible",
```
tw4l: Seems like this description is missing an operative word or two

docs/docs/user-guide/proxies.md
```diff
@@ -121,12 +121,16 @@ files, the crawler can be started with:
...
-
 The same `--proxyServerConfig` option can also be used in browser profile creation with the `create-login-profile` command in the same way.
 
+### SSH Tunnels
+
+For any SSH proxy that is listed as a possible proxy in `matchHosts` configuration, the crawler will attempt to set up an SSH tunnel before starting the crawl. By default, the crawl will not start if any of the SSH tunnel connections fail and the crawl
+will be interrupted. However, the `--proxyServerConfigIgnoreFailedProxies` flag can be provided to ignore any failing
+proxies and use a direct connection instead.
+
+SOCKS5 and HTTP proxy connections are made by the browser only on first use, and are not checked for successful connection until/if they are used.
```
tw4l:
```suggestion
SOCKS5 and HTTP proxy connections are made by the browser only on first use, and are not checked for successful connection until they are used.
```

The "if" was adding a lot of ambiguity here.

src/util/argParser.ts
```diff
@@ -703,10 +703,17 @@ class ArgParser {
 
         proxyServerConfig: {
           describe:
-            "if set, path to yaml/json file that configures multiple path servers per URL regex",
+            "if set, path to yaml/json file that configures multiple proxy servers per URL regex",
           type: "string",
         },
 
+        proxyServerConfigIgnoreFailedProxies: {
```
tw4l: The name is quite long - I think we can probably drop the last word and make it `proxyServerConfigIgnoreFailed` without affecting legibility if we want to shorten it a bit

src/util/argParser.ts
tw4l:
```suggestion
            "if set, ignore any per-URL regex proxy servers set with --proxyServerConfig if they are not available rather than interrupting crawl",
```

tw4l (approved):
Left a few small suggestions, working well. Thanks for the tests!

src/util/argParser.ts
ikreymer: Updated the flag to `proxyServerConfigIgnore`

ikreymer:
I wonder, if perhaps avoiding the flag and instead putting this config into the proxies config itself is better, either per-proxy or globally, eg.:


```
matchHosts:
  example.com: ssh-proxy
  
ssh-proxy:
   ...
   ignoreFailedSSH: true
```

or

```
matchHosts:
  example.com: ssh-proxy

ssh-proxy:
   ...

ignoreFailedSSHProxies: true
```

ikreymer:
Updated to use `ignoreOnFailedSSHTunnel` per proxy config.

docs/docs/user-guide/proxies.md
```diff
@@ -125,9 +128,7 @@ The same `--proxyServerConfig` option can also be used in browser profile creati
 
 ### SSH Tunnels
 
-For any SSH proxy that is listed as a possible proxy in `matchHosts` configuration, the crawler will attempt to set up an SSH tunnel before starting the crawl. By default, the crawl will not start if any of the SSH tunnel connections fail and the crawl
-will be interrupted. However, the `--proxyServerConfigIgnoreFailed` flag can be provided to ignore any failing
-proxies and use a direct connection instead.
+For any SSH proxy that is listed as a possible proxy in `matchHosts` configuration, the crawler will attempt to set up an SSH tunnel before starting the crawl. By default, the crawl will not start and will be interrupted if any of the SSH tunnel connections fail. To allow ignoring failing SSH proxies (and perform a direct connection) the `ignoreOnFailedSSHTunnel: true` can be set for each SSH proxy that should be ignored on failure.
```
tw4l:
```suggestion
For any SSH proxy that is listed as a possible proxy in `matchHosts` configuration, the crawler will attempt to set up an SSH tunnel before starting the crawl. By default, the crawl will not start and will be interrupted if any of the SSH tunnel connections fail. To allow ignoring failing SSH proxies and perform a direct connection instead, `ignoreOnFailedSSHTunnel: true` can be set for each SSH proxy that should be ignored on failure.
```

docs/docs/user-guide/proxies.md
```diff
@@ -112,6 +112,9 @@ proxies:
   default-proxy:
     url: ssh://user@my-social-proxy.example.com
     privateKeyFile: /proxies/default-proxy-private-key
+
+    # optional, if set, ignore this proxy if tunnel can not be established at crawl start time
+    ignoreOnFailedSSHTunnel: true
```
tw4l, line 117: Should this be more generic, e.g. `ignoreOnFailedConnection`, so that it can apply for any proxy, not just ones connecting via ssh? A non-ssh SOCKS5 proxy might still fail to connect at times as well.

tests/proxy.test.ts
tw4l:
```suggestion
test("proxy per-host, ignore only if option set", () => {
```

docs/docs/user-guide/proxies.md
ikreymer: This is intentional since non-SSH proxies don't result in a tunnel at crawl start, and can't really be ignored, the connection from the browser will just fail at a later time.

tw4l (approved):

PR #1080, opened 2026-07-07
Rate Limit Docs + more flexible custom rules

ikreymer:
Follow-up to #1015:
- Replace --rateLimitOn200MatchText with --rateLimitOnMatch to allow specifying either a '<regex> or '<regex>:<status_code>' to match to determine if a page is rate limited.
- Add docs for rate limit detection.


src/util/argParser.ts
```diff
@@ -764,18 +765,18 @@ class ArgParser {
...
           type: "array",
-          default: RATE_LIMIT_MATCH_200,
+          default: [403, 429, 503],
         },
 
-        rateLimitStatusCodes: {
+        rateLimitOnMatch: {
           describe:
-            "Consider responses with these status codes to be treated as rate-limited/blocked responses",
+            "One or more rules of format <regex> or <regex>:<status code> to match on page to mark page as rate limited. If no status provided, matches only apply for that status code",
```
tw4l: "If no status provided, matches only apply for that status code" is a bit ambiguous. If no status is provided, it looks like we fall back to matching on 200, yeah? Maybe better to be explicit

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
+# Rate Limit / CAPTCHA Page Detection
+
+When crawling, it is often possible to encounter rate limits, CAPTCHA or auth pages that are shown instead of valid content that crawler is intending to capture.
```
tw4l:
```suggestion
When crawling, it is often possible to encounter rate limits, CAPTCHAs, or auth pages that are shown instead of valid content that the crawler is intending to capture.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+These 'auth-walled' pages are often done for a good reason, and bypassing such limits if outside of the scope of the crawler.
```
tw4l:
```suggestion
These 'auth-walled' pages are often implemented for good reason, and bypassing such limits is outside of the scope of the crawler.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+However, the crawler can *detect* such pages and avoid archiving 'bad'/undesirable data.
+
+The crawler includes several options to help with this. 
+
+When pages are flagged as rate limited, the pages are still loaded in the browser, but the content is not archived.
+Instead, the pages may be queued to be retried at a later time, or skipped altogether.
```
tw4l:
```suggestion
Instead, the pages may be queued to be retried at a later time or skipped altogether.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Flagging Rate Limited / Auth-walled pages
+
+The simplest mechanism for detection is by checking the status code. By default, the crawler now skips all pages that have a 403, 429, 503 status codes. This list can be customized with the `--rateLimitStatusCodes` which can be set to a list of status codes to always skip and retry.
```
tw4l:
```suggestion
The simplest mechanism for detection is by checking the status code. By default, the crawler now skips all pages that have a 403, 429, or 503 status code. This list can be customized with the `--rateLimitStatusCodes` setting, which accepts a list of status codes to always skip and retry.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+The crawler also includes a custom option, `--rateLimitOnMatch` which can take one more values in the form of
+`<regex>` or `<regex>:<status>`, for example, `--rateLimitOnMatch <regex-1> --rateLimitOnMatch <regex-2>:<status>`.
+
+This setting allows for custom detection of CAPTCHA / rate-limited pages based on page content, and optionally status code,
+so even 200 pages can be flagged as rate limited.
```
tw4l:
```suggestion
The crawler also includes a custom option `--rateLimitOnMatch`, which can take one more values in the form of `<regex>` or `<regex>:<status>`, e.g. `--rateLimitOnMatch <regex-1> --rateLimitOnMatch <regex-2>:<status>`. This setting allows for custom detection of CAPTCHAs and rate-limited pages based on page content, and optionally status code, so even 200 pages can be flagged as rate limited.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Rate Limit Counters
+
+The crawler also keeps track of how many pages have been flagged as limited during the duration of the crawl,
+by combination of status code and direct fetch/browser-based indicator. These stats are stored in Redis as part of the crawl stats.
+
+The crawler also keeps track of how many pages have been rate limited *within the last N seconds*, set by the `--rateLimitTimeout` and defaulting to 300 seconds (5 minutes).
+
+If a site provides a `Retry-After` header, for example, with a 429 response, the timeout is used instead of the global default set in `--rateLimitTimeout`.
```
tw4l:
```suggestion
If a site provides a `Retry-After` header, for example with a 429 response, the header value is used instead of the global default set in `--rateLimitTimeout`.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Rate Limit Retry Count
+
+By default, rate limited pages are retried indefinitely, allowing the crawl to possibly complete, but at a much slower pace. If the pages are blocked do to auth requirements, it may be possible to add a browser profile and restart the crawl, allowing for previously flagged pages to be captured successfully.
```
tw4l:
```suggestion
By default, rate limited pages are retried indefinitely, allowing the crawl to possibly complete but at a much slower pace. If the pages are blocked due to auth requirements it may be possible to add a browser profile and restart the crawl, allowing for previously flagged pages to be captured successfully.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+However, it may be desirable to set a total number of retries for pages flagged as rate limited, which can be done
+with `--rateLimitMaxRetries` to a value >= 0, where 0 implies no retries at all.
```
tw4l:
```suggestion
However, it may be desirable to set a total number of retries for pages flagged as rate limited, which can be done
by setting `--rateLimitMaxRetries` to a value >= 0, where 0 implies no retries at all.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Interrupting Crawl on Rate Limit Threshold
+
+By default, the crawler will continue, skipping rate limited pages and retrying them indefinitely, to match existing behavior.
+
+If the `--rateLimitInterruptCount M` flag is set, the crawler will exit with a rate limit exit code (exit code 21) after M rate limited pages within the N seconds, configured via `--rateLimitTimeout`.
+
+This can allow another application (eg. Kubernetes) to provide an exponential backoff system when the crawler is repeatedly exiting due to rate limited pages.
```
tw4l:
```suggestion
This can allow another application (e.g. Kubernetes) to provide an exponential backoff system when the crawler is repeatedly exiting due to rate limited pages.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+Additionally, if direct fetch reaches this threshold, further direct fetches will also be skipped until the timeout expires.
+
+If a page loads successfully, the rate limit counter is also cleared.
```
tw4l:
```suggestion
If a page loads successfully, the rate limit counter is cleared.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Caveats and multi-domain crawling
```
tw4l:
```suggestion
## Caveats and Multi-Domain Crawling
```

For consistency with other headers

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+The rate limit interrupt is disabled by default, as it may not always be desirable for crawling many seeds across different domains crawling, where one site may be rate limited while others' are not. Currently, the rate limits are applied globally and not per-domain.
```
tw4l:
```suggestion
Interrupting the crawl when the rate limit threshold is reached is disabled by default as it may not always be desirable when crawling many seeds across different domains, where one site may be rate limited while others are not. Currently, the rate limits are applied globally and not per-domain.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+For multi-domain crawling, the threshold could be set to a large number, so that if multiple domains are flagged as rate limited, the crawl is still interrupted, while for a single domain crawl, the number may be set lower.
```
tw4l:
```suggestion
For multi-domain crawling the threshold could be set to a large number so that if multiple domains are flagged as rate limited the crawl is still interrupted, while for a single domain crawl the number may be set lower.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+It may make sense to set rateLimitMaxRetries to a fixed value so that larger crawls do eventually finish with certain pages skipped as rate limited.
```
tw4l:
```suggestion
It may make sense to set `--rateLimitMaxRetries` to a fixed value so that larger crawls do eventually finish with certain pages skipped as rate limited.
```

docs/docs/user-guide/rateLimits.md
```diff
@@ -0,0 +1,63 @@
...
+
+Finally, depending on the reason for the 'auth-wall' on each domain, the crawler may or may not be able to complete the crawl on each domain.
+- For 429 errors and 503 errors, waiting and retrying may be successful.
+- For 403 errors, logging in with a browser profile may provide additional access.
+- Other sites and custom CAPTCHA pages may require additional permission to be allowed to crawl the sites.
+
+If facing repeated errors, requesting permission may be the best approach to successful crawl a domain.
```
tw4l:
```suggestion
If facing repeated errors, requesting permission may be the best approach to successfully crawl a domain.
```

docs/mkdocs.yml
```diff
@@ -55,6 +55,7 @@ nav:
     - user-guide/browser-profiles.md
     - user-guide/dedupe.md
     - user-guide/proxies.md
+    - user-guide/rateLimits.md
```
tw4l: Should rename this file to `rate-limits.md` to be consistent.

tw4l (changes requested):

src/util/argParser.ts
ikreymer: It's now set to match any status if none is provided (initially was just 200)

ikreymer:
```suggestion
            "One or more rules in the format <regex> or <regex>:<status code> to match on page text via regex to flag the page as rate limited. If a status is provided, only matches if the response has the specified status",
```

ikreymer: @tw4l how about this?

src/util/argParser.ts
```diff
@@ -966,6 +967,23 @@ class ArgParser {
       logger.info("Updating profile on successful crawl");
     }
 
+    if (argv.rateLimitOnMatch) {
+      const rules: RateLimitRule[] = [];
+      for (const rule of argv.rateLimitOnMatch) {
+        let status = 200;
```
tw4l: Re: other comment, seems like it'll default to status of 200 here and only match on that if no explicit status is provided?

src/util/argParser.ts
tw4l:
```suggestion
            "One or more rules in the format <regex> or <regex>:<status code> matched against page text to determine if a page is rate limited. If a status is provided, only matches if the response has the specified status",
```

How about this?

src/util/argParser.ts
ikreymer: Good catch, fixed that to 0

src/util/argParser.ts
ikreymer: Should it be `if a page` too?

docs/docs/user-guide/rate-limits.md
```diff
@@ -0,0 +1,59 @@
...
+The crawler includes several options to help with this. 
+
+When pages are flagged as rate limited, the pages are still loaded in the browser, but the content is not archived.
+Instead, the pages may be queued to be retried at a later time or skipped altogether.
+
+## Flagging Rate Limited / Auth-walled pages
+
+The simplest mechanism for detection is by checking the status code. By default, the crawler now skips all pages that have a 403, 429, or 503 status code. This list can be customized with the `--rateLimitStatusCodes` setting, which accepts a list of status codes to always skip and retry.
+
+The crawler also includes a custom option `--rateLimitOnMatch`, which can take one more values in the form of `<regex>` or `<regex>:<status>`, e.g. `--rateLimitOnMatch <regex-1> --rateLimitOnMatch <regex-2>:<status>`. This setting allows for custom detection of CAPTCHAs and rate-limited pages based on page content, and optionally status code, so even 200 pages can be flagged as rate limited.
```
mistydemeo:
```suggestion
The crawler also includes a custom option `--rateLimitOnMatch`, which can take additional values in the form of `<regex>` or `<regex>:<status>`, e.g. `--rateLimitOnMatch <regex-1> --rateLimitOnMatch <regex-2>:<status>`. This setting allows for custom detection of CAPTCHAs and rate-limited pages based on page content, and optionally status code, so even 200 pages can be flagged as rate limited.
```

mistydemeo: We should at least briefly mention *what* the regex matches.

src/util/argParser.ts
tw4l: Updated

docs/docs/user-guide/rate-limits.md
```diff
@@ -0,0 +1,59 @@
...
+However, it may be desirable to set a total number of retries for pages flagged as rate limited, which can be done
+by setting `--rateLimitMaxRetries` to a value >= 0, where 0 implies no retries at all.
+
+## Interrupting Crawl on Rate Limit Threshold
+
+By default, the crawler will continue, skipping rate limited pages and retrying them indefinitely, to match existing behavior.
+
+If the `--rateLimitInterruptCount M` flag is set, the crawler will exit with a rate limit exit code (exit code 21) after M rate limited pages within the N seconds, configured via `--rateLimitTimeout`.
+
+This can allow another application (e.g. Kubernetes) to provide an exponential backoff system when the crawler is repeatedly exiting due to rate limited pages.
```
mistydemeo: Maybe clarify that this allows an application/job runner which is operating the crawler to provide such a backoff.

docs/docs/user-guide/rate-limits.md
```diff
@@ -0,0 +1,59 @@
...
+
+Additionally, if direct fetch reaches this threshold, further direct fetches will also be skipped until the timeout expires.
+
+If a page loads successfully, the rate limit counter is cleared.
```
mistydemeo: Does this mean any page?

docs/mkdocs.yml
ikreymer: Renamed!

docs/docs/user-guide/rate-limits.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Caveats and Multi-Domain Crawling
+
+Interrupting the crawl when the rate limit threshold is reached is disabled by default as it may not always be desirable when crawling many seeds across different domains, where one site may be rate limited while others are not. Currently, the rate limits are applied globally and not per-domain.
```
mistydemeo: Might be helpful to flip the order of this a bit to clarify:

```suggestion
Because rate limits are applied globally and not per-domain, and it may not always be desirable to interrupt a crawl containing many seeds across different domains, the crawl will not be automatically interrupted when the rate limit threshold is reached.
```

ikreymer: Well, there is no interrupt theshold by default for this reason, but it will be interrupt when it is reached, can reword.

docs/docs/user-guide/rate-limits.md
ikreymer: yes, updated to say 'any'

docs/docs/user-guide/rate-limits.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Interrupting Crawl on Rate Limit Threshold
+
+By default, the crawler will continue, skipping rate limited pages and retrying them indefinitely, to match existing behavior.
+
+If the `--rateLimitInterruptCount M` flag is set, the crawler will exit with a rate limit exit code (exit code 21) after M rate limited pages within the N seconds, configured via `--rateLimitTimeout`.
+
+This can allow a job runner or other system that starts and monitors the crawler container to implement an exponential backoff system if the crawler repeatedly exits due to being rate limited.
+The Browsertrix application provides this through built-in exponential backoff available in Kubernetes, where a crawler
+will restart only once every 5 minutes if continuously rate limited.
```
tw4l: Not the biggest detail, but is the 5 minutes bit here accurate given that exponential backoff works by increasing the amount of time between restarts each time?

tw4l (approved):
One last docs comment, otherwise looks good!

docs/docs/user-guide/rate-limits.md
ikreymer: Updated!

docs/docs/user-guide/rate-limits.md
```diff
@@ -0,0 +1,62 @@
...
+The crawler includes several options to help with this. 
+
+When pages are flagged as rate limited, the pages are still loaded in the browser, but the content is not archived.
+Instead, the pages may be queued to be retried at a later time or skipped altogether.
+
+## Flagging Rate Limited / Auth-walled pages
+
+The simplest mechanism for detection is by checking the status code. By default, the crawler now skips all pages that have a 403, 429, or 503 status code. This list can be customized with the `--rateLimitStatusCodes` setting, which accepts a list of status codes to always skip and retry.
+
+The crawler also includes a custom option `--rateLimitOnMatch`, which can take additional values in the form of `<regex>` or `<regex>:<status>`, e.g. `--rateLimitOnMatch <regex-1> --rateLimitOnMatch <regex-2>:<status>`. This setting allows for custom detection of CAPTCHAs and rate-limited pages based on page content, and optionally status code, so even 200 pages can be flagged as rate limited.
```
mistydemeo: Can we clarify what "page content" means in this context? I think it's helpful to be a bit more specific. I know what it means from reading the source, but the docs make it a little less clear. For example, we could specify the regular expression will be matched against the downloaded HTML/JavaScript.

mistydemeo (approved):
One comment, but otherwise looking good.

docs/docs/user-guide/rate-limits.md
ikreymer:
```suggestion
The crawler also includes a custom option `--rateLimitOnMatch`, which can take additional values in the form of `<regex>` or `<regex>:<status>`, e.g. `--rateLimitOnMatch <regex-1> --rateLimitOnMatch <regex-2>:<status>`. This setting allows for custom detection of CAPTCHAs and rate-limited pages based on page text content (HTML, JS, etc...), and optionally status code, so even 200 pages can be flagged as rate limited.
```

How about this?

mistydemeo: Looks good!

PR #1082, opened 2026-07-07
docs: mention YouTube

mistydemeo:
I forgot this in the previous PR.

### DIFF
--- a/docs/docs/user-guide/behaviors.md
+++ b/docs/docs/user-guide/behaviors.md
@@ -81,7 +81,7 @@ The Facebook behavior is able to automate the following features:
 
 The YouTube behavior is able to automate the following features:
 
-- Selecting a video format that the crawler will be able to capture
+- Selecting a video format that the crawler will be able to capture in a single stream. Currently, this will be a .mp4 with resolution of <= 320p due to YouTube limitations.
 - Ensuring that video playback has started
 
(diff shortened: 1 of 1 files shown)


docs/docs/user-guide/behaviors.md
```diff
@@ -76,6 +77,13 @@ The Facebook behavior is able to automate the following features:
 - Viewing full-size images, audio and video
 - Viewing quoted posts
 
+### YouTube
+
+The YouTube behavior is able to automate the following features:
+
+- Selecting a video format that the crawler will be able to capture
```
ikreymer: should we say, currently single mp4 with resolution <= 320p ?

ikreymer:
```suggestion
- Selecting a video format that the crawler will be able to capture in a single stream. Currently, this will be a .mp4 with resolution of <= 320p due to YouTube limitations.
```

ikreymer: Updated!

ikreymer (approved):

docs/docs/user-guide/behaviors.md
mistydemeo: It's currently 360p - I'll update that detail in a followup.

PR #1085, opened 2026-07-08
optimizations and fixes for pending list:

### LINKED ISSUE #1061: Possible for pending list to not be cleared under some circumstances.
This can result in a very large pending list, and exceeding large log files (see: webrecorder/browsertrix#3423).

The pending list should reflect what is actually pending (eg. no more than number of workers).

ikreymer:
- call loadDirectPage() immediately instead of queuing, as async queue may grow large with unfinished pages (will revisit at a later time)
- logging: print only first 10 entries in pending list, and total size of pending list, in case the list does grow large
- state: whenever removing a url from pending list, also remove the `<crawlid>:p:<url>` lock, since the page is no longer pending, the lock key should be removed.
- fixes #1061

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -2282,8 +2282,9 @@ self.__bx_behaviors.selectMainBehavior();
       return;
     }
 
-    const pendingPages = await this.crawlState.getPendingList();
-    const pending = pendingPages.length;
+    const pendingPagesFull = await this.crawlState.getPendingList();
+    const pending = pendingPagesFull.length;
+    const pendingPages = pendingPagesFull.slice(0, 10);
     const crawled = await this.crawlState.numDone();
     const failed = await this.crawlState.numFailed();
     const total = await this.crawlState.numFound();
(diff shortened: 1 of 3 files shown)


tw4l (approved):

src/crawler.ts
```diff
@@ -2282,8 +2282,9 @@ self.__bx_behaviors.selectMainBehavior();
       return;
     }
 
-    const pendingPages = await this.crawlState.getPendingList();
-    const pending = pendingPages.length;
+    const pendingPagesFull = await this.crawlState.getPendingList();
+    const pending = pendingPagesFull.length;
+    const pendingPages = pendingPagesFull.slice(0, 10);
```
tw4l, line 2287: My only question is I wonder if we should indicate somehow that this may not be the complete list, otherwise it might be unexpected

ikreymer: Hm, well `pending` is still the total number, I suppose could add something to the logs, this is mostly just for loggs, but maybe can revisit if its an issue

PR #1088, opened 2026-07-09
normalize seed scope to always 'wwwN.' + add docs on URL Normalization

ikreymer:
follow up to #1071 
- always ignore 'wwwN.' part of host on prefix and host scope types by updating the include regex for these scopes to include `(www[\d]*)?`
- tests: update tests to account for 'include' scope types ignoring wwwN
- no longer need to add a new seed if difference is only in 'wwwN.' since now included in original 'include' scope
- simplify --addRedirectedSeed to be a simple boolean, either adding redirected seeds or not
- docs: add docs about URL Normalization and updated addRedirectedSeed docs


ikreymer:
Not sure if this should depend on `addRedirectedSeed` value at all, or always be applied, since its probably the desirable behavior, and can be modified with an explicit exclusion if a www-normalized URL should *not* be included.

mistydemeo (approved):
Looks good. Do we have somewhere in the docs to update to indicate this change?

tw4l:
> Looks good. Do we have somewhere in the docs to update to indicate this change?

I assumed https://crawler.docs.browsertrix.com/user-guide/crawl-scope/#seed-redirects would need an update, but I think it is actually still accurate with this change so long as we continue to not ignore `wwwN.` if the `--addRedirectedSeed never` is used.

If we modify this to always ignore `wwwN.` even with `--addRedirectedSeed never`, then that section of the docs will need a change.

I'm actually leaning toward agreeing with Ilya that maybe this shouldn't depend on `--addRedirectedSeed` at all, since not wanting to follow `wwwN.` redirects seems like a real edge case and we do have another mechanism (exclusions) for that if needed.

ikreymer:
Yeah, I think we should document this change, perhaps in a different section, eg. URL Normalization, since it applies to not just seeds / all links, and not tied to redirects at all

mistydemeo (approved):

mistydemeo (approved):

ikreymer:
> If we modify this to always ignore wwwN. even with --addRedirectedSeed never, then that section of the docs will need a change.

Hm, I guess with this change `--addRedirectedSeed never` doesn't really apply, since www. / https URLs are automatically in scope and no new seed is needed..

Maybe we just have a `--addRedirectedSeed` flag that is equivalent to current `always` other current `strict`.
If someone wants to exclude a www. or http:// version, that could be added as an exclusion, but that seems to be an edge case.

mistydemeo (approved):
Looks good. I think this will be easier to understand, too. We didn't ship the intermediate behaviour, right? Didn't make a stable release?

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
+are ignored. For example, given a seed with `prefix` scope of `http://example.com/path/`, URLs starting with `https://example.com/path/` and `https://www.example.com/path/` will also be included.
 
-The other available options are:
+- For all pages, the query arguments are sorted, so that URL `https://example.com/query?A=1&B=2` and `https://example.com/query?B=2&A=1` are considered the same URL and crawled only once.
 
-- `always` - always add the redirect URL as a new seed to the crawl scope
+## Seed Page Redirects
 
-- `never` - never add the redirect URL as a new seed to the crawl scope
+Given the URL normalization rules above, if a seed pages redirect to another page, eg. if `https://example.com/` redirects to `http://example.com/` or `http://www.example.com/`, they are treated as the same seed, and no new seed is added.
```
tw4l:
```suggestion
Given the URL normalization rules above, if a seed page redirects to another page that differs only by scheme or www prefix, e.g. if `https://example.com/` redirects to `http://example.com/` or `http://www.example.com/`, they are treated as the same seed, and no new seed is added.
```

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
 
-If the URL a seed redirects to is not added as a new seed due to the option chosen, only the immediate page that is redirected to will be included in the crawl.
+If the seed page redirects to another URL, eg. `http://example.com/` to `https://example.org`, the new page is *not*
+added as a new seed by default, but that immediate page is crawled (unless the page was explicitly excluded)
```
tw4l:
```suggestion
If the seed page redirects to another URL, e.g. `http://example.com/` to `https://example.org`, the new page is *not*
added as a new seed by default, but that immediate page is crawled (unless the page was explicitly excluded).
```

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
+
+The `--addRedirectedSeeds` can be used to always add the new page URL as a seed, with the same scope as the original seed.
```
tw4l:
```suggestion
The `--addRedirectedSeeds` flag can be used to always add the new redirected page URL as a seed, with the same scope as the original seed.
```

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
+
+Note that this can lead to unintended consequences, for example if the seed is `https://example.com/path/` and it redirects to `https://example.org/` with `scopeType: prefix`, the new seed added is `https://example.org/` with `scopeType: prefix` as well, but potentially expanding the size of the crawl.
```
tw4l:
```suggestion
Note that this can lead to unintended consequences. For example, if the seed is `https://example.com/path/` and it redirects to `https://example.org/` with `scopeType: prefix`, the new seed added is `https://example.org/` with `scopeType: prefix` as well, potentially expanding the size of the crawl.
```

ikreymer:
> Looks good. I think this will be easier to understand, too. We didn't ship the intermediate behaviour, right? Didn't make a stable release?

Nope, only beta releases.

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
+
+
+!!! note "Previous behavior"
+
+    Prior to 1.14.0 release, crawler would always add all redirected seeds as new seeds. The equivalent behavior can now be enabled with the `--addRedirectedSeeds` flag.
```
tw4l:
```suggestion
    Prior to the 1.14.0 release, the crawler would always add all redirected seeds as new seeds. The equivalent behavior can now be enabled with the `--addRedirectedSeeds` flag.
```

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
+- For all pages, the query arguments are sorted, so that URL `https://example.com/query?A=1&B=2` and `https://example.com/query?B=2&A=1` are considered the same URL and crawled only once.
 
-- `always` - always add the redirect URL as a new seed to the crawl scope
+## Seed Page Redirects
 
-- `never` - never add the redirect URL as a new seed to the crawl scope
+Given the URL normalization rules above, if a seed pages redirect to another page, eg. if `https://example.com/` redirects to `http://example.com/` or `http://www.example.com/`, they are treated as the same seed, and no new seed is added.
 
-If the URL a seed redirects to is not added as a new seed due to the option chosen, only the immediate page that is redirected to will be included in the crawl.
+If the seed page redirects to another URL, eg. `http://example.com/` to `https://example.org`, the new page is *not*
```
ikreymer:
```suggestion
If the seed page redirects to another URL, eg. `http://example.com/` to `https://example.org/`, the new page is *not*
```

docs/docs/user-guide/crawl-scope.md
ikreymer:
```suggestion
added as a new seed by default, but that immediate page is crawled (unless the page was explicitly excluded).
```

tw4l (approved):
This is a good change, and I like the simpler `--addRedirectedSeeds` option it brings. I did a copyediting pass on the docs and left suggestions, also fixed a typo in the argument help text and regenerated the CLI options docs. Otherwise looks good and working as expected in testing.

docs/docs/user-guide/crawl-scope.md
ikreymer:
```suggestion
Note that this can lead to potentially broader or narrower scope, for example if the seed is `https://example.com/path/` and it redirects to `https://example.org/` with `scopeType: prefix`, the new seed added is `https://example.org/` with `scopeType: prefix` as well, but potentially expanding the size of the crawl.
```

docs/docs/user-guide/crawl-scope.md
```diff
@@ -24,7 +24,7 @@ To make this configuration as simple as possible, there are several predefined s
 
 - `custom` — crawl based on the `--include` regular expression rules.
 
-The scope settings for multi-page crawls (page-spa, prefix, host, domain) also include http/https versions, eg. given a prefix of `http://example.com/path/`, `https://example.com/path/` is also included.
+Note that the scope settings for multi-page crawls (page-spa, prefix, host, domain) are also affected by [URL Normalization](#url-normalization)
```
tw4l:
```suggestion
Note that the scope settings for multi-page crawls (page-spa, prefix, host, domain) are also affected by [URL Normalization](#url-normalization).
```

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
+- For all pages, the query arguments are sorted, so that URL `https://example.com/query?A=1&B=2` and `https://example.com/query?B=2&A=1` are considered the same URL and crawled only once.
 
-- `always` - always add the redirect URL as a new seed to the crawl scope
+## Seed Page Redirects
 
-- `never` - never add the redirect URL as a new seed to the crawl scope
+Given the URL normalization rules above, if a seed pages redirect to another page, eg. if `https://example.com/` redirects to `http://example.com/` or `http://www.example.com/`, they are treated as the same seed, and no new seed is added.
 
-If the URL a seed redirects to is not added as a new seed due to the option chosen, only the immediate page that is redirected to will be included in the crawl.
+If the seed page redirects to another URL, eg. `http://example.com/` to `https://example.org/`, the new page is *not*
```
tw4l:
```suggestion
If the seed page redirects to another URL, e.g. `http://example.com/` to `https://example.org/`, the new page is *not*
```

docs/docs/user-guide/crawl-scope.md
```diff
@@ -46,19 +46,31 @@ For example, this is most useful when crawling with a `host` or `prefix` scope,
...
+Given the URL normalization rules above, if a seed page redirects to another page that differs only by scheme or www prefix, e.g. if `https://example.com/` redirects to `http://example.com/` or `http://www.example.com/`, they are treated as the same seed, and no new seed is added.
 
-If the URL a seed redirects to is not added as a new seed due to the option chosen, only the immediate page that is redirected to will be included in the crawl.
+If the seed page redirects to another URL, eg. `http://example.com/` to `https://example.org/`, the new page is *not*
+added as a new seed by default, but that immediate page is crawled (unless the page was explicitly excluded).
+
+The `--addRedirectedSeeds` flag can be used to always add the new redirected page URL as a seed, with the same scope as the original seed.
+
+Note that this can lead to potentially broader or narrower scope, for example if the seed is `https://example.com/path/` and it redirects to `https://example.org/` with `scopeType: prefix`, the new seed added is `https://example.org/` with `scopeType: prefix` as well, but potentially expanding the size of the crawl.
```
ikreymer:
```suggestion
Note that this can lead to potentially broader or narrower scope. For example, if the seed is `https://example.com/path/` and it redirects to `https://example.org/` with `scopeType: prefix`, the new seed added is `https://example.org/` with `scopeType: prefix` as well, but potentially expanding the size of the crawl.
```

PR #1089, opened 2026-07-13
recorder: fix call to undefined variable

mistydemeo:
624128bc8c579806df64f1f28c91a82ecc2aa0bd moved the `pending` variable into the `while` loop, so it's no longer defined at the places we call these two log statements.


tw4l:
This should be addressed in https://github.com/webrecorder/browsertrix-crawler/pull/1090 (which also includes another fix), but thank you for the PR!

PR #1090, opened 2026-07-13
typo fix: don't reuse same alias for rateLimitRetries and it overrides maxPageRetries

ikreymer:
- don't reuse same alias for rateLimitRetries as it overrides regular retries
- use different variable for 'pending' as was not being detected as a out-of-scope due to type conflict

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -788,7 +788,6 @@ class ArgParser {
         },
 
         rateLimitMaxRetries: {
-          alias: "retries",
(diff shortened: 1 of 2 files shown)


tw4l (approved):
Thanks, good catches!

mistydemeo (commented):
Since Emma's got a fix for `pending`, maybe we can back that part out and handle it that way? Removing the duplicate `retries` flag definitely makes sense to me.

PR #1091, opened 2026-07-13
eslint: catch undefined variables

mistydemeo:
This will catch the variable that we noticed in #1089/#1090.


tw4l (approved):
Thank you!

mistydemeo:
Emma's got a fix that should catch this without eslint by giving typescript better awareness of the relevant global variables, so I'll close this for now.

PR #1092, opened 2026-07-13
typing fixes: don't include 'jest' in app globals, include only in tests

ikreymer:
recorder: remove undefined variable included due to incorrect typing

follow-up from #1090


mistydemeo (approved):
Confirmed working!

emma-sg (approved):

PR #1094, opened 2026-07-14
Fix --alwaysAddBehaviorLinks / smart scoping to work as expected:

ikreymer:
- ensure regular links always obey scope, as using same method
- support addLink(url, alwaysObeyScope) from added in, webrecorder/browsertrix-behaviors#150
- scope only ignored if alwaysObeyScope is false and alwaysAddBehaviorLinks is set
- move addLink() declaration to setupPage(), as extractLinks() not called for single page scope
- simplify, no longer need extra 'callbacks' in PageState

### DIFF
--- a/src/util/worker.ts
+++ b/src/util/worker.ts
@@ -19,8 +19,6 @@ export type WorkerState = {
   page: Page;
   cdp: CDPSession;
   workerid: WorkerId;
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
-  callbacks: Record<string, Function>;
   recorder: Recorder | null;
   markPageUsed: () => void;
   frameIdToExecId: Map<string, number>;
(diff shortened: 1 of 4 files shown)


src/util/worker.ts
```diff
@@ -171,13 +166,11 @@ export class PageWorker {
 
         this.page = page;
         this.cdp = cdp;
-        this.callbacks = {};
```
mistydemeo, line 174: Huh, was worker's version of this unused?

ikreymer: yeah, it was only used to set the addLink(), I think initially it was implemented in a different way, no longer needed now since we can create in in setupPage()

mistydemeo (approved):

mistydemeo:
Well, the tests are fixed as mysteriously as they failed. We may have had a flake in an unrelated test.

PR #1096, opened 2026-07-14
tests: fix custom_selector test

mistydemeo:
The IANA webpage has changed since the test was written. We should migrate this test to use a different site when we can, but in the meantime this should get it passing again.


ikreymer:
Superseded by #1093

PR #1098, opened 2026-07-15
Fix incorrect handling of redis extra seeds + extra logging for for excluded pages

### LINKED ISSUE #1097: Bug in handling of redirect seeds
Currently, seeds that are added as part of a redirect are kept in the Redis and referred to by their index.
Due to an unfortunate bug, this index is made made invalid when multiple extra seeds are added, as they are added at the beginning of Redis list using `lpush` instead of `rpush`. This affects crawls with multiple seeds where more than one seed has a redirect

This will be slightly mitigated by #1088 which expands seed scope to normalize wwwN and reduces need for adding extra redirect seeds, however the core issue should be fixed as well.

ikreymer:
- Fixes #1097 
- Using rpush instead of lpush to avoid seed index being off
- other cleanup: use lindex instead of lrange to get one entry
- Improved logging of excluded pages: consolidate markExcluded() in crawler which always logs to skippedPages report (if in enabled)
Add new SkippedReasons:
- outOfScopeMidCrawl if a page became out-of-scope for some reason while crawl is running (eg. crawl was restarted with scope changed)
- excludedMidCrawl if a page was excluded due to user-added exclusions mid crawl

### DIFF
--- a/src/util/constants.ts
+++ b/src/util/constants.ts
@@ -127,7 +127,7 @@ export const WARC_REFERS_TO_CONTAINER = "WARC-Refers-To-Container";
 export enum SkippedReason {
   OutOfScope = "outOfScope",
   OutOfScopeMidCrawl = "outOfScopeMidCrawl",
-  ExcludedMidCrawl = "excludedMidCrawl",
+  ExcludedMidCrawl = "outOfScopeUserExclusion",
   PageLimit = "pageLimit",
   RobotsTxt = "robotsTxt",
   RedirectToExcluded = "redirectToExcluded",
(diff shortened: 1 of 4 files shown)


src/util/constants.ts
```diff
@@ -126,6 +126,8 @@ export const WARC_REFERS_TO_CONTAINER = "WARC-Refers-To-Container";
 
 export enum SkippedReason {
   OutOfScope = "outOfScope",
+  OutOfScopeMidCrawl = "outOfScopeMidCrawl",
+  ExcludedMidCrawl = "excludedMidCrawl",
```
tw4l:
```suggestion
  ExcludedMidCrawl = "outOfScopeUserExclusion",
```
Or something similar, so that all of the OOS skipped reasons can be easily `grep`ed together

tw4l (approved):
Other than renaming the one constant, this looks great

PR #1099, opened 2026-07-15
tests: add tests for alwaysAddBehaviorLinks

mistydemeo:
These are specifically meant to follow up on Ilya's recent changes to this flag.

~~Since these perform an actual crawl they're *very slow*. I based them off the existing extra hops depth tests, but I don't know if we want to be doing something to try to speed these up.~~ Turns out those tests were full crawls by design, which we don't need here. Added `--scopeType page` and sped it up substantially.

### DIFF
--- a/tests/alwaysAddBehaviorLinks.test.ts
+++ b/tests/alwaysAddBehaviorLinks.test.ts
@@ -1,0 +1,156 @@
+import fs from "fs";
+
+import util from "util";
+import { exec as execCallback } from "child_process";
+
+const exec = util.promisify(execCallback);
+
+const behaviorLinksTimeout = 180000;
+
+test(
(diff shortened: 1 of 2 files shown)


tests/custom-behaviors/addlink.js
```diff
@@ -0,0 +1,21 @@
...
+    return "AddLinkBehavior";
+  }
+
+  static isMatch() {
+    return window.location.origin === "https://old.webrecorder.net";
+  }
+
+  async *run(ctx) {
+    ctx.log("Adding link to domain outside the crawl scope!");
+    await ctx.Lib.addLink("https://example.com");
```
ikreymer:
```suggestion
    await ctx.Lib.addLink("https://example-com.webrecorder.net/");
```

We have our own domain for this since example.com started blocking.

ikreymer (approved):

PR #1103, opened 2026-07-20
pending list optimizations: drop unstarted requests to avoid longer waits

ikreymer:
Follow-up to #1085 Further optimizations for awaiting page resources at end of page load:
- When a page is done, we can actually more precisely detect if requests that have been started (request sent), but not yet received and avoid waiting for those, by checking if either asyncLoading or intercepting is set. If not, the request is not being received yet, and can interrupt it immediately.
- Should also prevent new requests from starting by adding a `finishingPage` flag that stops additional requests from being saved.

Testing:
- Test on https://forum.webrecorder.net/
- Without this fix, observe there is a long polling XHR request (sometimes multiple) that holds up the page finish
- With this fix, this request is dropped quickly.
- Can also test with a large youtube video on http://oembed.link/youtube.com/... and ensure video is awaited properly.


src/util/recorder.ts
```diff
@@ -1126,9 +1128,12 @@ export class Recorder extends EventEmitter {
   }
 
   async awaitPageResources() {
+    // Prevent new requests from being added
+    this.finishingPage = true;
+
     // Check stale requests left in the pendingRequests list
     for (const [requestId, reqresp] of this.pendingRequests.entries()) {
-      // check if possibly being fetched, continue
+      // if being fetched, don't removed
```
mistydemeo:
```suggestion
      // if being fetched, don't remove
```

tw4l:
```suggestion
      // if being fetched, don't remove
```

src/util/recorder.ts
```diff
@@ -1138,8 +1143,9 @@ export class Recorder extends EventEmitter {
       if (reqresp.payload?.length) {
         this.removeReqResp(requestId);
         await this.serializeToWARC(reqresp);
-        // if no URL set, likely this has been abandoned, drop
```
mistydemeo, line 1141: We don't need specific "no URL" handling now I presume?

mistydemeo (approved):

src/util/recorder.ts
ikreymer: right, this is now part of the broader else condition.

src/util/recorder.ts
```diff
@@ -1126,9 +1128,12 @@ export class Recorder extends EventEmitter {
   }
 
   async awaitPageResources() {
+    // Prevent new requests from being added
+    this.finishingPage = true;
+
     // Check stale requests left in the pendingRequests list
     for (const [requestId, reqresp] of this.pendingRequests.entries()) {
-      // check if possibly being fetched, continue
+      // if being fetched, don't remove
```
mistydemeo: Inexplicably, the GitHub-generated patch added a `CRLF` here? Huh? You'll have to `yarn format --write` to fix it.

mistydemeo (approved):

PR #1104, opened 2026-07-20
various edge-case fixes/improvements:

ikreymer:
ensure crawl never accidentally ends up in done state:
- don't default crawl to 'done' after crawl(), default to 'interrupted' if no condition is met
- ensure isFinished() check also takes into account pending list!
- use type-safe getFinalCrawlState() function to get state
- make note of state precedence checks in comment

misc fixes:
- log number of seeds instead of seeds in case seed list is large
- log excluded count in stats
- always delete WACZ after uploading successfully, to avoid double uploads

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -561,7 +561,10 @@ export class Crawler {
     this.seeds = await parseSeeds(this.downloadsDir, this.params);
     this.numOriginalSeeds = this.seeds.length;
 
-    logger.info("Seeds", { seedCount: this.numOriginalSeeds });
+    logger.info("Seeds", {
+      firstSeed: this.seeds[0],
+      seedCount: this.numOriginalSeeds,
+    });
 
     logger.info("Link Selectors", this.params.selectLinks);
 
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -561,7 +561,7 @@ export class Crawler {
     this.seeds = await parseSeeds(this.downloadsDir, this.params);
     this.numOriginalSeeds = this.seeds.length;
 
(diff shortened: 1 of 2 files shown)


src/util/state.ts
```diff
@@ -1206,7 +1206,11 @@ return inx;
   }
 
   async isFinished() {
-    return (await this.queueSize()) == 0 && (await this.numDone()) > 0;
+    return (
+      (await this.queueSize()) == 0 &&
+      (await this.numDone()) > 0 &&
+      (await this.numPending()) === 0
```
tw4l, line 1212: Good catch!

src/crawler.ts
```diff
@@ -561,7 +561,7 @@ export class Crawler {
     this.seeds = await parseSeeds(this.downloadsDir, this.params);
     this.numOriginalSeeds = this.seeds.length;
 
-    logger.info("Seeds", this.seeds);
+    logger.info("Num Original Seeds", { numSeeds: this.numOriginalSeeds });
```
tw4l:
```suggestion
    logger.info("Seeds", { seedCount: this.numOriginalSeeds });
```
Probably should use full words here, so keep as "Seeds" or "Number of Seeds", "Seed Count," or something like that.

Could also log the first seed, or first few? I tend to find it helpful when looking at crawl logs later, but understand we don't want to log dozens or hundreds here.

mistydemeo (approved):

tw4l (approved):
One nit/suggestion, otherwise LGTM

src/crawler.ts
ikreymer: Ah good idea, doesn't hurt to include first seed, since often that is what we care about and want to see quickly..

PR #1112, opened 2026-07-21
treat connection timeouts as rate limits:

ikreymer:
Connection timeouts (both for direct fetch and browser page load) should be considered as rate limits and also increment the corresponding counters:
- Use STATUS_CONNECTION_ERROR as 999
- Add STATUS_DNS_ERROR 899 as DNS failure (ENOTFOUND error)
- Increment rate limit counts, but treat both statuses as failures (use fail page retry max for these failures)
- For DNS failure in direct fetch, don't attempt to load page in the browser, as will also certainly failure.


tw4l (approved):
Nice addition and renaming, thanks!

tw4l (approved):
Tested, working well

PR #1115, opened 2026-07-22
quickfix: follow up to regression caused by #1109:

ikreymer:
- if not 'final exit', (e.g. paused), the deletion code was not being run before exit
- now check if doDelete is set and call deletion and clear data

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -2168,6 +2168,7 @@ self.__bx_behaviors.selectMainBehavior();
         }
 
         if (this.storage && this.uploadAndDeleteLocal) {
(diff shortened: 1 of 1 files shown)


src/crawler.ts
```diff
@@ -2175,6 +2175,9 @@ self.__bx_behaviors.selectMainBehavior();
 
     // from here, actions that should happen on final crawler exit (not temp interrupt)
     if (!this.finalExit) {
```
mistydemeo, line 2178: Can we add an extra comment on why we now have two `if (doDeleteDir)` calls, just so we remember this context?

mistydemeo (approved):

PR #1117, opened 2026-07-23
deps: update to warcio.js 2.4.11 to fix serialization issue with WARC-Json-Metadata header

ikreymer:
It's possible for WARC-JSON-Metadata to contain ',,,' which was used as a special marker for multi-value headers.
warcio 2.4.11 always treats WARC-JSON-Metadata as a single value.
logging: log url when logging failed to serialize messages


src/util/recorder.ts
```diff
@@ -428,32 +428,38 @@ export class Recorder extends EventEmitter {
...
-    if (reqresp.url === this.finalPageUrl) {
+    if (url === this.finalPageUrl) {
       this.finalPageUrl = reqresp.getRedirectUrl();
     }
 
     this.serializeToWARC(reqresp).catch((e) =>
-      logger.warn("Error Serializing to WARC", e, "recorder"),
+      logger.warn(
+        "Error Serializing to WARC",
+        { url, ...formatErr(e) },
```
mistydemeo, line 460: Good call.

mistydemeo (approved):

PR #1119, opened 2026-07-23
Expand login field detection

wvengen:
Encountered a login form where no `name` was present on the inputs, but `id` was (login probably handled by Javascript).
There probably was a typo in the password selector as well.

### DIFF
--- a/src/create-login-profile.ts
+++ b/src/create-login-profile.ts
@@ -286,10 +286,10 @@ async function automatedProfile(
 
   try {
     u = await page.waitForSelector(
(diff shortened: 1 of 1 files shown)


ikreymer (approved):
Thanks, yes, that was a typo and seems reasonable to add.
We actually haven't been using this feature much, as we mostly use interactive profile creation,

wvengen:
Thank you for merging!
Interactive profile creation works great - except when the session time is limited (and not only by cookie lifetime). Though rare, it does happen on some websites.

PR #1120, opened 2026-07-23
postLoadDelay for automated create-login-profile

wvengen:
Support `postLoadDelay` for `create-login-profile`.
I found that on some websites, automated login did not store the logged in data in the browser profile. Adding a small delay, allowed the login process to fully finish.


ikreymer (approved):
Makes sense, thanks!

PR #1122, opened 2026-07-23
Fixes off-by-1 error in WACZ compressed cdx index creation.

### LINKED ISSUE #1121: Off-by-one error in compressed CDXJ creation
It turns out we've had an off-by-1 error in the zipped CDX block creation :(
As a result, the line in the index is actually the second line in the block, and the last line is added to the next block.
Instead of 256 lines per block, we were only getting 255 for every block except the first one..
This can affect replay when accessing what is the last line in the block (1/256 lines) if no additional index block are loaded.
Luckily, the bug is hit very rarely, but does affect replay in subtle ways (will have a separate replay fix to account for this)

Adding a PR with tests to ensure this is working as expected

ikreymer:
Fixes off-by-1 error where the index.idx actually references the second line in the index.cdx.gz block, and the last line was getting added to the next block.

Adding tests to ensure the compressed CDXJ is working as expected now.

Fixes #1121

### DIFF
--- a/src/util/wacz.ts
+++ b/src/util/wacz.ts
@@ -326,7 +326,7 @@ export async function mergeCDXJ(
   indexesDir: string,
   zipped: boolean | null = null,
 ) {
-  // most used for testing
+  // added as env vars as really just used for testing
   const zipLinesPerBlock =
     parseInt(process.env.ZIP_LINES_PER_BLOCK || "") || LINES_PER_BLOCK;
   const zipCdxMinSize =
--- a/src/util/wacz.ts
+++ b/src/util/wacz.ts
@@ -326,6 +326,12 @@ export async function mergeCDXJ(
   indexesDir: string,
   zipped: boolean | null = null,
 ) {
+  // most used for testing
+  const zipLinesPerBlock =
+    parseInt(process.env.ZIP_LINES_PER_BLOCK || "") || LINES_PER_BLOCK;
+  const zipCdxMinSize =
+    parseInt(process.env.ZIP_CDX_MIN_SIZE || "") || ZIP_CDX_MIN_SIZE;
+
   async function* readLinesFrom(stdout: Readable): AsyncGenerator<string> {
     for await (const line of readline.createInterface({ input: stdout })) {
       yield line + "\n";
@@ -335,6 +341,7 @@ export async function mergeCDXJ(
   async function* generateCompressed(
     reader: AsyncGenerator<string>,
     idxFile: Writable,
+    zipLinesPerBlock: number,
   ) {
     let offset = 0;
 
@@ -345,7 +352,6 @@ export async function mergeCDXJ(
     let cdxLines: string[] = [];
 
     let key = "";
-    let count = 0;
 
     idxFile.write(
       `!meta 0 ${JSON.stringify({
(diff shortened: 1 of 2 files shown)


src/util/wacz.ts
```diff
@@ -326,6 +326,12 @@ export async function mergeCDXJ(
   indexesDir: string,
   zipped: boolean | null = null,
 ) {
+  // most used for testing
```
ikreymer:
```suggestion
  // added as env vars as really just used for testing
```

tests/zipped-cdx-index.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    allCdxLines = [...allCdxLines, ...cdxLines];
+
+    // add to line counts
+    blockLineCounts.push(cdxLines.length);
+  }
+
+  // each line count is equal to number of lines per block, except last one
+  for (let i = 0; i < blockLineCounts.length - 1; i++) {
+    expect(blockLineCounts[0]).toBe(LINES_PER_BLOCK);
```
tw4l:
```suggestion
    expect(blockLineCounts[i]).toBe(LINES_PER_BLOCK);
```

Should iterate through list, not just keep checking first one, yeah?

tests/zipped-cdx-index.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const LINES_PER_BLOCK = 16;
+
+test("ensure basic crawl run with docker run passes", async () => {
+  execSync(
+    `docker run -e ZIP_LINES_PER_BLOCK=${LINES_PER_BLOCK} -e ZIP_CDX_MIN_SIZE=10000 -v $PWD/test-crawls:/crawls webrecorder/browsertrix-crawler crawl --url https://old.webrecorder.net/ --limit 6 --collection zipped-cdx-merge --generateWACZ`,
+  );
+});
+
+test("check that indexes/index.cdx.gz and indexes/index.idx exist", () => {
```
mistydemeo, line 13: Is it safe to run these as tests that come after the previous one if it's setting up their state? Could they be run in other orders? Maybe we should make this all one test.

mistydemeo (approved):

tw4l (approved):

tests/zipped-cdx-index.test.ts
ikreymer: Yes, this is standard for how these test suites are set up, they run in order, we do this other places also, grouped by file.

tests/zipped-cdx-index.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const blockLineCounts = [];
+
+  let allCdxLines: string[] = [];
+
+  for (const line of indexBuff.trim().split("\n")) {
+    if (line.startsWith("!meta 0")) {
+      continue;
+    }
+    // parse index line in format surt ts json, eg. com,example)/ 20261226010203 {...}
```
ikreymer:
```suggestion
    // parse index line in format <surt> <ts> <json>, e.g. 
    // com,example)/ 20261226010203 {...}
```

ikreymer:
Did an additional sanity check to ensure this test was failing before this fix, and getting only 15 lines per block :/

PR #1123, opened 2026-07-23
set rateLimitMaxRetries to 4 instead of -1 to avoid crawls getting st…

ikreymer:
…uck retrying indefinitely by default!

update docs as well


tw4l (approved):
Makes sense! Thanks for the docs updates

PR #1126, opened 2026-07-28
ensure direct-fetch connection are always closed.

### LINKED ISSUE #1125: [Bug]: Possible crash from undici if direct fetch connections are not aborted.
Found a reproducible error/assertion failure from undici.

```
docker run -v $PWD/crawls:/crawls -p 9222:9222 -p 9223:9223 -it webrecorder/browsertrix-crawler crawl --url "https://www.autorenlexikon.lu/page/author/500/5003/FRE/index.html" --scopeType page --logging debug,stats
...
    throw err;
    ^

AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(!this.paused)

    at Parser.finish (/app/node_modules/undici/lib/dispatcher/client-h1.js:367:5)
    at TLSSocket.onHttpSocketEnd (/app/node_modules/undici/lib/dispatcher/client-h1.js:951:30)
    at TLSSocket.emit (node:events:521:24)
    at endReadableNT (node:internal/streams/readable:1735:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '==',
  diff: 'simple'
}
```

Upon investigation, it turns out that it can be caused by a connection that is not aborted and not continued.

(It's also fixed in undici 8, but that requires other changes to upgrade).

ikreymer:
ensure connection that is not continued, (eg. due to HTML response) is always aborted and closed, and more generally,
doCancel() is always called:
- add finally and doCancel() in directFetch()
- clear body and stream on successful fetch in loadBody()
- also add doCancel() to loadBody() exception handler
- fixes #1125

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -1563,31 +1563,39 @@ export class Recorder extends EventEmitter {
       cdp,
     });
 
-    if (!(await fetcher.loadHeaders())) {
-      return fetcher.lastError === "ENOTFOUND"
-        ? STATUS_DNS_ERROR
-        : STATUS_CONNECTION_ERROR;
-    }
+    try {
+      if (!(await fetcher.loadHeaders())) {
+        return fetcher.lastError === "ENOTFOUND"
+          ? STATUS_DNS_ERROR
+          : STATUS_CONNECTION_ERROR;
+      }
 
-    const mime = reqresp.getMimeType() || "";
-    // cancel if not 200 or mime is html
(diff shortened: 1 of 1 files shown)


tw4l (approved):
Now that we're awaiting `fetcher.loadDirectPage(state);` this makes sense, I think if we switch back to async queuing later on we may need to modify this implementation a bit.

PR #1131, opened 2026-07-29
qa: fix QA of small WACZ files:

### LINKED ISSUE #1127: qa: replay server crashes with ERR_OUT_OF_RANGE when the source WACZ is smaller than 65,558 bytes
## Summary

The `qa` entrypoint crashes immediately — before a single page is compared — when the
source WACZ is smaller than 65,558 bytes. The internal `ReplayServer` does not clamp
HTTP suffix ranges (`Range: bytes=-N`) to the file size, so it passes a negative `start`
offset to `fs.createReadStream()`, which throws `RangeError [ERR_OUT_OF_RANGE]`. Because
the throw happens inside an un-awaited async request handler, it surfaces as an
unhandled rejection and takes the whole process down.

Large WACZ files are unaffected (QA runs fine here on ~429 MB and ~756 MB sources).

## Environment

- Crawler image: `webrecorder/browsertrix-crawler:1.13.2` (image created 2026-06-20)
- Node inside the image: v24.17.0 — Ubuntu 24.04.4
- Docker Engine: 29.6.2, `linux/amd64` (Docker Desktop, WSL2 backend, kernel 6.18.33.2-microsoft-standard-WSL2)
- Host OS: Windows 11 (x86_64)
- Verified still present on `main` as of 2026-07-28 (latest release: v1.14.0): `src/util/replayserver.ts` still computes `opts.start = total - opts.end` without clamping.

## Steps to reproduce

This is fully self-contained: it serves a trivial page from inside the container, so it
needs no external network and no fixture files. Paste as-is.

```bash
docker run --rm --entrypoint bash webrecorder/browsertrix-crawler:1.13.2 -c '
  echo "<html><body>hi</body></html>" > /tmp/i.html
  node -e "const h=require(\"fs\").readFileSync(\"/tmp/i.html\");require(\"http\").createServer((q,s)=>{s.writeHead(200,{\"Content-Type\":\"text/html\"});s.end(h)}).listen(8080)" &
  sleep 1
  crawl --url http://localhost:8080/ --scopeType page --collection tiny \
        --generateWACZ --pageLimit 1 --behaviors "" > /dev/null 2>&1
  ls -l /crawls/collections/tiny/tiny.wacz
  qa --qaSource /crawls/collections/tiny/tiny.wacz --collection tiny-qa --generateWACZ
'
```

The crawl step produces a ~9.7 KB WACZ here. Any WACZ below 65,558 bytes will do.

If you prefer to skip the crawl entirely, this reproduces the same crash against the
shipped `ReplayServer` directly, against a 43,008-byte dummy file:

```bash
docker run --rm --entrypoint node webrecorder/browsertrix-crawler:1.13.2 -e '
  const fs = require("fs");
  fs.writeFileSync("/tmp/tiny.wacz", Buffer.alloc(43008, 1));
  import("/app/dist/util/replayserver.js").then(({ ReplayServer }) => {
    new ReplayServer("/tmp/tiny.wacz");
    setTimeout(() => fetch("http://localhost:9990/source.wacz", { headers: { range: "bytes=-65536" } }), 400);
  });
'
```

## Expected behavior

Per RFC 9110 §14.1.2, a suffix-length larger than the representation must be clamped:
"if the selected representation is shorter than the specified suffix-length, the entire
representation is used". QA should proceed normally against small WACZ files.

## Actual behavior

The process dies with exit code 1 right after `Worker starting`:

```
{"timestamp":"2026-07-24T01:48:28.120Z","logLevel":"info","context":"worker","message":"Creating 1 workers","details":{}}
{"timestamp":"2026-07-24T01:48:28.121Z","logLevel":"info","context":"worker","message":"Worker starting","details":{"workerid":0}}
node:internal/fs/streams:208
    validateInteger(this.start, 'start', 0);
    ^

RangeError [ERR_OUT_OF_RANGE]: The value of "start" is out of range. It must be >= 0 && <= 9007199254740991. Received -55856
    at new ReadStream (node:internal/fs/streams:208:5)
    at Object.createReadStream (node:fs:3190:10)
    at ReplayServer.handleRequest (file:///app/dist/util/replayserver.js:85:28)
    at Server.<anonymous> (file:///app/dist/util/replayserver.js:43:65)
    at Server.emit (node:events:509:28)
    at parserOnIncoming (node:_http_server:1226:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:125:17) {
  code: 'ERR_OUT_OF_RANGE'
}

Node.js v24.17.0
```

`-55856` is exactly `9702 - 65558` for that run's WACZ size. Other observed values line
up the same way: a 33,186-byte WACZ gives `-32372`, and a 43,008-byte dummy file gives
`-22528`.

## Analysis

**Where the 65,558 comes from.** The bundled ReplayWeb.page service worker
(`/app/html/rwp/sw.js`) reads the ZIP end-of-central-directory with

```js
const e = 65558, r = await this.loader.getRangeFromEnd(e, !1)
```

65,558 = 65,535 (max ZIP comment) + 22 (EOCD record). The HTTP loader sets
`canDoNegativeRange = true` in its constructor, and `getRangeFromEnd` then issues a real
suffix range:

```js
e < 0 ? n.set("Range", `bytes=${e}`) : n.set("Range", `bytes=${t}-${t + e - 1}`)
```

so the server receives `Range: bytes=-65558`. (The loader has a fallback that downloads
the whole file instead, but it only applies when `canLoadOnDemand` is false — and
`ReplayServer` advertises `Accept-Ranges: bytes` and answers 206, so that path is never
taken.)

**Where it breaks.** `getRespOptsForRequest` in `/app/dist/util/replayserver.js`:

```js
opts.start = parseInt(array[1]);
opts.end = parseInt(array[2]);
// negative value, subtract from end
if (isNaN(opts.start) && !isNaN(opts.end)) {
    opts.start = total - opts.end;   // <-- negative when total < 65558
    opts.end = total - 1;
}
...
return { status: 206, opts, contentRange, contentLength: opts.end - opts.start + 1 };
```

`opts.start` goes negative and is handed to `fs.createReadStream(this.origFileSource, opts)`
at line 85. Note that `contentLength` is wrong in the same case too — it computes 65,558
even though the file is smaller, so even a non-throwing stream would produce a
`Content-Length` larger than the resource.

**Why it kills the process.** `handleRequest` is `async` and the server callback does not
await or catch it (`http.createServer((req, res) => this.handleRequest(req, res))`), so
the synchronous throw becomes an unhandled rejection and terminates Node.

## Suggested fix

Clamp the suffix range and derive the length from the clamped values, e.g.:

```js
if (isNaN(opts.start) && !isNaN(opts.end)) {
    opts.start = Math.max(0, total - opts.end);
    opts.end = total - 1;
}
```

Guarding `handleRequest` with a try/catch that returns 500 (or 416 for an unsatisfiable
range) would additionally keep a malformed `Range` header from taking down the crawler.

I'm happy to open a PR if that's useful.

ikreymer:
- avoid negative index, clamp to 0
- also ignore extraPages.jsonl if it does not exist and only use pages.jsonl
- fixes #1127

### DIFF
--- a/src/replaycrawler.ts
+++ b/src/replaycrawler.ts
@@ -225,7 +225,11 @@ export class ReplayCrawler extends Crawler {
         }
       }
     } catch (e) {
-      logger.warn("No extraPages.jsonl found, ignoring", {}, "replay");
+      logger.warn(
+        "No extraPages.jsonl found or error reading extraPages, ignoring",
+        e,
+        "replay",
+      );
     }
   }
 
--- a/src/replaycrawler.ts
+++ b/src/replaycrawler.ts
@@ -213,15 +213,19 @@ export class ReplayCrawler extends Crawler {
       }
     }
 
-    const extraPagesReader = await loader.loadFile("pages/extraPages.jsonl");
+    try {
+      const extraPagesReader = await loader.loadFile("pages/extraPages.jsonl");
 
-    if (extraPagesReader) {
-      for await (const buff of extraPagesReader.iterLines()) {
-        await this.addPage(buff, count++);
-        if (this.limitHit) {
-          break;
+      if (extraPagesReader) {
+        for await (const buff of extraPagesReader.iterLines()) {
+          await this.addPage(buff, count++);
+          if (this.limitHit) {
+            break;
+          }
         }
       }
+    } catch (e) {
+      logger.warn("No extraPages.jsonl found, ignoring", {}, "replay");
     }
   }
 
--- a/src/util/replayserver.ts
+++ b/src/util/replayserver.ts
@@ -139,7 +139,7 @@ export class ReplayServer {
       opts.end = parseInt(array[2]);
       // negative value, subtract from end
       if (isNaN(opts.start) && !isNaN(opts.end)) {
-        opts.start = total - opts.end;
+        opts.start = Math.max(0, total - opts.end);
         opts.end = total - 1;
       } else if (isNaN(opts.end)) {
         opts.end = total - 1;


src/replaycrawler.ts
```diff
@@ -213,15 +213,19 @@ export class ReplayCrawler extends Crawler {
...
+      if (extraPagesReader) {
+        for await (const buff of extraPagesReader.iterLines()) {
+          await this.addPage(buff, count++);
+          if (this.limitHit) {
+            break;
+          }
         }
       }
+    } catch (e) {
+      logger.warn("No extraPages.jsonl found, ignoring", {}, "replay");
```
mistydemeo: Is there a specific exception we might want to catch here? Just so this doesn't accidentally catch unrelated exceptions?

ikreymer: Hm, well if treat extraPages.jsonl as optional, then we can ignore any errors related to reading it.
I suppose could detect if its there by some other error occurred and log more specifically, but still ignore it (rather than aborting the process)

mistydemeo: Yeah, maybe we can log the exception so we get a better sense of what happened? Just want to make sure we're not masking other errors.

ikreymer: Updated!

mistydemeo (approved):

PR #1133, opened 2026-07-29
argParser: mark CSS selector parse failure unrecoverable

mistydemeo:
Since these are provided in the config/CLI arguments, these selectors won't change between runs. If that happens, and we have the crawler set to restart on error, we could end up in a crashloop where we keep trying to run this crawl over and over again.

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -15,6 +15,7 @@ import {
   SERVICE_WORKER_OPTS,
   DEFAULT_SELECTORS,
   BEHAVIOR_TYPES,
+  ExitCodes,
   ExtractSelector,
   DEFAULT_MAX_RETRIES,
   BxFunctionBindings,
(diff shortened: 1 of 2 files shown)


src/util/argParser.ts
```diff
@@ -877,6 +878,11 @@ class ArgParser {
         try {
           parser(argv.clickSelector);
         } catch (e) {
+          // Try to avoid triggering a restart since this is unrecoverable
+          // without changing the config.
+          if (argv.restartsOnError) {
```
ikreymer, line 883: Hm, this isn't needed here, instead we can pass an exit code to `logger.fatal(...)` and probably should do that for all of the calls in argParser.ts

src/util/argParser.ts
```diff
@@ -877,6 +878,11 @@ class ArgParser {
...
+            logger.setOverrideFatalExitCode(ExitCodes.InvalidConfig);
+          }
           void logger.fatal("Invalid Autoclick CSS Selector", {
             selector: argv.clickSelector,
```
ikreymer, line 887:
```suggestion
            selector: argv.clickSelector,
            exitCode: ExitCodes.InvalidConfig
```

Instead, should do this for all of the fatal() cals in here.

ikreymer:
This makes sense since this would be more clearer exit code for before redis is inited.
As mentioned, should pass this to all of the fatal() calls, and also update exit_codes.md and regenerate the cli docs.

src/util/argParser.ts
mistydemeo: `exitCode` isn't part of the `data` argument, it's the fourth positional argument. It's part of the JSON that gets logged to the terminal.

ikreymer:
Closing in favor of #1138 which adds this more generally

PR #1138, opened 2026-07-31
more clearly mark config / invalid arg errors with distinct exit code,

ikreymer:
to be used before Redis state is inited to ensure crawler can exit with distinct exit code that indicates invalid config, to complement webrecorder/browsertrix#3552
- add ExitCodes.InvalidConfig exit code (4) and 'config' log type
- to be tested for in app via webrecorder/browsertrix#3552

### DIFF
--- a/docs/docs/user-guide/exit-codes.md
+++ b/docs/docs/user-guide/exit-codes.md
@@ -7,6 +7,7 @@ The crawler uses following exit codes to indicate crawl result.
 | 0 | Success | Crawl completed normally |
 | 1 | GenericError | Unspecified error, check logs for more details |
 | 3 | OutOfSpace | Disk is already full |
+| 4 | InvalidConfig | A CLI argument or YAML config options is invalid, fatal (non-retryable) |
 | 8 | RedisUnavailable | Could not connect to the Redis instance for managing queues or deduplication |
 | 9 | Failed | Crawl failed unexpectedly, might be worth retrying |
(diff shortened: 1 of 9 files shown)


docs/docs/user-guide/exit-codes.md
```diff
@@ -7,6 +7,7 @@ The crawler uses following exit codes to indicate crawl result.
 | 0 | Success | Crawl completed normally |
 | 1 | GenericError | Unspecified error, check logs for more details |
 | 3 | OutOfSpace | Disk is already full |
+| 4 | InvalidConfig | A CLI argument or YAML config options is invalid, fatal (non-retryable) |
```
tw4l:
```suggestion
| 4 | InvalidConfig | A CLI argument or YAML config option is invalid, fatal (non-retryable) |
```

src/util/logger.ts
```diff
@@ -61,6 +61,7 @@ export const LOG_CONTEXT_TYPES = [
   "scope",
   "robots",
   "dedupe",
+  "config",
```
tw4l, line 64: I'm not totally sure this needs a new context since it's only used with fatal messages, but no harm either

tw4l (approved):
Makes sense. Might be good to add a test or two with options that will trigger this to verify the right exit code is used.

ikreymer:
> Makes sense. Might be good to add a test or two with options that will trigger this to verify the right exit code is used.

Done, also had to updated exiting test which tested for 17 and now get a 4 for this new exit code.

PR #1140, opened 2026-08-05
argParser: disable startEarly to behaviours

### LINKED ISSUE #1142: Investigate disabling startEarly in behaviours
We've been passing `startEarly` as true when initializing the behaviours since c247189474e2786ad381a67e49447b03f2f35f0f in response to certain sites where we weren't waiting long enough for them to initialize, in particular Instagram.

Checking what `MutationObserver` actually does, I don't think it was originally *intended* to delay page load. The commit message is short on detail, but it appears that we were instead relying on the side effect that initializing a `MutationObserver` too early, before the page had finished loading, would cause the browser to hang for 5-7 seconds. That hang was papering over cases where we needed to wait for sites to finish loading; the discussion thread I'd found from when we added this indicates we did it to make sure that Instagram finished loading before we began crawling.

Now that we have `awaitPageLoad` for behaviours, we don't always have to do this for every page; we can use targeted behaviours on specific sites to ensure we wait on exactly the things we need them to, but avoid doing this on other sites where it's not gaining us anything. Instagram, for example, doesn't benefit from this wait at all anymore since we now have Instagram-specific waiting behaviours.

We'll want to ensure we do a lot of testing to see what this does for crawling fidelity. There's a good chance we've ended up accidentally relying on this to paper over issues with page content being loaded, and we'll want to ensure we're not going to see a regression in content captured from disabling this.

I should also note that setting `startEarly` to false doesn't turn off the `MutationObserver` entirely. It just changes when we initialize it. If `startEarly` is set, we initialize the `MutationObserver` instantly as soon as we construct the `AutoFetcher` class: https://github.com/webrecorder/browsertrix-behaviors/blob/d89f7c668bf2edea5bc991620f106fed1cc96cf7/src/autofetcher.ts#L54-L56

If `startEarly` isn't set, we'll initialize it later at the time we call `start()` on `AutoFetcher`: https://github.com/webrecorder/browsertrix-behaviors/blob/d89f7c668bf2edea5bc991620f106fed1cc96cf7/src/autofetcher.ts#L68

Which means it still gets initialized and run either way, but at a slightly later point in time.

For further context on the performance implications, I found this pretty informative looking Stack Overflow answer: https://stackoverflow.com/a/39332340

Haven't had the chance to fully vet this information, but the note on *when* you initialize `MutationObserver` lines up exactly with what we've been seeing here.

mistydemeo:
Passing `startEarly` causes us to initialize autofetch's `MutationObserver` immediately on the behaviour class being constructed instead of at the time we call `start()`. This has performance implications and ends up causing us to wait around an extra few seconds at the start of page load, causing many more complex pages to take longer to load than they properly need to.

We originally added this in c247189474e2786ad381a67e49447b03f2f35f0f in response to certain sites where we weren't waiting long enough for them to initialize, in particular Instagram. Now that we have `awaitPageLoad` for behaviours, we don't always have to do this for every page; we can do targeted behaviours to specific sites to wait on exactly the things we need them to, but avoid doing this on other sites where it's not gaining us anything.

This should be paired with https://github.com/webrecorder/browsertrix-behaviors/pull/160 to make sure fetching YouTube videos still works.

### DIFF
--- a/src/util/argParser.ts
+++ b/src/util/argParser.ts
@@ -904,7 +904,7 @@ class ArgParser {
         }
       });
       behaviorOpts.log = BxFunctionBindings.BehaviorLogFunc;
-      behaviorOpts.startEarly = true;
+      behaviorOpts.startEarly = false;
       behaviorOpts.clickSelector = argv.clickSelector;
       argv.behaviorOpts = JSON.stringify(behaviorOpts);
     } else {


mistydemeo:
Converting to draft until we've done more thorough testing.

ikreymer:
Superseded by #1154 - just removing the flag altogether, behaviors will always start autoplay detection early but mutationobserver will start later.

PR #1143, opened 2026-08-07
save the detected content-type:

ikreymer:
- set WARC-Identified-Payload-Type to `${document.contentType}; charset="${document.characterSet}"` from the page
- fixes capture part of webrecorder/replayweb.page#541
- depends on webrecorder/wabac.js#366

### DIFF
--- a/src/util/recorder.ts
+++ b/src/util/recorder.ts
@@ -584,8 +584,11 @@ export class Recorder extends EventEmitter {
       return;
     }
 
-    if (this.shouldSaveStorage && url === this.finalPageUrl) {
-      await this.saveStorage(reqresp, cdp);
+    if (url === this.finalPageUrl) {
+      if (this.shouldSaveStorage) {
+        await this.saveStorage(reqresp, cdp);
+      }
+      await this.saveDetectedContentType(reqresp, cdp);
     }
 
     try {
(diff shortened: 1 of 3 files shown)


src/util/recorder.ts
```diff
@@ -599,6 +602,25 @@ export class Recorder extends EventEmitter {
     }
   }
 
+  async saveDetectedContentType(reqresp: RequestResponseInfo, cdp: CDPSession) {
+    try {
+      const { result } = await cdp.send("Runtime.evaluate", {
+        expression:
+          '`${document.contentType}; charset="${document.characterSet}"`',
```
mistydemeo, line 609: Hm, it's a shame these values don't tell us whether these values were explicit in the underlying document or whether they were determined by the browser. It'd have been nice to only track this if it differed from what was in the document, not just if they were non-UTF8.

mistydemeo (approved):

src/util/recorder.ts
ikreymer: We sort of can, but it's more work.. we do have access to the Content-Type header here already, but not the `<meta>` tags that could also override it, and it'd be more effort to try to detect that. At replay time, we have access to all 3 if needed to identify discrepancies, but might as well prefer this value if it is available.

PR #1146, opened 2026-08-11
dedupe: don't add same WACZ file to per-crawl WACZ index

### LINKED ISSUE #1145: Dedupe: Restarting crawl can result in same WACZ file added multiple times to WACZ index
The `c:{crawlid}:wacz` may have multiple entries if crawler is restarted multiple times with same WACZ, eg. due to interruption.

ikreymer:
- Only add WACZ file to `c:{crawlid}:wacz` index when setting a new WACZ filename, not on every crawl start up
- Add it even if not using external redis, to keep logic simpler
- Fixes #1145 

Testing:
- Start a crawl with dedupe using an external Redis
- Interrupt crawler
- Restart crawl with dedupe
- Check that the `c:{crawlid}:wacz` has only one entry still.

In Browsertrix, start a crawl with scale > 1:
- check the coll index redis for `c:{crawlid}:wacz`
- force delete each of the crawler pods, eg. `delete pod -n crawlers {pod}-0 --force` and `delete pod -n crawlers {pod}-1 --force`.`
- The `lrange `c:{crawlid}:wacz` 0 -1` should still have only 2 entries. Without this fix, each time the pod restarts, a new entry is added to the list, incorrectly.

### DIFF
--- a/src/crawler.ts
+++ b/src/crawler.ts
@@ -1929,10 +1929,7 @@ self.__bx_behaviors.selectMainBehavior();
       this.params.generateWACZ &&
       (this.storage || this.isExternalDedupeStore)
     ) {
-      const filename = await this.crawlState.setWACZFilename();
-      if (this.isExternalDedupeStore) {
-        await this.crawlState.addSourceWACZForDedupe(filename);
-      }
+      await this.crawlState.setWACZFilename();
     }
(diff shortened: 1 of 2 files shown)


tw4l (approved):
Good catch and nice work!

PR #1150, opened 2026-08-20
dedupe: ensure WACZ filename is not cleared before it is set in per-wacz dedupe key

ikreymer:
- fixes regression introduced in #1146
- tests: refactor dedupe-basic test to allow accessing redis more easily
- tests: add test to verify `c:<crawlID>:wacz key` is as expected


tw4l (approved):
Thanks for the additional test!

PR #1151, opened 2026-08-21
Bump to Brave 1.93

ikreymer:
Disable query redir test pending fix for brave/brave-browser#58320


tests/brave-query-redir.test.ts
```diff
@@ -1,7 +1,9 @@
 import fs from "fs";
 import { ExecException, execSync } from "child_process";
 
-test("check that gclid query URL is automatically redirected to remove it", async () => {
+// disable pending brave issue
```
mistydemeo: Let's add the URL to the issue in here too so we don't lose track of it.

mistydemeo (changes requested):

tests/brave-query-redir.test.ts
ikreymer: Updated with link. Tested again with latest Brave build, still happening.
We're not relying on it, but good to know if this feature is available or not. Ideally, Brave fixes it, but also ok if it stays this way.

mistydemeo (approved):

PR #1153, opened 2026-08-24
Show the current page URL as a clickable link in the screencast view

aponb:
The screencast transport already sends the URL with each screencast image. This change only updates the static screencast HTML to render the URL as a link above the images, providing convenient additional information while watching a screencast session.


ikreymer:
Makes sense, tested locally, will add to 1.15 release.

PR #1160, opened 2026-09-04
browser: add configurable --browserStartTimeout

aponb:
puppeteer.launch() used its hardcoded 30s default for browser startup, too short on loaded VMs where
  the launch fails with a TimeoutError and aborts the crawl. Adds --browserStartTimeout (seconds, default
  30) for crawl and create-login-profile, plus an actionable fatal on launch timeout.

### DIFF
--- a/docs/docs/user-guide/cli-options.md
+++ b/docs/docs/user-guide/cli-options.md
@@ -174,6 +174,11 @@ Options:
       --profile, --loadProfile              Path or HTTP(S) URL to tar.gz file w
                                             hich contains the browser profile di
                                             rectory                     [string]
+      --browserStartTimeout                 Timeout (in seconds) for the browser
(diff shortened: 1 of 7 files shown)


ikreymer (approved):

ikreymer:
This should work, though I wonder if the load is that high, it seems other parts of the crawl will also be degraded?
Eg. behavior timeout and link extraction will likely also timeout?

aponb:
Good point! I was a bit imprecise when I used the term "high load." In our virtualization setup, we are experiencing high I/O latency when certain data needs to be retrieved from central storage and made available to the VM. This is the actual issue causing our timeout. Once the data is in cache, everything is running smoothly. Of course this underlying problem needs to be resolved first. Until then, increasing the timeout is a functional workaround for us.

PR #1162, opened 2026-09-10
skippedPages: write text format + enable reportSkipped by default

mistydemeo:
This hints to frictionless that it shouldn't validate it the same way as the other page documents we write.

With this fix, can now enable `--reportSkipped` to be true by default!


ikreymer:
This needs to be added to datapackage.json `resources` field, not here.

mistydemeo:
Moved into datapackage.json.

ikreymer (approved):
Also made reportSkipped to be true by default, which is now possible with this fix, as all tests can pass.

src/util/argParser.ts
```diff
@@ -764,7 +764,7 @@ class ArgParser {
           describe:
             "If set, write information about URLs encountered but not queued to reports/skippedPages.jsonl",
           type: "boolean",
-          default: false,
+          default: true,
```
mistydemeo: We're good swapping defaults here?

PR #1165, opened 2026-09-11
Register execution contexts via injected behaviors.

### LINKED ISSUE #1164: Use behaviors to register execution context/frames that can have behaviors + link extraction.
Instead of using page.frames(), which appears to be unreliable in latest Chromium for cross-origin OOPIF, use the existing system of automatically injecting behaviors into each frame to register the frame with the crawler, allowing the crawler to run link extraction + behaviors in each registered frame.

This ensures that all frames that have behaviors injected will have link extraction + behaviors executed, instead of using a different process to discover the frames.

This can also ensure that dynamically added frames will have link extraction + behaviors run.

ikreymer:
- Use CDP bindings to discover and register iframes
- Behaviors calls registerFrame() (in 0.14.1) which registers the 'execution context id' of the caller.
- Crawler keeps track of context ids, calls runInExecContext() which uses Runtime.evaluate() for each registered context id.
- Also calls get current URL of each context, filter out empty / about:blank contexts same as before.
- Use same approach for both link extraction and running behaviors.
- Support dynamically added frames: if new frames (execution contexts) added after link extraction or behaviors have started, run those on those frames immediately.

Notes:
- This works as behaviors are reliably injected into all iframes with page.evaluateOnNewDocument(), including cross-origin iframes. However, page.frames() appears to not discover these frames. May need to do add further handling (eg. with setAutoAttach) if behaviors are not injected into cross-origin iframes.
- Need to use raw CDP Runtime.addBinding and Runtime.bindingCalled events to get the execution context ids, which are not exposed by puppeteer. The __bx_registerFrame binding is handled directly, unlike other function bindings added with puppeteer's page.exposeFunction.

Fixes #1164


src/crawler.ts
```diff
@@ -872,7 +873,58 @@ export class Crawler {
...
+              logger.debug(
+                "Dynamically added iframe behaviors failed",
+                e,
+                "behavior",
+              );
+            }
+          }
+        };
+
+        void processNewIframe();
```
mistydemeo: Do we not need to `await` this?

src/util/state.ts
```diff
@@ -110,7 +109,11 @@ export class PageState {
 
   isDirectFetched = false;
 
-  filteredFrames: Frame[] = [];
+  //filteredFrames: Frame[] = [];
```
mistydemeo: What's the longterm plan for `filteredFrames`?

src/crawler.ts
ikreymer: It doesn't matter too much, since it already in an event handler, but can make the event handler async and await.

src/util/state.ts
ikreymer: This can be removed

src/replaycrawler.ts
```diff
@@ -462,7 +462,7 @@ export class ReplayCrawler extends Crawler {
 
     data.isHTMLPage = true;
 
-    data.filteredFrames = page.frames().slice(SKIP_FRAMES);
+    //data.filteredFrames = page.frames().slice(SKIP_FRAMES);
```
mistydemeo: Should we just remove this?

ikreymer: Yes, but also reminded me to add proper filtering for replaycrawler. Now filtering out the top frame and the source frame from any link extraction + behaviors for QA, as that's just the top-level frame.

src/crawler.ts
```diff
@@ -1801,7 +1801,8 @@ self.__bx_behaviors.selectMainBehavior();
     }
 
     const includeFrame =
-      this.adBlockRules && !this.adBlockRules.isAdUrl(frameUrl);
+      this.shouldIncludeFrameUrl(frameUrl, isTop) &&
+      (!this.adBlockRules || !this.adBlockRules.isAdUrl(frameUrl));
```
mistydemeo, line 1805: Is this altered conditional right? This would mean if `this.shouldIncludeFrameUrl(frameUrl, isTop)` was true and no adblock rules were defined, we wouldn't include the frame.

ikreymer: I think its the other way? `(!this.adBlockRules || !this.adBlockRules.isAdUrl(frameUrl))` should be either 'no adBlockRules' or 'not an ad'?

mistydemeo (approved):

